PR #66, opened 2017-05-05
Mention the available parameters for the DSN.

mqsoh:
### Description
I pulled these out of the switch in `dsn.parseDSNParams`. I didn't include `account`, `database`, or `schema` because they are in the body of the DSN.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing


smtakeda:
Thanks for updating the doc. Looking.

smtakeda:
@mqsoh I didn't include `protocol` as there is no meaning for customers. That's used only in our test env.

Issue #111, opened 2017-09-12
Make logging optional


disq:
Looks like logging is currently used only for debugging. Also, libraries using their own logging mechanisms are mostly frowned upon, here are some references:

- https://www.0value.com/about-Go-logging
- https://peter.bourgon.org/go-best-practices-2016/#logging-and-instrumentation
- https://github.com/go-kit/kit/issues/42

To remedy this, maybe we could create a compatible wrapper for glog and separate the actual glog import/usage from a no-op one using build tags. Say, a build tag named `sfdebug` would enable the actual glog wrapper and compile with glog, and without it would just use a null-logger (to satisfy the logging calls during compile) and not import glog.

This would also fix #106 since `flag.Parse()` is called by `glog`.

Would you consider such a solution? If so, we could send in a PR and go from there.

smtakeda:
That sounds good. Please go ahead to file PR! thanks.

PR #112, opened 2017-09-13
Make logging optional

disq:
### Description
glog is now optional as described in #111.

This is achieved using a package-level variable named `glog` instead of the glog import. glog is only imported in `log_debug.go`, which is included only if the `sfdebug` build-tag is set, like `go build -tags=sfdebug ./cmd/select1`. In normal mode glog calls are now no-ops and `log.go` provides appropriate wrappers for them.

The use of goimports was unnecessary and was also causing a bug: if there is glog installed in GOPATH, it was being incorrectly imported, breaking the build. This is a known issue of goimports, as described in https://github.com/golang/go/issues/7463. The use of goimports was removed only from the main Makefile, to be replaced with `gofmt` which is included in the standard Go distribution.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ all: fmt lint cov
 setup:
 	go get github.com/Masterminds/glide
 	go get github.com/golang/lint/golint
-	go get golang.org/x/tools/cmd/goimports
 	go get github.com/Songmu/make2help/cmd/make2help
 
 ## Run tests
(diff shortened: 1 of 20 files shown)


smtakeda:
will look into this soon.

smtakeda:
@disq can you please review https://github.com/snowflakedb/gosnowflake/pull/114? I wanted to push update here but was not sure how to do it... Instead, I've raised another PR to include your change and other misc updates.

disq:
Closed by #114

PR #113, opened 2017-09-13
Remove committed binary and fix .gitignore definitions

disq:
A `selectmany` binary was incorrectly committed, because of a glitch in the tool's `.gitignore`. This PR removes it, also adding entries for each cmd to the main `.gitignore` file.


smtakeda:
Thanks for cleaning up! Evaluating the changes.

PR #115, opened 2017-09-15
Move signal handling to cmd/

disq:
### Description
Signal handling code was moved to individual cmd/ commands. Queries are still automatically canceled with the parent context.

Since this package already requires Go 1.8, and the "context" family of database/sql methods are included in the 1.8 standard library, there is no need to use an internal and implicit signal handler to cancel queries.

### Reasoning
Signal handling should only be handled by a process, always explicitly (often by `package main`)
In the current implementation, this functionality is hidden in various parts of the library.

This change enables the user to handle `os.Interrupt` signals differently, ie. A single Ctrl+C could warn the user or do a soft-cancel, while a second one could cancel the running queries and quit. If the user wants to cancel a specific query, he/she can explicitly cancel the parent context.

Please see [./cmd/selectmany](https://github.com/peakgames/gosnowflake/blob/contextonly-cancellation/cmd/selectmany/selectmany.go#L55) for an example.

### Notes
Additionally, in retry.go:113, the parent context's cancellation was ignored. So, if the user wanted to cancel a query, by cancelling the provided context, retry logic would not stop on some occasions. (try with a ` context.WithTimeout(context.Background(), n)`, where `n` is lower than the backend's average response time)

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/cmd/select1/select1.go
+++ b/cmd/select1/select1.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"context"
 	"database/sql"
 	"flag"
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
 
 	_ "github.com/snowflakedb/gosnowflake"
 )
@@ -15,6 +17,20 @@ func main() {
 		// enable glog for Go Snowflake Driver
 		flag.Parse()
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	defer func() {
+		signal.Stop(c)
+		cancel()
+	}()
+	go func() {
+		<-c
+		log.Println("Caught signal, canceling...")
+		cancel()
+	}()
+
 	// get environment variables
 	env := func(k string) string {
 		if value := os.Getenv(k); value != "" {
@@ -35,7 +51,7 @@ func main() {
 		log.Fatalf("failed to connect. %v, err: %v", dsn, err)
 	}
 	query := "SELECT 1"
-	rows, err := db.Query(query)
+	rows, err := db.QueryContext(ctx, query)
 	if err != nil {
 		log.Fatalf("failed to run a query. %v, err: %v", query, err)
 	}
--- a/cmd/selectmany/selectmany.go
+++ b/cmd/selectmany/selectmany.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"context"
 	"database/sql"
 	"flag"
 	"fmt"
 	"log"
 	"os"
+	"os/signal"
 
 	_ "github.com/snowflakedb/gosnowflake"
 )
@@ -15,6 +17,20 @@ func main() {
 		// enable glog for Go Snowflake Driver
 		flag.Parse()
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	defer func() {
+		signal.Stop(c)
+		cancel()
+	}()
+	go func() {
+		<-c
+		log.Println("Caught signal, canceling...")
+		cancel()
+	}()
+
 	// get environment variables
 	env := func(k string) string {
 		if value := os.Getenv(k); value != "" {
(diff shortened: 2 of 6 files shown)


mathyourlife-fitbit:
agreed, signal handling should be left up to the application not at the package level.

disq:
In this case we could also remove `ErrCanceled` and `ErrCodeCanceled` since they are now replaced by `context.Canceled`. Thoughts?

smtakeda:
@disq thanks for PR. yes, it makes sense. Let's remove `ErrCanceled` and `ErrCodeCanceled`.

smtakeda:
If no other change is required, I'll merge all change together.

smtakeda:
Perhaps it is better to have a default signal handler if the application call methods that don't take context? For those applications, there is no control on query cancellation.

smtakeda:
hmm, it is tough to add _default_ context. Golang SQL driver manager directly calls `QueryContext` with `context.Background()` even if `Query` method exits in the driver. I attempted to add the default cancel context in `Query` method in the driver, but the method never used.

In `QueryContext`, the context could be checked if it is `context.Background()`, but it doesn't tell whether the application really wants to use `context.Background()` or not.

connection.go:
```
func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
	glog.V(2).Infoln("Query")
	if sc.rest == nil {
		return nil, driver.ErrBadConn
	}
	if ctx == context.Background() {
		ctx1, cancel := context.WithCancel(context.Background())
		c := make(chan os.Signal, 1)
		signal.Notify(c, os.Interrupt)
		defer func() {
			signal.Stop(c)
			cancel()
		}()
		go func() {
			<-c
			glog.V(2).Info("Caught signal, canceling...")
			cancel()
		}()
		ctx = ctx1
	}
	// TODO: handle noResult and isInternal
	data, err := sc.exec(ctx, query, false, false, args)
	if err != nil {
		glog.V(2).Infof("error: %v", err)
		return nil, err
	}

	rows := new(snowflakeRows)
	rows.sc = sc
	rows.RowType = data.Data.RowType
	rows.ChunkDownloader = &snowflakeChunkDownloader{
		sc:                 sc,
		ctx:                ctx,
		CurrentChunk:       data.Data.RowSet,
		ChunkMetas:         data.Data.Chunks,
		Total:              int64(data.Data.Total),
		TotalRowIndex:      int64(-1),
		Qrmk:               data.Data.Qrmk,
		FuncDownload:       downloadChunk,
		FuncDownloadHelper: downloadChunkHelper,
		FuncGet:            getChunk,
	}
	rows.ChunkDownloader.start()
	return rows, err
}
```

In other words, all applications that naively use `Query` and `Exec` will not able to cancel the query. Is this acceptable? If above snippet is reasonable, I would rather go with the option on top of the given change.

igungor:
I agree with @disq. Signal handling should be explicit, and packages should have no side-effects. Handling a process signal (also which ones? os.Interrupt? what about syscall.SIGTERM, which is OS-dependent.) has nothing to do with a database driver.

Correctness and having no-surprises is the main driver for me to add a package to my dependency list. Implicit signal handling (even if it is documented) makes me think harder about checking the package out to my codebase. I have to remember whether if a database driver would interfere with my cancelation logic. What if I want a program to cancel all queries on a second Ctrl-C?

As for Query and Exec (non-context counterparts) cancelation, there could be a driver constructor that accepts several options to change driver behaviour. Even though I've not seen such an example in a database driver, if the programmer really wants to cancel queries on an interrupt, he/she can register the driver (with custom options) manually. Please take a look at gRPC (https://godoc.org/google.golang.org/grpc#NewServer). gosnowflake could take a similar approach:

```
func NewDriver(opts ...Option) (driver.Driver, error)

type WithSignalhandler(signals ...os.Signal) Option

type Options func(*options)
```

It is explicit (which is always welcome), *and* the programmer can specify which signals to listen to. What do you think?

smtakeda:
@igungor and @disq Fair enough. I'll merge this PR #115 and drop PR #116.

smtakeda:
#117 includes doc, comment and sample code updates based on the discussion here. Code reviews are welcome.

smtakeda:
merged #117 including #115.

PR #116, opened 2017-09-16
Based on PR 115, added default cancel context for legacy method calls

smtakeda:
### Description
Added the default context on top of #115

This patch is for those applications that naively calls `Query` or `Execute` that don't have context parameters to still allow them to cancel queries by Ctrl+C. While #115 yields flexibility of signal handling to applications (that's great), ones already written to call legacy methods can no longer take advantage of builtin Ctrl+C handler.

I'm not sure if we want to enforce all applications to have context explicitly or don't need to care those applications that use legacy `Query` and `Execute` methods just because 1.8 is the minimum requirement.
 
@disq and anybody watching this repo, any thought?

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -175,6 +175,36 @@ To get the logs for a specific module, use the ``-vmodule`` option. For example,
 
     The logger may be changed in the future for better logging. Currently if the applications use the same parameters as ``glog``, you cannot collect both application and driver logs at the same time.
 
+Canceling a Query by Ctrl+C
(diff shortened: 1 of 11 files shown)


igungor:
I've written my comments on the original PR. Please take a look: https://github.com/snowflakedb/gosnowflake/pull/115#issuecomment-330141600

smtakeda:
Won't merge.

PR #119, opened 2017-09-19
Fix checking the identical errors

igungor:
### Description

left and right handside of the comparison are identical. These types of typos/errors are easily be spotted with a static analyzer.

What do you think about adding such a tool to CI?

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -151,7 +151,7 @@ func (sc *snowflakeConn) BeginTx(ctx context.Context, opts driver.TxOptions) (dr
 		return nil, driver.ErrBadConn
 	}
 	_, err := sc.exec(ctx, "BEGIN", false, false, nil)
-	if err != err {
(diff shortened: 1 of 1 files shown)


smtakeda:
Ah, thanks for reporting a bug. What tool would you recommend? Will look into the options.

smtakeda:
Adding https://goreportcard.com/report/github.com/snowflakedb/gosnowflake to README.rst

igungor:
gometalinter is very popular but my I use [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck) after before every commit.

smtakeda:
Thanks. I'm adding megacheck to lint target. See https://github.com/snowflakedb/gosnowflake/pull/120

PR #121, opened 2017-10-06
Added dep manifest files

CrimsonVoid:
### Description
Added Gopkg.{toml,lock} files which mimic glide.{yaml,lock} respectively. `dep` automatically imported the glide manifests when running `dep init`, but I also did a manual check the lock files were consistent.

Unfortunately I wasn't able to completely get rid of glide from the makefile, because dep doesn't seem to have an equivalent to `glide novendor`. I'm not sure if it is still viable to go ahead with `dep` at this point due to the limitations.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


smtakeda:
Really appreciate your work on dep fix! Unfortunately I'm half offline next couple of weeks. Will follow up as soon as I can. Thanks!

smtakeda:
Pushed the change to master and bumped up the version to v1.0.0
Thank you!

PR #124, opened 2017-11-17
Raise error if the specified db, schema or warehouse doesn't exist

### LINKED ISSUE #110: raise error if the specified db, schema or warehouse doesn't exist
### Issue description
Golang driver should raise error if the specified db, schema or warehouse doesn't exist.
The role is already validated, but not others are. This is more or less historical reason. no-error-for-invalid db behavior has been around since there was no concept of default db, schema or warehouse. Well, even no warehouse exists 3.5 years ago.

smtakeda:
### Description
Fixed #110 

This change will raise errors if the specified db, schema or warehouse doesn't exist at connection time. The role is already validated, so others should be too.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


smtakeda:
Also error code issue was fixed too. Somehow `iota` doesn't work to assign the continuous ids for errors but all were set to the same number...

PR #145, opened 2018-02-05
Support for OAuth with Snowflake.

tjj5036:
### Description
OAuth Authentication Support (followup to previous PR)

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


dsn.go
```diff
@@ -98,6 +100,13 @@ func DSN(cfg *Config) (dsn string, err error) {
 	if cfg.Application != clientType {
 		params.Add("application", cfg.Application)
 	}
+	if cfg.Token != "" {
+		escapedToken, err := url.QueryUnescape(cfg.Token)
```
smtakeda: https://github.com/snowflakedb/gosnowflake/blob/0285e80731d13676d55c1929f8b7d71e42d2d897/dsn.go#L373 already unescaped the value before cfg.Token is assigned. Is there any reason why unescape here again?

smtakeda: Other than the above comment. Looks good to me.

tjj5036: Ah I didn't see that earlier QueryUnescape... just removed mine and re-ran the tests and it works! Will update the branch accordingly.

auth.go
```diff
@@ -207,6 +209,10 @@ func authenticate(
 		requestMain.Authenticator = externalBrowserAuthenticator
 	} else if !bytes.Equal(samlResponse, []byte{}) {
 		requestMain.RawSAMLResponse = string(samlResponse)
+	} else if token != "" {
```
smtakeda: This checking may not be good enough?

tjj5036: I think it's ok for now since token is a string and you can't have nil for a string:

```
var s string
fmt.Println(s=="") // prints "true"
```

Provided that the token is at least one character, we'll go into that else if.

smtakeda: I mean token != "" doesn't necessarily oAuth authenticator, right?

tjj5036: I think right now a token that != "" means that the user is attempting to use OAuth since the golang driver does not support any other token based authentication mechanisms. What makes the `authenticate` call in `auth.go` a little unclean here is that "authenticator" is never passed in to authenticate - we just take in every possible value that each authenticator uses and infer which authenticator to use based on the values. For instance, `samlResponse` not being an empty byte array  implies Okta, `proofKey` not being an empty byte array implies external browser authentication, etc.

smtakeda: ok, fair enough. I agree that we need to clean up those checkings.

tjj5036: I can move some stuff around... let me create another PR with those changes!

PR #148, opened 2018-02-06
Cleanup of auth.go

tjj5036:
### Description
This is a followup to an earlier PR. The authenticator connection parameter should be used in auth.go, rather than inferring the authenticator from the values passed in.

Tested via setting connection parameters in `parameters.json` and running `make test`.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -3,7 +3,6 @@
 package gosnowflake
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


smtakeda:
Looks great! thanks you!

smtakeda:
@tjj5036 Do you want me to add a release tag including the recent changes? If you are planning to do the public preview of OAuth support, it should have a release.

tjj5036:
That's a good question - regarding external browser authentication, I think that should have a release tag. I would hold off on OAuth at the moment.

smtakeda:
ok. I should have noticed that time... Realized that I would need to bump up the version in version.go as well. Is it ok to add v1.1.5 to the latest commit along with version.go change but not to mention OAuth in release note?

tjj5036:
Sounds good!

PR #152, opened 2018-02-26
Detach chunk dowloader memory alter used

smtakeda:
### Description
Detach chunk loader memory after used so that GC may reclaim the memory. However it is not guaranteed that GC will run right after it is detached but usually in 5 min. In my experiments, the memory is freed after 3~4 minutes. In order to force to remove the memory, the applications may call `debug.FreeOSMemory`.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/benchmark/largesetresult/largesetresult_test.go
+++ b/benchmark/largesetresult/largesetresult_test.go
@@ -13,8 +13,11 @@ import (
 	"database/sql"
 
 	"context"
-	_ "github.com/snowflakedb/gosnowflake"
 	"os/signal"
+
+	"runtime/debug"
+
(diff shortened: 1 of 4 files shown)


smtakeda:
Note here is how I measured memory usage.

Go to the directory.
https://github.com/snowflakedb/gosnowflake/tree/master/benchmark/largesetresult
and run
```
go test -trace trace.out
```
followed by
```
go tool trace largesetresult.test trace.out
```
Check "Heap" history.

In the experiments, the maximum memory usage was 2.3GB and shrunk to 600MB around the end of process. Note `debug.FreeOSMemory()` is called on every 1 million records in the test case or it will need to wait for 5 min to get memory freed.

PR #157, opened 2018-03-11
Added OS info to the session

smtakeda:
### Description
Updated OS info in the session.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -28,14 +28,18 @@ const (
 	authenticatorOkta            = "OKTA"
 )
 
-// platform consists of compiler, OS and architecture type in string
-var platform = fmt.Sprintf("%v-%v-%v", runtime.Compiler, runtime.GOOS, runtime.GOARCH)
+// platform consists of compiler and architecture type in string
+var platform = fmt.Sprintf("%v-%v", runtime.Compiler, runtime.GOARCH)
+
+// operatingSystem is the runtime operating system.
+var operatingSystem = runtime.GOOS
 
 // userAgent shows up in User-Agent HTTP header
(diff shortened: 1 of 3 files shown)


auth.go
```diff
@@ -28,14 +28,18 @@ const (
 	authenticatorOkta            = "OKTA"
 )
 
-// platform consists of compiler, OS and architecture type in string
-var platform = fmt.Sprintf("%v-%v-%v", runtime.Compiler, runtime.GOOS, runtime.GOARCH)
+// platform consists of compiler and architecture type in string
+var platform = fmt.Sprintf("%v-%v", runtime.Compiler, runtime.GOARCH)
+
+// operatingSystem is the runtime operating system.
+var operatingSystem = runtime.GOOS
```
smtakeda: Moved OS info from OS_VERSION to OS.

Currently no OS version can be obtained by API unless it runs a command, e.g., uname, cmd ver. I want to avoid to run an external command.

auth.go
```diff
@@ -28,14 +28,18 @@ const (
...
 
 // userAgent shows up in User-Agent HTTP header
 var userAgent = fmt.Sprintf("%v/%v/%v/%v", clientType, SnowflakeGoDriverVersion, runtime.Version(), platform)
 
 type authRequestClientEnvironment struct {
 	Application string `json:"APPLICATION"`
+	Os          string `json:"OS"`
```
smtakeda: The value would be `linux`, `darwin`, `windows`, etc.

This information is used to collect client stats.

Issue #160, opened 2018-04-05
Handle 390114 "Authentication token has expired" response from server


yemble:
### Issue description

We've been receiving the following error from Snowflake:
```
390114: Authentication token has expired.  The user must authenticate again.
```

It's not being caught and handled in the driver. `restful.go` has
```go
sessionExpiredCode       = "390112"
```

Perhaps the error number has changed, or this is a new variant which needs to be handled the same way?

### Error log

As above

### Configuration

Long running kafka->snowflake process.

smtakeda:
There are two tokens issued for the session, master and session token. The master token is valid for 4 hours, while the session token is valid 1 hours. Obviously the session token expires first, `390112` is returned, then the driver will automatically renew it using the master token and update the master token itself. So in theory, if the token renewal occurs every hour, the connection is valid forever.

However in order to renew tokens, the clients require to communicate with the server continuously, so if the session is idle for 4 hours or more, both master and session tokens get expired, and there is no way to renew them. That's the timing when you get `390114`.

`390112` is used internally for token renew, so the application doesn't need to worry about it, but `390114` requires to be caught and ask the user and password to the users.

One way to mitigate this is that the application keeps running a query, for example `select 1` every hour to force the tokens are refreshed.

JDBC and ODBC has heartbeat option but not implemented in Gosnowflake yet.

yemble:
Thanks for that explanation. Reopening the connection will suffice until the heartbeat feature is implemented. In practice we're unlikely to be idle for such a long time.

jerry-b:
We had to work around this as well, the driver will be much more convenient when there is a built in solution.

smtakeda:
An initial implementation. A simple test passed.
https://github.com/snowflakedb/gosnowflake/pull/164

Would be great if somebody reviews the code.

rem7:
@smtakeda any particular reason why you guys are choosing to do a heartbeat? I believe usually clients (mysql, postgres) that implement database/sql simply auto-reconnect if the connection gets broken. Why not just issue a reconnect in the background when `390114` is encountered?

smtakeda:
[...]

PR #164, opened 2018-05-21
Heartbeat

smtakeda:
### Description
Added CLIENT_SESSION_KEEP_ALIVE option support that keeps the connection alive by heartbeating it.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/cmd/keepalive/keepalive.go
+++ b/cmd/keepalive/keepalive.go
@@ -1,6 +1,6 @@
 // Example: client session keep alive
 // By default, the token expires in 4 hours if the connection is idle. CLIENT_SESSION_KEEP_ALIVE parameter will
-// have a heartbeat in the backend to keep the connection alive explicit heartbeats.
+// have a heartbeat in the background to keep the connection alive by making explicit heartbeats
 package main
 
 import (
--- a/cmd/keepalive/keepalive.go
+++ b/cmd/keepalive/keepalive.go
@@ -34,6 +34,10 @@ func getDSN() (string, *sf.Config, error) {
 	port := env("SNOWFLAKE_TEST_PORT", false)
 	protocol := env("SNOWFLAKE_TEST_PROTOCOL", false)
 
+	params := make(map[string]*string)
+	valueTrue := "true"
+	params["client_session_keep_alive"] = &valueTrue
+
 	portStr, _ := strconv.Atoi(port)
 	cfg := &sf.Config{
 		Account:  account,
@@ -42,10 +46,10 @@ func getDSN() (string, *sf.Config, error) {
 		Host:     host,
 		Port:     portStr,
 		Protocol: protocol,
+		Params:   params,
 	}
 
 	dsn, err := sf.DSN(cfg)
-	dsn += "&client_session_keep_alive=true" // CLIENT_SESSION_KEEP_ALIVE
 	return dsn, cfg, err
 }
 
--- a/cmd/keepalive/keepalive.go
+++ b/cmd/keepalive/keepalive.go
@@ -1,0 +1,105 @@
+// Example: client session keep alive
+// By default, the token expires in 4 hours if the connection is idle. CLIENT_SESSION_KEEP_ALIVE parameter will
+// have a heartbeat in the backend to keep the connection alive explicit heartbeats.
+package main
+
+import (
+	"database/sql"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+
+	sf "github.com/snowflakedb/gosnowflake"
+	"time"
+)
+
+// getDSN constructs a DSN based on the test connection parameters
+func getDSN() (string, *sf.Config, error) {
+	env := func(k string, failOnMissing bool) string {
+		if value := os.Getenv(k); value != "" {
+			return value
+		}
+		if failOnMissing {
+			log.Fatalf("%v environment variable is not set.", k)
+		}
+		return ""
+	}
+
+	account := env("SNOWFLAKE_TEST_ACCOUNT", true)
+	user := env("SNOWFLAKE_TEST_USER", true)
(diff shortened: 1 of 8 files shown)


smtakeda:
Hey @tjj5036 , would you have a chance to review this change? Thanks.

cmd/keepalive/keepalive.go
```diff
@@ -0,0 +1,109 @@
...
+	account := env("SNOWFLAKE_TEST_ACCOUNT", true)
+	user := env("SNOWFLAKE_TEST_USER", true)
+	password := env("SNOWFLAKE_TEST_PASSWORD", true)
+	host := env("SNOWFLAKE_TEST_HOST", false)
+	port := env("SNOWFLAKE_TEST_PORT", false)
+	protocol := env("SNOWFLAKE_TEST_PROTOCOL", false)
+
+	params := make(map[string]*string)
+	valueTrue := "true"
+	params["client_session_keep_alive"] = &valueTrue
```
smtakeda: A new parameter CLIENT_SESSION_KEEP_ALIVE for golang driver.
This is already supported by JDBC and ODBC drivers.

heartbeat.go
```diff
@@ -0,0 +1,115 @@
...
+			hbTimer.Stop()
+			glog.V(2).Info("stopping")
+			return
+		default:
+			// no more download
+			glog.V(2).Info("no shutdown signal")
+		}
+		time.Sleep(1 * time.Second)
+	}
+}
```
smtakeda: main routine. Not sure if this way is optimal. Any suggestion would be appreciated.

cmd/keepalive/keepalive.go
```diff
@@ -0,0 +1,109 @@
+// Example: client session keep alive
+// By default, the token expires in 4 hours if the connection is idle. CLIENT_SESSION_KEEP_ALIVE parameter will
+// have a heartbeat in the backend to keep the connection alive explicit heartbeats.
```
tjj5036: have a heartbeat in the background to keep the connection alive by making explicit heartbeats

cmd/keepalive/keepalive.go
```diff
@@ -0,0 +1,109 @@
...
+
+	time.Sleep(90 * time.Minute)
+
+	runQuery(db, query)
+
+	time.Sleep(4 * time.Hour)
+
+	runQuery(db, query)
+
+	time.Sleep(8 * time.Hour)
```
tjj5036: Did you actually run this and wait for it to finish? Impressive if so :)

connection.go
```diff
@@ -304,3 +311,33 @@ func (sc *snowflakeConn) populateSessionParameters(parameters []nameValueParamet
 		sc.cfg.Params[strings.ToLower(param.Name)] = &v
 	}
 }
+
+func (sc *snowflakeConn) isClientSessionKeepAliveEnabled() bool {
+	ret := false
+	for k, v := range sc.cfg.Params {
+		if k == sessionClientSessionKeepAlive {
+			if strings.Compare(*v, "true") == 0 {
```
tjj5036: Will this always be lowercased? Also, instead of iterating through the entire map, I think you can do something like:

```
_ ok := sc.cfg.Params["sessionClientSessionKeepAlive"]
if !ok {
    return false
}
return strings.Compare(sc.cfg.Params["sessionClientSessionKeepAlive"], "true") == 0
```

heartbeat.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (hc *heartbeat) start() {
+	hc.ShutdownChan = make(chan int)
+	go hc.run()
```
tjj5036: I think a go routine here is the right way to go

heartbeat.go
```diff
@@ -0,0 +1,115 @@
...
+		glog.V(2).Infof("heartbeatMain: resp: %v", resp)
+		var respd execResponse
+		err = json.NewDecoder(resp.Body).Decode(&respd)
+		if err != nil {
+			glog.V(1).Infof("failed to decode JSON. err: %v", err)
+			glog.Flush()
+			return err
+		}
+		if respd.Code == sessionExpiredCode {
+			err = hc.restful.FuncRenewSession(context.TODO(), hc.restful)
```
tjj5036: Does this update Token on restful?

connection.go
```diff
@@ -122,6 +126,7 @@ func (sc *snowflakeConn) exec(
 	sc.QueryID = data.Data.QueryID
 	sc.SQLState = data.Data.SQLState
 	sc.populateSessionParameters(data.Data.Parameters)
+	sc.startHeartBeat()
```
tjj5036: Does this start heartbeating after a successful connection is made?

cmd/keepalive/keepalive.go
smtakeda: I did three time in the tests. All worked well.

cmd/keepalive/keepalive.go
smtakeda: Fixed.

connection.go
smtakeda: ah, ok. Yes, the value should be lower cases.

smtakeda:
```
	v, ok := sc.cfg.Params[sessionClientSessionKeepAlive]
	if !ok {
		return false
	}
	return strings.Compare(*v, "true") == 0
```

heartbeat.go
smtakeda: yes, it is exactly the same as postRestfulQueryHelper's FuncRenewSession

connection.go
smtakeda: Yes, it is

smtakeda:
That's right. This session won't expire with this option. This is already implemented in JDBC and ODBC drivers.

smtakeda:
Fixed #160

smtakeda:
@tjj5036 thanks for the review!

PR #171, opened 2018-06-16
Change UUID dependency

kenshaw:
The satori UUID package causes problems with go get and vgo when
building in a Go application using a different version of the satori
UUID package. This is because the satori package introduced breaking API
changes and has not tagged a new version yet.

This change converts the UUID generation to use the Google UUID package
instead.

Additionally, this is needed by github.com/xo/usql so that go get and
vgo build both work "out of the box".

The following is a small Go program that demonstrates that the UUIDs
generated by the two packages will have the same format:

```go
package main

import (
        "log"

        guuid "github.com/google/uuid"
        huuid "github.com/hashicorp/go-uuid"
        suuid "github.com/satori/go.uuid"
)

func main() {
        g, err := guuid.NewRandom()
        if err != nil {
                log.Fatal(err)
        }

        h, err := huuid.GenerateUUID()
        if err != nil {
                log.Fatal(err)
        }

        s, err := suuid.NewV4()
        if err != nil {
                log.Fatal(err)
        }

        log.Printf(
                "google: %s, hashicorp: %s, satori: %s",
                g, h, s,
        )
}
```

### Description

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


smtakeda:
ok, will take a look soon.

smtakeda:
Looks like https://github.com/google/uuid is not stable too. But I understand it can mitigate go get issue to some extent.

smtakeda:
Merged to #172

PR #180, opened 2018-07-28
optimize decoding of large-chunk 

mhseiden:
### Description
From the primary commit message.

>    The primary goal is to reduce to in-flight memory
    consumption to be as low as possible. This is done
    by making assumptions about the chunk structure
    and using a streaming decoder.
    
>    The secondary goal is to optimize the CPU cycles
    needed to decode the chunk. There's still more
    to do on this front, but the design in this patch
    should be conducive to future optimizations.

The impetus for this change was primarily to reduce memory pressure when downloading large `json` results. However this optimization should reduce pressure for all large results. For example, I see the following improvement for the existing large-result benchmark.

```
-- before change
BenchmarkLargeResultSet-4 1    68703150935 ns/op   6451121456 B/op    270103889 allocs/op
-- after change
BenchmarkLargeResultSet-4 1    57140518986 ns/op   3198717528 B/op    220123790 allocs/op
```

For the new `json` case, I see the following improvement. It's worth noting that I ensured my test runs used the same cached result; future comparisons may see different numbers if the cached rowset (in `S3`) is different.

```
-- before change
BenchmarkJsonResultSet-4 1  114852005941 ns/op  2964585432 B/op  149820 allocs/op
-- after change
BenchmarkJsonResultSet-4 1  68804250571 ns/op   1056399984 B/op  288150 allocs/op
```

It's worth mentioning that before hand-writing a `json` decoder, I tried a few different libraries. None of them performed as well as I would have liked, and one of the streaming libraries actually performed worse (with regards to memory pressure) than `encoding/json`. 

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/chunk.go
+++ b/chunk.go
@@ -1,0 +1,307 @@
+package gosnowflake
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"unicode"
+	"unicode/utf16"
+	"unicode/utf8"
+)
+
+const (
+	defaultChunkBufferSize  int64 = 8 << 10 // 8k
+	defaultStringBufferSize int64 = 512
+)
+
+type largeChunkDecoder struct {
+	r io.Reader
+
+	rows  int // hint for number of rows
+	cells int // hint for number of cells/row
+
+	rem int // bytes remaining in rbuf
+	ptr int // position in rbuf
+
+	rbuf []byte
+	sbuf *bytes.Buffer // buffer for decodeString
+
+	ioError error
+}
+
+func decodeLargeChunk(r io.Reader, rowCount int, cellCount int) ([][]*string, error) {
+	lcd := largeChunkDecoder{
+		r, rowCount, cellCount,
+		0, 0,
+		make([]byte, defaultChunkBufferSize),
+		bytes.NewBuffer(make([]byte, defaultStringBufferSize)),
+		nil,
+	}
+
+	rows, err := lcd.decode()
+	if lcd.ioError != nil && lcd.ioError != io.EOF {
+		return nil, lcd.ioError
+	} else if err != nil {
+		return nil, err
+	}
+
+	return rows, nil
+}
+
+func (lcd *largeChunkDecoder) mkError(s string) error {
+	return fmt.Errorf("corrupt chunk: %s", s)
+}
+
+func (lcd *largeChunkDecoder) decode() ([][]*string, error) {
+	if '[' != lcd.nextByteNonWhitespace() {
+		return nil, lcd.mkError("expected chunk to begin with '['")
+	}
+
+	rows := make([][]*string, 0, lcd.rows)
+	if ']' == lcd.nextByteNonWhitespace() {
+		return rows, nil // special case of an empty chunk
+	}
+	lcd.rewind(1)
+
+OuterLoop:
+	for {
+		row, err := lcd.decodeRow()
+		if err != nil {
+			return nil, err
+		}
+		rows = append(rows, row)
+
+		switch c := lcd.nextByteNonWhitespace(); {
+		case c == ',':
+			continue // more elements in the array
+		case c == ']':
+			return rows, nil // we've scanned the whole chunk
+		default:
+			break OuterLoop
+		}
+	}
+	return nil, lcd.mkError("invalid row boundary")
+}
+
+func (lcd *largeChunkDecoder) decodeRow() ([]*string, error) {
+	if '[' != lcd.nextByteNonWhitespace() {
+		return nil, lcd.mkError("expected row to begin with '['")
+	}
+
+	row := make([]*string, 0, lcd.cells)
+	if ']' == lcd.nextByteNonWhitespace() {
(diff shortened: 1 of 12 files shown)


smtakeda:
Thank for PR. Will look into this after coming back from off.

chunk_test.go
```diff
@@ -0,0 +1,145 @@
...
+	testDecodeOk(t, `[["\uD834\uDD1E"]]`) // surrogate pair
+	testDecodeOk(t, `[["\uD834\u0000"]]`) // corrupt surrogate pair
+
+	testDecodeOk(t, `[["$"]]`)      // "$"
+	testDecodeOk(t, `[["\u0024"]]`) // "$"
+
+	testDecodeOk(t, `[["\uC2A2"]]`) // "¢"
+	testDecodeOk(t, `[["¢"]]`)      // "¢"
+
+	testDecodeOk(t, `[["\u00E2\u82AC"]]`) // "€"
```
smtakeda: This is not a euro sign but two characters, right?
\u is used to escape UTF-16 code point.

chunk.go
```diff
@@ -0,0 +1,306 @@
...
+	// variable does not escape to the heap
+	c := lcd.nextByte()
+
+	switch c {
+	case '"', '\\', '/', '\'':
+		lcd.sbuf.WriteByte(c)
+	case 'b':
+		lcd.sbuf.WriteByte('\b')
+	case 'f':
+		lcd.sbuf.WriteByte('\f')
```
smtakeda: what is this for? Looks like no \f is defined in http://www.json.org/.

smtakeda:
Other than those two test cases, looks good to me.

smtakeda:
In my verification tests, the memory footprint was reduced from 2G to 600MB, while no much difference in CPU usage.

smtakeda:
hmm, for some reason, the test doesn't pass:
https://travis-ci.org/snowflakedb/gosnowflake/builds/414998093

smtakeda:
Ran TestLargeSetResult and got the log:
https://travis-ci.org/snowflakedb/gosnowflake/jobs/415002366

The cause is
```
"Client.Timeout exceeded while reading body"
```
Adjusting the buffer size, timeout, downloader goroutine...

chunk.go
mhseiden: Hm, I was basing this off of the `golang` implementation. 

https://golang.org/src/encoding/json/decode.go (line 1269)

chunk_test.go
mhseiden: It's actually the multi-byte utf-8 encoding for the euro sign. I pull this from the `wikipedia` article (along with the other currency tests).

https://en.wikipedia.org/wiki/UTF-8#Example

chunk_test.go
```diff
@@ -0,0 +1,145 @@
...
+	testDecodeOk(t, `[["€"]]`)            // "€"
+
+	testDecodeOk(t, `[["\uF090\u8D88"]]`) // "𐍈"
```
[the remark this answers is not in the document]
mhseiden: See the above comment r.e. `UTF-8` encodings of multi-byte characters.

mhseiden:
Huh, that's funky. Even when running the large result set tests on a bad network, I wasn't running into timeout issues 🤔

chunk_test.go
smtakeda: ok, that's what I thought, but JSON doesn't accept escaped UTF-8 encoding, and `\u00E2\u82AC` won't be converted to \xe2\x82\xac in bytes.

mhseiden: According to the RFC, a `UTF-8` encoding is perfectly acceptable. 
> https://tools.ietf.org/html/rfc7159#section-8.1

They also explicitly call out the possibility of a "12-byte" unicode sequence in Section 7.
> To escape an extended character that is not in the Basic Multilingual Plane, the character is represented as a 12-character sequence, encoding the UTF-16 surrogate pair.  So, for example, a string containing only the G clef character (U+1D11E) may be represented as "\uD834\uDD1E".

smtakeda:
It worked ok on my dev VM (3 virtual cpu) and MacOS (4 cores), but it is not on Travis (2 cores). I speculate the decoder walks through the response body incrementally, so it takes times to read whole data and hit timeout.

Changing the max number of chunk downloader worker from 10 to 2 and increasing the client timeout to 900 worked, but the performance significantly degraded.

I'll look into the parameters further, but in either case, if we want to have the custom decoder, I will have a feature flag to switch between the native JSON and the custom decoders for the applications to choose. At this time, I feel it is a bit risky to completely switch to the new decoder.

mhseiden:
Yea, adding a feature flag definitely makes sense. I'm happy to add that in, unless you're already working on it?

smtakeda:
Working on it.

smtakeda:
Wrapping up in https://github.com/snowflakedb/gosnowflake/pull/184

mhseiden:
Sounds great! Thx.

PR #183, opened 2018-08-13
Remove busy wait for chunk downloader

smtakeda:
### Description
Removed busy wait/loop in downloadchunk method.

ChunkDownloader waits for the result set chunks in busy wait/loop that checks the status every second. This change mitigates it by leveraging `sync.Cond`.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -239,6 +239,8 @@ func (scd *snowflakeChunkDownloader) Next() ([]*string, error) {
 				scd.ChunksMutex.Unlock()
 				return nil, err
 			}
+            // wait for chunk downloader goroutine to broadcast the event,
+            // 1) one chunk download finishes or 2) an error occurs.
 			scd.DoneDownloadCond.Wait()
 		}
 		glog.V(2).Infof("ready: chunk %v", scd.CurrentChunkIndex+1)


rows.go
```diff
@@ -229,30 +231,23 @@ func (scd *snowflakeChunkDownloader) Next() ([]*string, error) {
...
-			scd.ChunksMutex.Unlock()
-			if scd.CurrentChunk != nil {
-				ticker.Stop()
-				// kick off the next download
-				glog.V(2).Infof("ready: chunk %v", scd.CurrentChunkIndex)
-				scd.CurrentChunkSize = len(scd.CurrentChunk)
-				scd.schedule()
-				break
-			}
+			scd.DoneDownloadCond.Wait()
```
smtakeda: Wait for the downloader to broadcast the event, 1) one download finishes or 2) an error occurs.

rows.go
```diff
@@ -229,30 +231,23 @@ func (scd *snowflakeChunkDownloader) Next() ([]*string, error) {
 		if scd.CurrentChunkIndex > 1 {
 			scd.Chunks[scd.CurrentChunkIndex-1] = nil // detach the previously used chunk
 		}
-		scd.ChunksMutex.Unlock()
-
-		ticker := time.NewTicker(time.Second)
-		for range ticker.C {
-			scd.ChunksMutex.Lock()
+		for scd.Chunks[scd.CurrentChunkIndex] == nil {
```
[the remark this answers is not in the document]
smtakeda: This should be `for`, because the broadcast signal indicates other chunk's completion. For example, the main thread here waits for the chunk 1 to finish, but the broadcast signal could come from chunk 2 first, then this should be retried.

rows.go
```diff
@@ -386,5 +389,15 @@ func downloadChunkHelper(ctx context.Context, scd *snowflakeChunkDownloader, idx
 				Message:     errMsgFailedToGetChunk,
 				MessageArgs: []interface{}{idx},
 			}}
+		scd.DoneDownloadCond.Broadcast()
+		scd.ChunksMutex.Unlock()
 	}
 }
+
+func raiseDownloadError(scd *snowflakeChunkDownloader, idx int, err error) {
```
tjj5036: Does this actually raise an error? Looks like we're just storing it.

rows.go
```diff
@@ -229,30 +231,23 @@ func (scd *snowflakeChunkDownloader) Next() ([]*string, error) {
...
 			scd.Chunks[scd.CurrentChunkIndex-1] = nil // detach the previously used chunk
 		}
-		scd.ChunksMutex.Unlock()
-
-		ticker := time.NewTicker(time.Second)
-		for range ticker.C {
-			scd.ChunksMutex.Lock()
+		for scd.Chunks[scd.CurrentChunkIndex] == nil {
+			glog.V(2).Infof("waiting for chunk idx: %v/%v",
+				scd.CurrentChunkIndex+1, len(scd.ChunkMetas))
```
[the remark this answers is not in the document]
smtakeda: This converts zero-based index to one-based for logging. All other logging shows one based.

rows.go
smtakeda: yes, storing the error to an error channel and broadcasting to tell the main thread that this chunk loader failed.

rows.go
smtakeda: sure, will do.

smtakeda:
@tjj5036 let me know if you need further review. thanks.

tjj5036:
LGTM!

PR #185, opened 2018-08-21
Change timestamp to UTC one in the glog

ChTimTsubasa:
### Description
One line change in the glog's code to use UTC timestamp in log record to address SNOW-50629. Need further changes in the makefile to avoid the change being wiped.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


smtakeda:
@ChTimTsubasa let's keep this PR opened until we find the solution.

PR #186, opened 2018-08-28
Support Key Pair Authentication SNOW-52565

ChTimTsubasa:
### Description
We now support Key pair authentications. Necessary unit test and integration tests are included. Address #186 

### Checklist
- [x] Code compiles correctly
- [x] Run ``make fmt`` to fix inconsistent formats
- [x] Run ``make lint`` to get lint errors and fix all of them
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

PR #187, opened 2018-08-28
Implement JWT key pair authentication 

ChTimTsubasa:
### Description
The JWT key pair authentication is added to address SNOW-52565 and #186 

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


smtakeda:
Let's talk about this tomorrow. I know the cause. cc: @howryu 
```
=== RUN   TestJWTAuthentication
--- FAIL: TestJWTAuthentication (1.48s)
	driver_test.go:1928: error changin role: 003013 (42501): SQL access control error:
		Requested role 'ACCOUNTADMIN' is not assigned to the executing user.  Specify another role to activate.
```

auth.go
```diff
@@ -14,7 +14,14 @@ import (
 	"strings"
 	"time"
 
+	jcrypto "github.com/SermoDigital/jose/crypto"
```
smtakeda: can you please explain why you chose this library? Did you have any other alternative?

smtakeda:
Update `doc.go` to include the new authenticator usage, e.g,. how to generate a key pair and and an example code.

auth.go
```diff
@@ -26,6 +33,11 @@ const (
 	authenticatorOAuth           = "OAUTH"
 	authenticatorSnowflake       = "SNOWFLAKE"
 	authenticatorOkta            = "OKTA"
+	authenticatorJWT             = "SNOWFLAKE_JWT"
+)
+
+const (
+	timeoutJWTSeconds = 60
```
smtakeda: Can you make this a public variable so that the customers can set the validity? @howryu do you have any suggestion about the naming? I like `TimeoutJWTInSeconds`

auth_test.go
```diff
@@ -182,6 +186,38 @@ func postAuthCheckPasscodeInPassword(_ *snowflakeRestful, _ *url.Values, _ map[s
 	}, nil
 }
 
+func postAuthCheckJWTToken(_ *snowflakeRestful, _ *url.Values, _ map[string]string, jsonBody []byte, _ time.Duration) (*authResponse, error) {
+	var ar authRequest
+	if err := json.Unmarshal(jsonBody, &ar); err != nil {
+		return nil, err
+	}
+	if ar.Data.Authenticator != authenticatorJWT {
+		return nil, errors.New("Authenticator is not JWT")
```
smtakeda: can you wrap this by a SnowflakeError?

smtakeda: never mind this is a mock.

howryu:
We should create some keypair test user and pre-registered the public key with accountadmin privilege.

auth.go
howryu: I would rename it to something like `jwtExpInSeconds`. However, in the server side, we had another check. No matter the `exp` value set in the jwt payload, it will always fail the authentication if `exp` is larger than 10 seconds, which is not ideal in case of client side need to retry the request.

I would defer exposing the parameter to customer.

auth.go
ChTimTsubasa: I chose it from https://jwt.io/#libraries, where there are lot of JWT library alternatives. Among all of Golang's, this library is the one with most star while still has all features listed included. There is one with more star but does not support issuer check and some other function.

auth.go
ChTimTsubasa: Sure. I saw the timeout is hardcoded as 60 second in JDBC so I chose this value. We can definitely exposed this as a settable parameter with default as 60.

driver_test.go
```diff
@@ -78,6 +92,9 @@ func createDSN(timezone string) {
 	if rolename != "" {
 		parameters.Add("role", rolename)
 	}
+
+	parameters.Add("privateKey", base64.RawStdEncoding.EncodeToString(x509.MarshalPKCS1PrivateKey(TestPrivateKey)))
```
howryu: We should avoid using PKCS1 private key. We should use PKCS8. You should add another sanity check to reject PKCS1 format of private key if found.

doc.go
```diff
@@ -278,6 +278,43 @@ This option will reduce the memory footprint to half or even quarter, but it can
...
+	openssl genpkey -algorithm RSA \
+    	-pkeyopt rsa_keygen_bits:2048 \
+    	-pkeyopt rsa_keygen_pubexp:65537 | \
+  		openssl pkcs8 -topk8 -nocrypt -outform der > rsa-2048-private-key.p8
+
+	# extravt 2048-bit PKI encoded RSA public key from the private key
+	openssl pkey -pubout -inform der -outform der \
+    	-in rsa-2048-private-key.p8 \
+    	-out rsa-2048-public-key.spki
+
```
smtakeda: +1

smtakeda:
@ChTimTsubasa please rebase origin/master and push the change again:
```
git rebase origin/master
```

ChTimTsubasa:
@smtakeda I have actually already merged the origin/master to this one so that I do not need to force push to to this pull request.

BTW. Not sure if you have seen this but we were failing some of the configured test environment given that its running time was too long so that travis killed it. When I restarted, it passed. Could it be a connectivity issue or the server is crushed by the testcases?

ChTimTsubasa:
Let me run the tests again and see if the crush can still happen. Could be just connectivity issue between travis and GS, because the history on GS shows that the query that was canceled had ran for only 1.5s. And the two failing environment are not failing at the same test cases.

smtakeda:
@ChTimTsubasa Travis has 10 min timeout. If one test hits it, it fails. Which test case cause the issue?

See below merge status. "This branch cannot be rebased due to conflicts". So it means there are conflicts with master.

ChTimTsubasa:
They are failing on different ones. One is Large Result set, another is something like SleepSession? Right now we are failing on different one. Guess we are just failing because test cases are running too long.

PR #193, opened 2018-10-04
add a configurable retry limit

mhseiden:
### Description
We have observed cases where goroutines end up in an infinite retry loop, likely because a `context` was used that never times out. While this may sometimes be a user error, there are also a few places in this codebase where `context.Background()` or `context.TODO()` is used when issuing requests. As such, this patch is a stop-gap to ensure that _all_ retry attempts eventually terminate. 

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -16,6 +16,11 @@ import (
 	"sync"
 )
 
+var (
+	// MaxRetryHTTPCount specifies the maximum number of times an HTTP request may be retried
+	MaxRetryHTTPCount = 3
+)
+
 var random *rand.Rand
 
 func init() {
@@ -96,6 +101,12 @@ func retryHTTP(
 		}
 		// cannot just return 4xx and 5xx status as the error can be sporadic. retry often helps.
 		if err != nil {
+			// bound the number of retry attempts so we don't loop forever
+			if retryCounter > MaxRetryHTTPCount {
+				glog.V(2).Infof(
+					"failed http connection. no response is returned. err: %v. retry limit exceeded\n", err)
+				break
+			}
 			glog.V(2).Infof(
 				"failed http connection. no response is returned. err: %v. retrying...\n", err)
 		} else {


smtakeda:
To me, the max retry count 3 is too optimistic. Retry often works well when the server is flaky for many reasons (we are improving to mitigate it).

For all drivers/connectors, we use timeouts instead of retry limit. Here are two main timeout parameters:

- `LoginTimeout` is, by default, 60 seconds, for login request. It can be changed to shorter if the application/user wants to give up earlier (or longer for resiliency)
- `RequestTimeout` mainly used for query request. It is indefinite by default, because the network issue is very likely transient as login request is already success. It can also be changed to a finite timing as well.

Would not those timeout work?

ChTimTsubasa:
Agree with @smtakeda .

Also such hardcoded global variable like MaxRetryHTTPCount is generally not recommended.

mhseiden:
@ChTimTsubasa it could be moved to a Config variable; that seems like an implementation detail.

@smtakeda this is largely to mitigate the infinite loop caused by the the default for `RequestTimeout` - we found that this driver was leaking goroutines which would never terminate. I agree that `RequestTimeout` can be tuned to avoid this issue, however it took a fair amount of reading through source code to understand the root cause of the leaks. If you aren't keen on adding a retry limit, then I'd suggest using a bounded default - perhaps the same 900s used for the `http.Client` timeout?

mhseiden:
At a minimum, it seems like change the variable names to `RequestRetryTimeout` and `LoginRetryTimeout` would make it more clear to the reader what the behavior is.

ChTimTsubasa:
We would add more documentations on the behavior of these variables and how to set them. Having a bounded request timeout is reasonable, but the concrete number requires more investigation, since the for OLAP queries it could be potentially very long running.

mhseiden:
If I understand the code correctly, this timeout has no relationship to query timeouts. It’s only a retry timeout. 

As a user of the driver, it would be somewhat unexpected behavior if a 3hr query ran for 2hr, failed in an unexpected way, was retried, and then completed in 3hr. My expectation regarding retries is that they smooth out network instability when establishing a request, rather than retrying a (potentially destructive) long-running operation.

ChTimTsubasa:
@mhseiden You are right that there is no query timeout controlled by the driver side. However the retry timeout would somehow affected how query timeout. There is a new PR can potentially fix your go-routine leakage issue.

smtakeda:
See #194

PR #194, opened 2018-10-04
Timeout fix

ChTimTsubasa:
### Description
This PR intentions to address requirements presented in PR#193 including: 
1. Address documentations for time out explain
2. Enable configuration for RequestTimeout for result polling. Previously if we success on issuing a query but fail on one of the result chunk polling, the driver would retry forever even if the customer set the `requestTimeout` to be some finite number.

It has also shorten the network Client timeout.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/restful.go
+++ b/restful.go
@@ -190,7 +190,7 @@ func postRestfulQueryHelper(
 			fullURL := fmt.Sprintf(
 				"%s://%s:%d%s", sr.Protocol, sr.Host, sr.Port, resultURL)
 
-			resp, err = sr.FuncGet(ctx, sr, fullURL, headers, 0)
+			resp, err = sr.FuncGet(ctx, sr, fullURL, headers, timeout)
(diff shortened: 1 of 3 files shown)


mhseiden:
This seems fine.

restful.go
```diff
@@ -190,7 +190,7 @@ func postRestfulQueryHelper(
 			fullURL := fmt.Sprintf(
 				"%s://%s:%d%s", sr.Protocol, sr.Host, sr.Port, resultURL)
 
-			resp, err = sr.FuncGet(ctx, sr, fullURL, headers, 0)
+			resp, err = sr.FuncGet(ctx, sr, fullURL, headers, timeout)
```
smtakeda: Thanks for this fix!

ChTimTsubasa:
I would merge this when the tests pass.

PR #195, opened 2018-10-05
Implement retryIDReplacer and add test cases.

ChTimTsubasa:
### Description
For debugging reason, we want to generate an retry id every time we resend to server

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -6,9 +6,11 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/google/uuid"
 	"io"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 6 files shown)


retry.go
```diff
@@ -22,6 +24,62 @@ func init() {
 	random = rand.New(rand.NewSource(time.Now().UnixNano()))
 }
 
+const retryKey string = "request_guid"
```
smtakeda: can you change the name to `equestGUID`
retry key is a separate one and not part of this change (and only JDBC has the retry parameter...)

PR #199, opened 2018-10-09
do not sleep through a context timeout

mhseiden:
### Description
Rather than unconditionally sleeping during retry, the wait code should select on 1) the timer and 2) the context. This way a large `sleepTime` doesn't interfere with a higher-level context timeout.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -198,7 +198,15 @@ func retryHTTP(
 		}
 		fullURL = rIDReplacer.replace()
 		glog.V(2).Infof("sleeping %v. to timeout: %v. retrying", sleepTime, totalTimeout)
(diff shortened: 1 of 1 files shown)


retry.go
```diff
@@ -198,7 +198,15 @@ func retryHTTP(
...
-		time.Sleep(sleepTime)
+
+		await := time.NewTimer(sleepTime)
+		select {
+		case <-await.C:
+			// retry the request
+		case <-ctx.Done():
+			await.Stop()
+			return res, ctx.Err()
+		}
```
smtakeda: yeah, this makes sense.

PR #210, opened 2019-01-10
Fix Custom Json Parser

### LINKED ISSUE #209: CustomJSONDecoder missing millions of rows
### Issue description

I had performance issues (well, memory issues) with a database that had 5 million rows (memory usage ballooned to ~2GB), so I turned on `CustomJSONDecoderEnabled` and set `MaxChunkDownloadWorkers = 4` (not sure if that's a reasonable number or not).

Instead of getting 5 million rows back (`rows.Next()` from sqlx), I only got around 1.5 million.

### Example code

This is two or three snippets of code pasted together, but it should include all relevant pieces.

```go
	sf.CustomJSONDecoderEnabled = false
	sf.MaxChunkDownloadWorkers = 2

	db, err := sqlx.Open("snowflake", dburlstr)
	if err != nil {
		return nil, err
	}
	db.SetMaxIdleConns(5)
	db.SetMaxOpenConns(15)
	db.SetConnMaxLifetime(10 * time.Minute)

	defer db.Close()

	// remove SQLi security checks for brevity

	rows, err := db.Queryx(fmt.Sprintf(`SELECT * FROM %s`, id))
	if err != nil {
		return
	}

	// Write the CSV header
	colnames, err := rows.Columns()
	if err != nil {
		return
	}
	csvw := csv.NewWriter(out)
	err = csvw.Write(colnames)
	if err != nil {
		return
	}

	// Write each CSV row (as type string)
	numcols := len(colnames)
	rowstr := make([]string, numcols)
	for rows.Next() {
		var row []interface{}
		row, err = rows.SliceScan()
		if nil != err {
			return
		}
		for i, j := range row {
			switch v := j.(type) {
			case string:
				rowstr[i] = v
			case fmt.Stringer:
				rowstr[i] = v.String()
			default:
				rowstr[i] = fmt.Sprintf("%v", v)
			}
		}
		err = csvw.Write(rowstr)
		if nil != err {
			// must close rows if not read to completion as per sqlx docs
			rows.Close()
			return
		}
	}
	csvw.Flush()
```

### Error log

No errors that I'm aware of.

No rows are missing when run without the memory efficient JSON decoder.

### Configuration
*Driver version (or git SHA):* v1.1.15

*Go version:* go1.11.4

mhseiden:
### Description
This fixes a logic bug in the custom json parser's buffer management code. It also includes a test case to expose the original bug.

AFAIK this fixes #209, since we hit a similar issue when downloading a large number of rows, patched our fork, and saw the issue go away. 

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/chunk.go
+++ b/chunk.go
@@ -285,6 +285,9 @@ func (lcd *largeChunkDecoder) nextByte() byte {
 func (lcd *largeChunkDecoder) ensureBytes(n int) {
 	if lcd.rem <= n {
 		rbuf := make([]byte, defaultChunkBufferSize)
+		// NOTE when the buffer reads from the stream, there's no
+		// guarantee that it will actually be filled. As such we
+		// must use (ptr+rem) to compute the end of the slice.
 		off := copy(rbuf, lcd.rbuf[lcd.ptr:lcd.ptr+lcd.rem])
 		add := lcd.fillBuffer(rbuf[off:])
 
--- a/chunk.go
+++ b/chunk.go
@@ -285,8 +285,8 @@ func (lcd *largeChunkDecoder) nextByte() byte {
 func (lcd *largeChunkDecoder) ensureBytes(n int) {
 	if lcd.rem <= n {
 		rbuf := make([]byte, defaultChunkBufferSize)
-		copy(rbuf, lcd.rbuf[lcd.ptr:])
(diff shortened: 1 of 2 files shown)


coolaj86:
Thanks!

coolaj86:
Would you please tag a release for this?

PR #229, opened 2019-05-30
Add test case to address ignoring Session Gone

ChTimTsubasa:
### Description
We already ignore whatever error returned from CloseSession(). Add a mock test to address it.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -173,11 +173,6 @@ func (sc *snowflakeConn) Close() (err error) {
 	glog.V(2).Infoln("Close")
 	sc.stopHeartBeat()
 
-	// ensure transaction is rollbacked
-	_, err = sc.exec(context.Background(), "ROLLBACK", false, false, nil)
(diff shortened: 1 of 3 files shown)


connection.go
```diff
@@ -173,11 +173,6 @@ func (sc *snowflakeConn) Close() (err error) {
 	glog.V(2).Infoln("Close")
 	sc.stopHeartBeat()
 
-	// ensure transaction is rollbacked
```
ChTimTsubasa: all transactions would automatically rollback when the session is closed.

smtakeda: hmm, I don't think the transaction is rollbacked automatically.

smtakeda: Was the behavior changed?

ChTimTsubasa: This is based on an conversation with Ashish. Let me add a test case to see.

PR #232, opened 2019-06-09
Snow 76958

smtakeda:
### Description
- Added OCSP fail open mode
- Added OCSP Cache server support
- Replaced megacheck with staticcheck and fixed the errors

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## Version 1.1.18
+- Ignore session gone error when closing session.
+- Set the latest Go version to 1.12
+
+## Version 1.1.17
+- TIMESTAMP_LTZ is not converted properly
+- Add SERVICE_NAME support to Golang
+
+## Version 1.1.16
+- Fix custom json parser (@mhseiden)
+- Modify Support link on Gosnowflake github
+
 ## Version 1.1.15
 - Perform check on error for pingpong response (@ChTimTsubasa)
 - Add new 1.11.x for testing (@ChTimTsubasa)
--- a/gosnowflake.mak
+++ b/gosnowflake.mak
@@ -3,18 +3,18 @@ SHELL := /bin/bash
 SRC = $(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
 setup:
-	go get -u github.com/golang/dep/cmd/dep
-	go get golang.org/x/lint/golint
-	go get github.com/Songmu/make2help/cmd/make2help
-	[[ $$(go version | awk '{print $3}' | cut -d'.' -f 2) != "8" ]] && go get honnef.co/go/tools/cmd/megacheck || true
+	@which dep &> /dev/null || go get -u github.com/golang/dep/cmd/dep
+	@which golint &> /dev/null  || go get golang.org/x/lint/golint
+	@which make2help &> /dev/null || go get github.com/Songmu/make2help/cmd/make2help
+	@which staticcheck &> /dev/null || go get honnef.co/go/tools/cmd/staticcheck
 
 ## Install dependencies
 deps: setup
-	dep ensure
+	@dep ensure
 
 ## Update dependencies
 update: setup
-	dep ensure -update
+	@dep ensure -update
 
 ## Show help
 help:
@@ -22,12 +22,13 @@ help:
 
 # Format source codes (internally used)
 cfmt: setup
-	gofmt -l -w $(SRC)
+	@gofmt -l -w $(SRC)
 
 # Lint (internally used)
 clint: setup
-	[[ $$(go version | awk '{print $3}' | cut -d'.' -f 2) != "8" ]] && echo "Running megacheck" && megacheck || echo "No megacheck run, because Go1.8 is not supported."
-	for pkg in $$(go list ./... | grep -v /vendor/); do \
+	@echo "Running staticcheck" && staticcheck || echo "No staticcheck run, because Go1.8 is not supported."
+	@echo "Running go vet and lint"
+	@for pkg in $$(go list ./... | grep -v /vendor/); do \
 		echo "Verifying $$pkg"; \
 		go vet $$pkg; \
 		golint -set_exit_status $$pkg || exit $$?; \
@@ -35,7 +36,7 @@ clint: setup
 
 # Install (internally used)
 cinstall:
-	export GOBIN=$$GOPATH/bin; \
+	@export GOBIN=$$GOPATH/bin; \
 	go install -tags=sfdebug $(CMD_TARGET).go
 
 # Run (internally used)
--- a/ocsp.go
+++ b/ocsp.go
@@ -3,6 +3,7 @@
 package gosnowflake
 
 import (
+	"bufio"
 	"context"
 	"crypto"
 	"crypto/tls"
@@ -12,7 +13,10 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
+	"golang.org/x/crypto/ocsp"
+	"io"
 	"io/ioutil"
 	"math/big"
 	"net/http"
@@ -20,10 +24,9 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
-
-	"golang.org/x/crypto/ocsp"
 )
 
 // caRoot includes the CA certificates.
@@ -38,16 +41,35 @@ var cacheDir = ""
 // cacheFileName is the file name of OCSP response cache file
 var cacheFileName = ""
 
+// cacheUpdated is true if the memory cache is updated
+var cacheUpdated = true
+
+const (
+	ocspFailOpenNotSet = 0
+	ocspFailOpenTrue   = 1
+	ocspFailOpenFalse  = 2
+)
+const (
+	ocspModeFailOpen   = "FAIL_OPEN"
+	ocspModeFailClosed = "FAIL_CLOSED"
+	ocspModeInsecure   = "INSECURE"
+)
+
+// OCSP fail open mode
+var ocspFailOpen = ocspFailOpenTrue
+
 const (
 	// retryOCSPTimeout is the total timeout for OCSP checks.
-	retryOCSPTimeout     = 120 * time.Second
-	retryOCSPHTTPTimeout = 30 * time.Second
+	retryOCSPTimeout     = 60 * time.Second
+	retryOCSPHTTPTimeout = 20 * time.Second
 )
 
 const (
-	cacheFileBaseName = "ocsp_response_cache"
+	cacheFileBaseName = "ocsp_response_cache.json"
 	// cacheExpire specifies cache data expiration time in seconds.
-	cacheExpire = float64(24 * 60 * 60)
+	cacheExpire           = float64(24 * 60 * 60)
+	cacheServerURL        = "http://ocsp.snowflakecomputing.com"
+	cacheServerEnabledEnv = "SF_OCSP_RESPONSE_CACHE_SERVER_ENABLED"
 )
 
 const (
@@ -63,25 +85,43 @@ type ocspStatus struct {
 }
 
 const (
-	ocspSuccess               ocspStatusCode = 0
-	ocspNoServer              ocspStatusCode = -1
-	ocspFailedParseOCSPHost   ocspStatusCode = -2
-	ocspFailedComposeRequest  ocspStatusCode = -3
-	ocspFailedDecomposeRequst ocspStatusCode = -4
-	ocspFailedEncodeCertID    ocspStatusCode = -5
-	ocspFailedSubmit          ocspStatusCode = -6
-	ocspFailedResponse        ocspStatusCode = -7
-	ocspFailedExtractResponse ocspStatusCode = -8
-	ocspFailedParseResponse   ocspStatusCode = -9
-	ocspInvalidValidity       ocspStatusCode = -10
-	ocspRevokedOrUnknown      ocspStatusCode = -11
-	ocspMissedCache           ocspStatusCode = -12
-	ocspCacheExpired          ocspStatusCode = -13
-	ocspFailedDecodeResponse  ocspStatusCode = -14
+	ocspSuccess                ocspStatusCode = 0
+	ocspStatusGood             ocspStatusCode = -1
+	ocspStatusRevoked          ocspStatusCode = -2
+	ocspStatusUnknown          ocspStatusCode = -3
+	ocspStatusOthers           ocspStatusCode = -4
+	ocspNoServer               ocspStatusCode = -5
+	ocspFailedParseOCSPHost    ocspStatusCode = -6
+	ocspFailedComposeRequest   ocspStatusCode = -7
+	ocspFailedDecomposeRequest ocspStatusCode = -8
+	ocspFailedSubmit           ocspStatusCode = -9
+	ocspFailedResponse         ocspStatusCode = -10
+	ocspFailedExtractResponse  ocspStatusCode = -11
+	ocspFailedParseResponse    ocspStatusCode = -12
+	ocspInvalidValidity        ocspStatusCode = -13
+	ocspMissedCache            ocspStatusCode = -14
+	ocspCacheExpired           ocspStatusCode = -15
+	ocspFailedDecodeResponse   ocspStatusCode = -16
 )
 
+// copied from crypto/ocsp.go
+type certID struct {
+	HashAlgorithm pkix.AlgorithmIdentifier
+	NameHash      []byte
+	IssuerKeyHash []byte
+	SerialNumber  *big.Int
+}
+
+// cache key
+type certIDKey struct {
+	HashAlgorithm crypto.Hash
+	NameHash      string
+	IssuerKeyHash string
+	SerialNumber  string
+}
+
 var (
-	ocspResponseCache     map[string][]interface{}
+	ocspResponseCache     map[certIDKey][]interface{}
 	ocspResponseCacheLock *sync.RWMutex
 )
 
@@ -93,15 +133,7 @@ var hashOIDs = map[crypto.Hash]asn1.ObjectIdentifier{
 	crypto.SHA512: asn1.ObjectIdentifier([]int{2, 16, 840, 1, 101, 3, 4, 2, 3}),
 }
 
-// copied from crypto/ocsp.go
-type certID struct {
-	HashAlgorithm pkix.AlgorithmIdentifier
-	NameHash      []byte
-	IssuerKeyHash []byte
-	SerialNumber  *big.Int
-}
-
-// copied from crypto/ocsp.go
+// copied from crypto/ocsp
 func getOIDFromHashAlgorithm(target crypto.Hash) asn1.ObjectIdentifier {
 	for hash, oid := range hashOIDs {
 		if hash == target {
@@ -110,6 +142,16 @@ func getOIDFromHashAlgorithm(target crypto.Hash) asn1.ObjectIdentifier {
 	}
 	glog.V(0).Infof("no valid OID is found for the hash algorithm. %#v", target)
 	return nil
+}
+
+func getHashAlgorithmFromOID(target pkix.AlgorithmIdentifier) crypto.Hash {
+	for hash, oid := range hashOIDs {
+		if oid.Equal(target.Algorithm) {
+			return hash
+		}
+	}
+	glog.V(0).Infof("no valid hash algorithm is found for the oid. Falling back to SHA1: %#v", target)
+	return crypto.SHA1
 }
 
 // calcTolerableValidity returns the maximum validity buffer
@@ -128,7 +170,7 @@ func isInValidityRange(currTime, thisUpdate, nextUpdate time.Time) bool {
 	return true
 }
 
-func retryRevocationStatusCheck(totalTimeout *time.Duration, sleepTime time.Duration) (ok bool) {
+func checkTotalTimeout(totalTimeout *time.Duration, sleepTime time.Duration) (ok bool) {
 	if *totalTimeout > 0 {
 		*totalTimeout -= sleepTime
 	}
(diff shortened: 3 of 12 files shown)


smtakeda:
@h-chaturvedi FYI, this is OCSP change for Golang driver.

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+	return r.Status
 }
 
 // writeOCSPCacheFile writes a OCSP Response cache file. This is called if all revocation status is success.
 // lock file is used to mitigate race condition with other process.
 func writeOCSPCacheFile() {
 	glog.V(2).Infof("writing OCSP Response cache file. %v\n", cacheFileName)
 	cacheLockFileName := cacheFileName + ".lck"
-	statinfo, err := os.Stat(cacheLockFileName)
+	err := os.Mkdir(cacheLockFileName, 0600)
```
smtakeda: use a directory for lock file

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
 			verifiedChains[i] = append(verifiedChains[i], rca)
 			n++
 		}
-		wg.Add(n)
-		results := callback(&wg, verifiedChains[i])
-		wg.Wait()
+		results := getAllRevocationStatus(verifiedChains[i])
 		for _, r := range results {
-			if r.err != nil {
+			if r.err != nil && (r.code == ocspRevoked || ocspFailOpen == ocspFailOpenFalse) {
```
smtakeda: OCSP Fail open mode check.
Except Revoked status, all other errors will be ignored when ocspFailOpen==ocspFailOpenTrue

ocsp.go
```diff
@@ -567,30 +786,19 @@ func createOCSPCacheDir() {
...
 	Proxy:           http.ProxyFromEnvironment,
 }
 
-// SnowflakeTransport includes the certificate revocation check with OCSP in parallel. By default, the driver uses
+// SnowflakeTransport includes the certificate revocation check with OCSP in sequential. By default, the driver uses
 // this transport object.
 var SnowflakeTransport = &http.Transport{
-	TLSClientConfig: &tls.Config{
-		RootCAs:               certPool,
-		VerifyPeerCertificate: verifyPeerCertificateParallel,
```
smtakeda: Parallel check was removed to simplify the code. No performance advantage was observed in my experiments.

ocsp.go
```diff
@@ -567,30 +786,19 @@ func createOCSPCacheDir() {
 func init() {
 	readCACerts()
 	createOCSPCacheDir()
-	readOCSPCacheFile()
+	initOCSPCache()
```
smtakeda: read a cache file once when a process starts.

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+	cacheUpdated = true
+	ocspResponseCacheLock.Unlock()
 }
 
-func getAllRevocationStatusSerial(wg *sync.WaitGroup, verifiedChains []*x509.Certificate) []*ocspStatus {
+func getAllRevocationStatus(verifiedChains []*x509.Certificate) []*ocspStatus {
+	cached := validateWithCacheForAllCertificates(verifiedChains)
+	if !cached {
+		downloadOCSPCacheServer()
+	}
```
smtakeda: validate the status using cache first. If not cached, download the cache from the OCSP response cache server.

gosnowflake.mak
```diff
@@ -3,39 +3,40 @@ SHELL := /bin/bash
...
 
 setup:
-	go get -u github.com/golang/dep/cmd/dep
-	go get golang.org/x/lint/golint
-	go get github.com/Songmu/make2help/cmd/make2help
-	[[ $$(go version | awk '{print $3}' | cut -d'.' -f 2) != "8" ]] && go get honnef.co/go/tools/cmd/megacheck || true
+	@which dep &> /dev/null || go get -u github.com/golang/dep/cmd/dep
+	@which golint &> /dev/null  || go get golang.org/x/lint/golint
+	@which make2help &> /dev/null || go get github.com/Songmu/make2help/cmd/make2help
+	@which staticcheck &> /dev/null || go get honnef.co/go/tools/cmd/staticcheck
```
smtakeda: This change significantly improved `fmt` and `lint` performance.

ocsp.go
```diff
@@ -38,16 +40,24 @@ var cacheDir = ""
...
 	// retryOCSPTimeout is the total timeout for OCSP checks.
-	retryOCSPTimeout     = 120 * time.Second
-	retryOCSPHTTPTimeout = 30 * time.Second
+	retryOCSPTimeout     = 60 * time.Second
+	retryOCSPHTTPTimeout = 20 * time.Second
 )
 
 const (
-	cacheFileBaseName = "ocsp_response_cache"
+	cacheFileBaseName = "ocsp_response_cache.json"
```
smtakeda: This change allows the Golang driver to share the cache file with other Snowflake drivers.

ocsp.go
```diff
@@ -63,25 +73,41 @@ type ocspStatus struct {
...
+	SerialNumber  *big.Int
+}
+
+// cache key
+type certIDKey struct {
+	HashAlgorithm crypto.Hash
+	NameHash      string
+	IssuerKeyHash string
+	SerialNumber  string
+}
```
smtakeda: Changed the cache key from a base64 encoded certID to this so that each element of key can be compared.

smtakeda:
Target release is July 1st. We'll make pre-announcement of this change.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## Version 1.1.18
```
ChTimTsubasa: Thanks

ocsp.go
```diff
@@ -140,117 +168,209 @@ func retryRevocationStatusCheck(totalTimeout *time.Duration, sleepTime time.Dura
...
+	switch ocspRes.Status {
+	case ocsp.Good:
 		return &ocspStatus{
-			code: ocspRevokedOrUnknown,
-			err:  fmt.Errorf("bad revocation status. %v: %v, cert: %v", ocspRes.Status, ocspRes.RevocationReason, subject.Subject),
+			code: ocspSuccess,
+			err:  nil,
+		}
+	case ocsp.Revoked:
+		deleteOCSPCache(encodedCertID)
```
h-chaturvedi: We should not be deleting an OCSP Response with revoked status.

ocsp.go
```diff
@@ -382,21 +490,21 @@ func getRevocationStatus(wg *sync.WaitGroup, ocspStatusChan chan<- *ocspStatus,
...
+		return ocspS
 	}
-	encodedCertIDBase64 := base64.StdEncoding.EncodeToString(encodedCertID)
-	ocspStatusChan <- validateOCSP(encodedCertIDBase64, ocspRes, subject)
+
+	ret := validateOCSP(encodedCertID, ocspRes, subject)
 	v := []interface{}{float64(time.Now().UTC().Unix()), base64.StdEncoding.EncodeToString(ocspResBytes)}
 	ocspResponseCacheLock.Lock()
-	ocspResponseCache[encodedCertIDBase64] = v
+	ocspResponseCache[*encodedCertID] = v
```
h-chaturvedi: We should ideally check the ret value before updating the cache with the response. If the response is invalid then in this case we would be updating the cache with an invalid entry.

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+		downloadOCSPCacheServer()
+	}
 	n := len(verifiedChains) - 1
 	results := make([]*ocspStatus, n)
 	for j := 0; j < n; j++ {
-		ocspStatusChan := make(chan *ocspStatus, 1)
-		getRevocationStatus(wg, ocspStatusChan, verifiedChains[j], verifiedChains[j+1])
-		results[j] = <-ocspStatusChan
-		close(ocspStatusChan)
+		results[j] = getRevocationStatus(verifiedChains[j], verifiedChains[j+1])
```
h-chaturvedi: Should we be looking at all the certs if one has returned an error? We can potentially break out of the loop if any one of the certs returned an invalid / revoked / unknown status.

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+func validateWithCache(subject, issuer *x509.Certificate) (bool, []byte, *certIDKey) {
+	ocspReq, err := ocsp.CreateRequest(subject, issuer, &ocsp.RequestOptions{})
+	if err != nil {
+		glog.V(2).Infof("failed to create OCSP request from the certificates.\n")
+		return false, nil, nil
 	}
-	close(ocspStatusChan)
-	return results
+	encodedCertID, ocspS := extractCertIDKeyFromRequest(ocspReq)
+	if ocspS.code != ocspSuccess {
```
h-chaturvedi: nit: using ocspSuccess as error codes can be a misnomer and confuse someone about the functionality of the function. This function does not process ocsp status and hence should not ideally have return value named ocspSuccess. More generic error code names like "success" "failure" would be better.

ocsp.go
```diff
@@ -140,117 +168,209 @@ func retryRevocationStatusCheck(totalTimeout *time.Duration, sleepTime time.Dura
...
 		return &ocspStatus{
 			code: ocspMissedCache,
-			err:  fmt.Errorf("miss cache data. CertID: %v", encodedCertIDBase64),
+			err:  fmt.Errorf("miss cache data. CertID: %v", encodedCertID),
 		}
 	}
-	glog.V(2).Infof("hit cache: %v, time: %v, value: %v", encodedCertIDBase64, gotValueFromCache[0], gotValueFromCache[1])
+	glog.V(2).Infof("hit cache: %v, time: %v, value: %v", encodedCertID, gotValueFromCache[0], gotValueFromCache[1])
 	currentTime := float64(time.Now().UTC().Unix())
 	if epoch, ok := gotValueFromCache[0].(float64); ok {
```
h-chaturvedi: Can we change the name from epoch to something else? Like timestamp/ ts? Epoch is confusing as it usually would mean the Unix epoch time.

ocsp.go
```diff
@@ -140,117 +168,209 @@ func retryRevocationStatusCheck(totalTimeout *time.Duration, sleepTime time.Dura
...
+			err:  nil,
+		}
+	case ocsp.Revoked:
+		deleteOCSPCache(encodedCertID)
+		return &ocspStatus{
+			code: ocspRevoked,
+			err:  fmt.Errorf("revoked status: reason:%v, at:%v cert: %v", ocspRes.RevocationReason, ocspRes.RevokedAt, subject.Subject),
+		}
+	default:
+		deleteOCSPCache(encodedCertID)
```
h-chaturvedi: The OCSP Responses with Unknown cert status should also be allowed to stay in the cache till they expire on their own.

ocsp.go
```diff
@@ -407,101 +515,212 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+		cacheKey := encodeCertIDKey(k)
+		ocspResponseCache[*cacheKey] = cacheValue
+	}
+	cacheUpdated = true
+	ocspResponseCacheLock.Unlock()
 }
 
-func getAllRevocationStatusSerial(wg *sync.WaitGroup, verifiedChains []*x509.Certificate) []*ocspStatus {
+func getAllRevocationStatus(verifiedChains []*x509.Certificate) []*ocspStatus {
+	cached := validateWithCacheForAllCertificates(verifiedChains)
```
h-chaturvedi: currently validateWithCacheForAllCertificates will return False for revoked/unknown certs which will lead the driver in downloading a new cache even though the cache already had a good / valid response. Either the function should not return False for those two conditions or this line should take in more context to decide whether new cache has to be downloaded.

ocsp.go
```diff
@@ -327,53 +447,41 @@ func retryOCSP(
...
-	ocspValidatedWithCache := checkOCSPResponseCache(encodedCertID, subject, issuer)
-	if ocspValidatedWithCache.code == ocspSuccess {
-		ocspStatusChan <- ocspValidatedWithCache
-		return
-	}
-	glog.V(2).Infof("cache missed: %v\n", ocspValidatedWithCache.err)
-
-	st := snowflakeInsecureTransport
 	ocspClient := &http.Client{
 		Timeout:   30 * time.Second,
```
h-chaturvedi: is 30 seconds too much?

ocsp.go
smtakeda: you are right. removing delete cache

ocsp.go
smtakeda: ok. will do.

ocsp.go
smtakeda: ok

ocsp.go
smtakeda: Agree. Separating function is success and revocation status good cases.

ocsp.go
smtakeda: ok. it is unix epoch time, though, ts is fine to me.

ocsp.go
smtakeda: ok

ocsp.go
smtakeda: will fix by separating the status of success and status good.

ocsp.go
smtakeda: updating.

ocsp.go
```diff
@@ -63,25 +73,41 @@ type ocspStatus struct {
...
+	ocspFailedResponse         ocspStatusCode = -8
+	ocspFailedExtractResponse  ocspStatusCode = -9
+	ocspFailedParseResponse    ocspStatusCode = -10
+	ocspInvalidValidity        ocspStatusCode = -11
+	ocspMissedCache            ocspStatusCode = -12
+	ocspCacheExpired           ocspStatusCode = -13
+	ocspFailedDecodeResponse   ocspStatusCode = -14
 )
 
+// copied from crypto/sfocsp.go
```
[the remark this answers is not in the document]
smtakeda: oh, I should fix this.

ocsp_test.go
```diff
@@ -250,7 +251,17 @@ func TestUnitValidateOCSP(t *testing.T) {
 	ocspRes.Status = ocsp.Good
 	ost = validateOCSP(ocspRes)
 	if ost.code != ocspStatusGood {
-		t.Fatalf("should have success. expected: %v, got: %v", ocspSuccess, ost.code)
+		t.Fatalf("should have success. expected: %v, got: %v", ocspStatusGood, ost.code)
+	}
+	ocspRes.Status = ocsp.Unknown
```
smtakeda: OCSP unknown status test

ocsp_test.go
```diff
@@ -250,7 +251,17 @@ func TestUnitValidateOCSP(t *testing.T) {
...
+	ost = validateOCSP(ocspRes)
+	if ost.code != ocspStatusUnknown {
+		t.Fatalf("should have failed. expected: %v, got: %v", ocspStatusUnknown, ost.code)
+	}
+	ocspRes.Status = ocsp.ServerFailed
```
smtakeda: OCSP all other status test. This is taken as invalid OCSP response, which will be ignored in Fail Open mode.

ocsp_test.go
```diff
@@ -315,3 +326,89 @@ func TestOCSPRetry(t *testing.T) {
...
+}
+
+func TestCanEarlyExitForOCSP(t *testing.T) {
+	testcases := []tcCanEarlyExit{
+		{ // 0
+			status: &ocspStatus{
+				code: ocspStatusGood,
+			},
+			retFailOpen:   false,
+			retFailClosed: false,
```
smtakeda: GOOD won't exit early when checking results.

ocsp_test.go
```diff
@@ -315,3 +326,89 @@ func TestOCSPRetry(t *testing.T) {
...
+		},
+		{ // 1
+			status: &ocspStatus{
+				code: ocspStatusRevoked,
+				err:  errors.New("revoked"),
+			},
+			retFailOpen:   true,
+			retFailClosed: true,
```
smtakeda: REVOKED will exit.

ocsp_test.go
```diff
@@ -315,3 +326,89 @@ func TestOCSPRetry(t *testing.T) {
...
+		},
+		{ // 2
+			status: &ocspStatus{
+				code: ocspStatusUnknown,
+				err:  errors.New("unknown"),
+			},
+			retFailOpen:   false,
+			retFailClosed: true,
```
smtakeda: UNKNOWN won't exit in fail open mode, otherwise exit to raise an error.

ocsp_test.go
```diff
@@ -315,3 +326,89 @@ func TestOCSPRetry(t *testing.T) {
...
+		},
+		{ // 3
+			status: &ocspStatus{
+				code: ocspFailedParseResponse,
+				err:  errors.New("failed to parse response"),
+			},
+			retFailOpen:   false,
+			retFailClosed: true,
```
smtakeda: Invalid OCSP response won't exit in fail open mode.

ocsp.go
```diff
@@ -553,7 +578,10 @@ func downloadOCSPCacheServer() {
 	}
 	ocspURL := fmt.Sprintf("%v/%v", cacheServerURL, cacheFileBaseName)
 	glog.V(2).Infof("downloading OCSP Cache from server %v", ocspURL)
-	ret := retryOCSPCacheServer(ocspClient, http.NewRequest, ocspURL, retryOCSPTimeout, retryOCSPHTTPTimeout)
+	ret, ocspStatus := retryOCSPCacheServer(ocspClient, http.NewRequest, ocspURL, retryOCSPTimeout, retryOCSPHTTPTimeout)
+	if ocspStatus.code != ocspSuccess {
+		return
+	}
```
smtakeda: exit if OCSP Cache server error occurs.

ocsp.go
```diff
@@ -416,7 +438,6 @@ func retryOCSP(
 	}
 	return ocspRes, ocspResBytes, &ocspStatus{
 		code: ocspSuccess,
-		err:  nil,
```
smtakeda: just redundant

ocsp.go
```diff
@@ -634,7 +616,8 @@ func initOCSPCache() {
 		}
 	}
 	for k, cacheValue := range buf {
-		if extractOCSPCacheResponseValue(cacheValue) == ocsp.Unknown {
+		status := extractOCSPCacheResponseValueWithoutSubject(cacheValue)
```
h-chaturvedi: Right now we are validating OCSP Responses multiple times. During init we parse and validate all OCSP Responses completely. This is slightly similar to Python where we validate the timestamp and the time validity of the OCSP Responses. Trying to remember why we decided to do this.

smtakeda: At init time, the OCSP response cache is read from a local disk, which may have invalid OCSP response. When downloading the OCSP response cache, the contents may be invalid, so do we.

h-chaturvedi: But this should be taken care when the driver actually needs that OCSP Response right? Otherwise the driver doesn't really care if the local cache has bad responses. The only case it would care would be in Python driver and that too when the driver is being used by the OCSP Cache Server because we don't want the caches to have bad responses.

ocsp.go
```diff
@@ -517,7 +498,7 @@ func verifyPeerCertificate(verifiedChains [][]*x509.Certificate) (err error) {
 		}
 		results := getAllRevocationStatus(verifiedChains[i])
```
h-chaturvedi: Does this mean that the Root CA Certs will also be verified?

ocsp.go
smtakeda: yeah, this is a kind of some optimization. For cost and code complexity wise, it is not big deal here. Both validations and the actual validations code share the same code path in Go driver.
Python Connector has duplications, so we can simplify it by refactoring codes.

ocsp.go
smtakeda: No, it won't. The root CA is appended as the issuer cert to validate the subject.

ocsp.go
```diff
@@ -598,6 +577,9 @@ func getAllRevocationStatus(verifiedChains []*x509.Certificate) []*ocspStatus {
 	results := make([]*ocspStatus, n)
 	for j := 0; j < n; j++ {
 		results[j] = getRevocationStatus(verifiedChains[j], verifiedChains[j+1])
+		if !isValidOCSPStatus(results[j].code) {
```
h-chaturvedi: This would skip further validation only if the Certificate Status of the OCSP Response is invalid. Ideally the only case where we should continue checking OCSP Responses for other Certificates is if the current OCSP Response has certificate status == Good. For ex, if even a single certificate in the chain is Revoked the whole chain is untrustworthy and should be discarded.

h-chaturvedi: Other than this comment .. LGTM.

h-chaturvedi: Actually, there is one more thing - the fail open log? I don't see the Fail Open specific Logging. Something like this - https://devtools1.int.snowflakecomputing.com/fisheye/browse/svn/Snowflake/trunk/Python/snowflake/connector/ocsp_snowflake.py?hb=true#to963

smtakeda: ah, ok, let me add the log.

ocsp.go
```diff
@@ -407,101 +517,272 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+	ocspResponseCacheLock.Lock()
+	if cacheUpdated {
+		writeOCSPCacheFile()
+	}
+	cacheUpdated = false
+	ocspResponseCacheLock.Unlock()
+	return nil
+}
+
+func canEarlyExitForOCSP(results []*ocspStatus, chainSize int) *ocspStatus {
```
smtakeda: Updated canEarlyExitForOCSP.

ocsp.go
```diff
@@ -407,101 +517,272 @@ func verifyPeerCertificate(callback func(*sync.WaitGroup, []*x509.Certificate) [
...
+	if ocspFailOpen == ocspFailOpenFalse {
+		// Fail closed. any error is returned to stop connection
```
smtakeda: In case of FAIL_CLOSED, if any error is found, return it.

ocsp.go
smtakeda: @h-chaturvedi , the last commit should include all changes.
Btw, can you write up the functional spec including how our drivers uses OCSP?

h-chaturvedi: Sure. Created task - SNOW-81523.

smtakeda:
merging to 1.2.0

Issue #228, opened 2019-05-29
Okta Authentication does not work


Veneciaaa:
### Issue description
Setting Authenticator in Config as per [Godoc instructions](https://godoc.org/github.com/snowflakedb/gosnowflake#hdr-Connection_Parameters): "To authenticate through Okta, specify https://<okta_account_name>.okta.com (URL prefix for Okta)." However, it is using default authentication instead of Okta authentication because it is comparing Authenticator directly to "OKTA" in a switch statement [here](https://github.com/snowflakedb/gosnowflake/blob/master/auth.go#L223). It should be checking if the Authenticator _includes_ "OKTA".

### Example code
Proposed solution to check if Authenticator _includes_ "OKTA": https://gist.github.com/Veneciaaa/0e648e45a951d9cfe110fe4f86b37145

### Error log
"Incorrect username or password was specified" because it is using the default authentication method [here](https://github.com/snowflakedb/gosnowflake/blob/master/auth.go#L236) instead of Okta

### Configuration
*Go version:* go1.12.1

*Server version:* 3.26.1

*Client OS:* Max OS

ChTimTsubasa:
Close this Ticket as it is addressed in #233

PR #233, opened 2019-06-12
Fix okta authentication

ChTimTsubasa:
### Description
This PR is from #228, where a previous change mix the okta URL and the okta type.
To avoid having this issue again, I refactor the code to separate an authenticator and the okta url stored internally.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -17,7 +17,6 @@ const (
 	defaultLoginTimeout   = 60 * time.Second  // Timeout for retry for login EXCLUDING clientTimeout
 	defaultRequestTimeout = 0 * time.Second   // Timeout for retry for request EXCLUDING clientTimeout
 	defaultJWTTimeout     = 60 * time.Second
-	defaultAuthenticator  = "snowflake"
 	defaultDomain         = ".snowflakecomputing.com"
 )
 
@@ -37,9 +36,12 @@ type Config struct {
 	Host     string // hostname (optional)
 	Port     int    // port (optional)
 
-	Authenticator      string // snowflake, okta URL, oauth or externalbrowser
+	Authenticator AuthType // The authenticator type
+
 	Passcode           string
 	PasscodeInPassword bool
+
+	OktaURL *url.URL
 
 	LoginTimeout     time.Duration // Login retry timeout EXCLUDING network roundtrip and read out http response
 	RequestTimeout   time.Duration // request retry timeout EXCLUDING network roundtrip and read out http response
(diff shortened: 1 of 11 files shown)


dsn_test.go
```diff
@@ -236,16 +236,30 @@ func TestParseDSN(t *testing.T) {
...
 			config: &Config{
 				Account: "ac", User: "u", Password: "p",
-				Authenticator: authenticatorJWT, PrivateKey: testPrivKey,
+				AuthType: AuthTypeJwt, PrivateKey: testPrivKey,
 				Protocol: "http", Host: "snowflake.local", Port: 9876,
 			},
 			err: nil,
 		},
+		{
+			dsn: fmt.Sprintf("u:p@ac.snowflake.local:9876?account=ac&protocol=http&authenticator=%v", url.QueryEscape("https://ac.okta.com")),
```
ChTimTsubasa: This is the test case to make sure the authenticator type is passed in correctly though dsn

smtakeda:
Adjust the target branch to prep-1.1.19

dsn.go
```diff
@@ -95,8 +97,12 @@ func DSN(cfg *Config) (dsn string, err error) {
 	if cfg.Region != "" {
 		params.Add("region", cfg.Region)
 	}
-	if cfg.Authenticator != defaultAuthenticator {
-		params.Add("authenticator", strings.ToLower(cfg.Authenticator))
+	if cfg.Authenticator != AuthTypeSnowflake {
+		if cfg.Authenticator == AuthTypeOkta {
+			params.Add("authenticator", strings.ToLower(url.QueryEscape(cfg.OktaURL.String())))
```
smtakeda: Do you need this QueryEscape? I thought Line 147 Encode escapes the parameter?

dsn_test.go
```diff
@@ -387,14 +409,36 @@ func TestDSN(t *testing.T) {
...
+				User:               "u",
+				Password:           "p",
+				Account:            "a",
+				Authenticator:      AuthTypeOkta,
+				OktaURL:   			&url.URL{
+					Scheme: "https",
+					Host: 	"sc.okta.com",
+				},
+			},
+			dsn: "u:p@a.snowflakecomputing.com:443?authenticator=https%253a%252f%252fsc.okta.com",
```
smtakeda: This escaped URL doesn't looks right. It should be:
```
https%3A%2F%2Fsc.okta.com
```
I guess you are double escaping the URL parameter. See the above comment.

ChTimTsubasa: Thanks for pointing it out. I just remove the double escaping

dsn.go
ChTimTsubasa: Addressed

PR #243, opened 2019-07-09
Expose trusted Root CA CertPool so that it could be used by SnowCD.

ChTimTsubasa:
### Description
Tls config has a RootCAs field that can be specified by user[https://golang.org/pkg/crypto/tls/], so that any network traffic using the config would look at the those RootCAs to perform validation.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -32,8 +32,8 @@ import (
 // caRoot includes the CA certificates.
 var caRoot map[string]*x509.Certificate
 
-// certPOol includes the CA certificates.
-var certPool *x509.CertPool
+// CertPool includes the CA certificates.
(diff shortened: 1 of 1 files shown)


h-chaturvedi:
Another question, the Cert Pool is updated by downloading the CACert.pem from cURL, when does this download happen? Who does the download? Does it happen on the customer side or we need to update the driver in case the root certs need to be updated?

ChTimTsubasa:
@h-chaturvedi @smtakeda I expose the CA CertPool to let SnowCD being able to reuse the same CertPool as the gosnowflake driver to detect public key chain signed by untrusted CA.

PR #244, opened 2019-07-16
Modify build script and test cases 

ChTimTsubasa:
### Description
The following changes would enable CI for the old driver test v1.1.0:
1. Remove lint and mega check as these tools are invalid
2. Remove 0 timestamp test as the behavior already changed;
3. Remove session timezone as the parameter schema has changed;
4. Remove check on connection parameters as the server behavior has changed recently;

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: go
 go:
   - 1.8
   - 1.9
+  - 1.10.x
+  - 1.11.x
 before_install:
   - openssl aes-256-cbc -k "$super_secret_password" -in parameters.json.enc -out parameters.json -d
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,12 @@ language: go
 go:
   - 1.8
   - 1.9
-  - tip
 before_install:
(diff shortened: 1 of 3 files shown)


ChTimTsubasa:
When this is in, I would enable daily-run test for this branch on Travis

.travis.yml
```diff
@@ -3,13 +3,14 @@ language: go
...
   - 1.9
-  - tip
+  - 1.10.x
+  - 1.11.x
 before_install:
   - openssl aes-256-cbc -k "$super_secret_password" -in parameters.json.enc -out parameters.json -d
 install:
   - go get github.com/satori/go.uuid
 script:
-  - make fmt lint cov
```
smtakeda: why lint was removed?

smtakeda:
ok, I thought a different way to keep the old test run but this should work too.

.travis.yml
ChTimTsubasa: Because the lint tool cannot be install in the old environment with some dependency missing.

ChTimTsubasa:
@smtakeda Can you share your thought on the alternative to build the test?

PR #250, opened 2019-08-12
Fail open fix

ChTimTsubasa:
### Description
There is a size gap between response and overall certificate chain when the chain contains root CA, leading to our code letting go revoked certificate.

https://snowflakecomputing.atlassian.net/browse/SNOW-90763
### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -512,18 +512,19 @@ func isValidOCSPStatus(status ocspStatusCode) bool {
 // verifyPeerCertificate verifies all of certificate revocation status
 func verifyPeerCertificate(verifiedChains [][]*x509.Certificate) (err error) {
 	for i := 0; i < len(verifiedChains); i++ {
-		numberOfNoneRootCA := len(verifiedChains[i]) - 1
-		if !verifiedChains[i][numberOfNoneRootCA].IsCA || string(verifiedChains[i][numberOfNoneRootCA].RawIssuer) != string(verifiedChains[i][numberOfNoneRootCA].RawSubject) {
+		// Certificate signed by Root CA. This should be one before the last in the Certificate Chain
+		numberOfNoneRootCerts := len(verifiedChains[i]) - 1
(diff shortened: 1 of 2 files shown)


smtakeda:
Please merge this to prep-1.3.0

ocsp.go
```diff
@@ -512,18 +512,18 @@ func isValidOCSPStatus(status ocspStatusCode) bool {
 // verifyPeerCertificate verifies all of certificate revocation status
 func verifyPeerCertificate(verifiedChains [][]*x509.Certificate) (err error) {
 	for i := 0; i < len(verifiedChains); i++ {
-		n := len(verifiedChains[i]) - 1
-		if !verifiedChains[i][n].IsCA || string(verifiedChains[i][n].RawIssuer) != string(verifiedChains[i][n].RawSubject) {
+		numberOfNoneRootCA := len(verifiedChains[i]) - 1
```
h-chaturvedi: Did you mean "numberofNonRootCA" ? If yes I think that would be a misnomer. Not all the certs in the chain are CA certs. There is also Leaf cert which is a cert for the Endpoint. So I'd rename it to "numberOfNonRootCerts"

h-chaturvedi: This part of the code was a little confusing to me at first so I'd suggest to put some comments in here. Something like - "Certificate signed by Root CA. This should be one before the last in the Certificate Chain".

ocsp.go
```diff
@@ -512,18 +512,18 @@ func isValidOCSPStatus(status ocspStatusCode) bool {
...
+		if !verifiedChains[i][numberOfNoneRootCA].IsCA || string(verifiedChains[i][numberOfNoneRootCA].RawIssuer) != string(verifiedChains[i][numberOfNoneRootCA].RawSubject) {
```
h-chaturvedi: Add comment - "Check if the last Non Root Cert is also a CA or is self signed." Also I am not sure if we need to perform the second check as the boolean IsCA is set if the certificate is self signed.

PR #255, opened 2019-08-28
Snow 80090

smtakeda:
### Description
Send clientStartTime and retryCount with each /queries/v1/query-request

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
@@ -25,10 +26,11 @@ func init() {
 	random = rand.New(rand.NewSource(time.Now().UnixNano()))
 }
 
+// request_guid is attached to every request against Snowflake
 const requestGUIDKey string = "request_guid"
 
-// Format of "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
-const uuidLen int = 36
+// retryCounter is attached to query-request from the second time
+const retryCounterParam string = "retryCounter"
 
 // This class takes in an url during construction and replace the
 // value of request_guid every time the replace() is called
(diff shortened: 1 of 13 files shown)


retry.go
```diff
@@ -25,60 +26,95 @@ func init() {
...
+	replace() *url.URL
 }
 
 // Make requestGUIDReplacer given a url string
-func makeRequestGUIDReplacer(url string) requestGUIDReplacerI {
-	startIndex := strings.Index(url, requestGUIDKey)
-	if startIndex == -1 {
-		return &transientReplacer{url}
+func makeRequestGUIDReplacer(urlPtr *url.URL) requestGUIDReplacerI {
+	_, err := url.ParseQuery(urlPtr.RawQuery)
```
ChTimTsubasa: The reason to have a transientReplacer is to not even go through the url changing logic in the later retry if the url doesn't contains the requestGUID at the first place. So we might want to detect if the retryGuid key is in the query and if not just return the transientReplacer.

retry.go
```diff
@@ -25,60 +26,95 @@ func init() {
...
-func (replacer *requestGUIDReplacer) replace() string {
-	return replacer.prefix + uuid.New().String() + replacer.suffix
+func (replacer *requestGUIDReplacer) replace() *url.URL {
+	vs, err := url.ParseQuery(replacer.urlPtr.RawQuery)
+	if err != nil {
+		return replacer.urlPtr
+	}
+	if len(vs.Get(requestGUIDKey)) == 0 {
+		return replacer.urlPtr
+	}
```
ChTimTsubasa: The logic above should be moved to the makeRequestGUIDReplacer()

retry.go
```diff
@@ -25,60 +26,95 @@ func init() {
...
+
+func (r *retryUpdate) replaceOrAdd(retry int) *url.URL {
+	if !strings.HasPrefix(r.TargetURL.Path, queryRequestPath) {
+		return r.TargetURL
+	}
+
+	vs, err := url.ParseQuery(r.TargetURL.RawQuery)
+	if err != nil {
+		return r.TargetURL
+	}
```
ChTimTsubasa: I would probably either put these logic in the constructor and have a another struct named transienUpdate as consistent with GUIDReplacer and get slightly better performance in the case there are lot of retries, or just implement this as a simple function with no OOP at all.

retry.go
```diff
@@ -25,60 +26,104 @@ func init() {
...
-	replace() string
+	replace() *url.URL
 }
 
 // Make requestGUIDReplacer given a url string
-func makeRequestGUIDReplacer(url string) requestGUIDReplacerI {
-	startIndex := strings.Index(url, requestGUIDKey)
-	if startIndex == -1 {
-		return &transientReplacer{url}
+func newRequestGUIDReplace(urlPtr *url.URL) requestGUIDReplacer {
```
smtakeda: fixed the optimization issue.

retry.go
```diff
@@ -25,60 +26,104 @@ func init() {
...
+
+// requestIDKey is attached to all requests to Snowflake
+const requestIDKey string = "requestId"
 
 // This class takes in an url during construction and replace the
 // value of request_guid every time the replace() is called
 // When the url does not contain request_guid, just return the original
 // url
-type requestGUIDReplacerI interface {
+type requestGUIDReplacer interface {
```
smtakeda: changed the naming convention, i.e., "er"

retry.go
```diff
@@ -25,60 +26,104 @@ func init() {
...
+	if !strings.HasPrefix(urlPtr.Path, queryRequestPath) {
+		// nop if not query-request
+		return &transientReplaceOrAdd{urlPtr}
+	}
+	values, err := url.ParseQuery(urlPtr.RawQuery)
+	if err != nil {
+		// nop if the URL is not valid
+		return &transientReplaceOrAdd{urlPtr}
+	}
+	return &retryCounterUpdate{urlPtr, values}
```
smtakeda: fixed here too.

ChTimTsubasa:
LGTM

PR #258, opened 2019-09-10
SNOW-90775: no retry connection even with Expired Certificates

smtakeda:
### Description
No retry connection with Expired Certificates

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -4,6 +4,7 @@ package gosnowflake
 
 import (
 	"bytes"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


retry.go
```diff
@@ -243,6 +244,12 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
 						return nil, err
```
ChTimTsubasa: Given the complexity of this whole error unwrapping from line 235, to improve readability and maintainability, I would recommend wrapping them as a function called:
func unwrapError(err) err
and just change the line 235 to be
```
err = unwrapError(err)
if (err) != nil {
  return nil, err
}

```

PR #267, opened 2019-09-18
Test PR for #266

ChTimTsubasa:
TestPR for #266 

CC:@tsuna

### Checklist
- [x] Code compiles correctly
- [x] Run ``make fmt`` to fix inconsistent formats
- [x] Run ``make lint`` to get lint errors and fix all of them
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

PR #268, opened 2019-09-20
time precision is returned in ColumnTypes()

mknister:
### Description
When there is a precision value for a time or timestamp column, a user is now able to retrieve that value using the ColumnTypes() function. A test created in driver_test.go confirms that it works.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,6 +15,7 @@
   version = "1.1"
 
 [[projects]]
+  branch = "master"
   digest = "1:6b98cc8ee2df9bbb5190f1cea00285ea625109667574d16853ba82eaf48d47fd"
   name = "github.com/golang/glog"
   packages = ["."]
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,7 +15,6 @@
   version = "1.1"
 
 [[projects]]
-  branch = "master"
   digest = "1:6b98cc8ee2df9bbb5190f1cea00285ea625109667574d16853ba82eaf48d47fd"
   name = "github.com/golang/glog"
   packages = ["."]
--- a/driver_test.go
+++ b/driver_test.go
@@ -1967,3 +1967,34 @@ func TestClientSessionKeepAliveParameter(t *testing.T) {
 	}
 	defer rows.Close()
 }
+
+func TestTimePrecision(t *testing.T) {
+	var db *sql.DB
+	var err error
+
+	if db, err = sql.Open("snowflake", dsn); err != nil {
+		t.Fatalf("failed to open db. %v, err: %v", dsn, err)
+	}
+	defer db.Close()
+
+	_, err = db.Exec("create or replace table z3 (t1 time(5))")
+	if err != nil {
+		t.Errorf("error while executing query. err : %v", err)
+	}
+	res, err := db.Query("select * from z3")
+	if err != nil {
+		t.Errorf("error while executing query. err : %v", err)
+	}
+
+	cols, _ := res.ColumnTypes()
+	pres, _, _ := cols[0].DecimalSize()
+	if pres != 5 {
+		t.Fatalf("Wrong value returned. Got %v instead of 5.", pres)
+	}
+}
+
+func init() {
+	if !flag.Parsed() {
+		flag.Parse()
+	}
+}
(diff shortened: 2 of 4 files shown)


smtakeda:
@mknister can you fix the test failures?
```
driver_test.go:1982:2: this value of res is never used (SA4006)
driver_test.go:1982:7: this value of err is never used (SA4006)
gosnowflake.mak:29: recipe for target 'clint' failed
```
Also please create PR to prep-1.3.2

smtakeda:
@mknister can you also rebase and push? There is a conflict in driver_test.go

Gopkg.lock
```diff
@@ -15,7 +15,6 @@
   version = "1.1"
 
 [[projects]]
-  branch = "master"
```
smtakeda: @mknister why did you remove?

query.go
```diff
@@ -24,8 +24,8 @@ type execResponseRowType struct {
 	ByteLength int64  `json:"byteLength"`
 	Length     int64  `json:"length"`
 	Type       string `json:"type"`
-	Scale      int64  `json:"scale"`
 	Precision  int64  `json:"precision"`
+	Scale      int64  `json:"scale"`
```
smtakeda: Good catch!

driver_test.go
```diff
@@ -1967,6 +1967,34 @@ func TestClientSessionKeepAliveParameter(t *testing.T) {
...
 
+func TestTimePrecision(t *testing.T) {
+	var db *sql.DB
+	var err error
+
+	if db, err = sql.Open("snowflake", dsn); err != nil {
+		t.Fatalf("failed to open db. %v, err: %v", dsn, err)
+	}
+	defer db.Close()
+	err = db.Ping()
```
smtakeda: Ping is not required as you run db.Exec soon.

Gopkg.lock
mknister: @smtakeda this is something that Goland did automatically and I assumed it was on purpose. Should I revert this change?

driver_test.go
mknister: Ok, I will remove Ping.

Gopkg.lock
smtakeda: @ChTimTsubasa do you know about this line?

smtakeda:
@mknister looks like still there is conflict. Please resolve it and update the change.

Gopkg.lock
ChTimTsubasa: This file is generated by dep. If dep changed it the we should respect its change.

driver_test.go
```diff
@@ -1967,3 +1967,34 @@ func TestClientSessionKeepAliveParameter(t *testing.T) {
...
+
+	cols, _ := res.ColumnTypes()
+	pres, _, _ := cols[0].DecimalSize()
+	if pres != 5 {
+		t.Fatalf("Wrong value returned. Got %v instead of 5.", pres)
+	}
+}
+
+func init() {
+	if !flag.Parsed() {
```
ChTimTsubasa: This is no longer supported in go 1.13. Please remove the whole init() function completely. See https://github.com/snowflakedb/gosnowflake/pull/267/files to look how the init should be performed.

driver_test.go
```diff
@@ -1967,3 +1967,28 @@ func TestClientSessionKeepAliveParameter(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("error while executing query. err : %v", err)
+	}
+
+	cols, _ := res.ColumnTypes()
+	pres, _, _ := cols[0].DecimalSize()
+	if pres != 5 {
+		t.Fatalf("Wrong value returned. Got %v instead of 5.", pres)
+	}
+}
```
smtakeda: Add a new line

driver_test.go
smtakeda: @ChTimTsubasa This change is no longer valid after rebase.

rows.go
```diff
@@ -106,6 +106,10 @@ func (rows *snowflakeRows) ColumnTypePrecisionScale(index int) (precision, scale
 	switch rows.RowType[index].Type {
 	case "fixed":
 		return rows.RowType[index].Precision, rows.RowType[index].Scale, true
+	case "time":
+		return rows.RowType[index].Scale, 0, true
```
ChTimTsubasa: We are setting the precision as the scale and the scale as 0, which is a little confusing to me. Let's hold the change for now and confirm with the customer what is the value he will be looking at.

PR #276, opened 2019-12-19
Snow 121930

snowstakeda:
### Description
SNOW-121930: ensure the new URL pattern, i.e., org-account.snowflakecomputing.com, works
Resurrected Global URL support. For some reason, the previous change was not included in the main branch.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -21,6 +21,75 @@ func TestParseDSN(t *testing.T) {
 	privKeyPKCS1 := generatePKCS1String(testPrivKey)
 	testcases := []tcParseDSN{
 		{
+			dsn: "user:pass@ac-1-laksdnflaf.global/db/schema",
+			config: &Config{
+				Account: "ac-1", User: "user", Password: "pass", Region: "global",
+				Protocol: "https", Host: "ac-1-laksdnflaf.global.snowflakecomputing.com", Port: 443,
+				Database: "db", Schema: "schema",
+				OCSPFailOpen:              OCSPFailOpenTrue,
+				ValidateDefaultParameters: ConfigBoolTrue,
+			},
+			ocspMode: ocspModeFailOpen,
+			err:      nil,
+		},
+		{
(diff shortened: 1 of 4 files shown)


dsn_test.go
```diff
@@ -492,6 +544,22 @@ func TestDSN(t *testing.T) {
...
+				Password: "p",
+				Account:  "a-aofnadsf.somewhere.azure",
+			},
+			dsn: "u:p@a-aofnadsf.somewhere.azure.snowflakecomputing.com:443?ocspFailOpen=true&region=somewhere.azure&validateDefaultParameters=true",
+		},
+		{
+			cfg: &Config{
+				User:     "u",
+				Password: "p",
+				Account:  "a-aofnadsf.global",
```
ChTimTsubasa: Shouldn't the account be "aofnadsf" and "a" should be the org?

ChTimTsubasa:
#275 will goes in after this one since it rely on the fix for SelfSignedURL test.

dsn_test.go
snowstakeda: @notkriswagner , do you have any thought?

snowstakeda: ah nevermind. This is the input, which can include everything before snowflakecomputing.com.

snowstakeda: And in this case, `a` is the account name and `aofnadsf` will be replication id, because the URL is `global` URL @notkriswagner, can you confirm?

ChTimTsubasa:
It feels to me there are 2 patterns here:
1. <org>-<account>.snowflakecomputing.com
2. <account>-<rep_id>.global.snowflakecomputing.com
@notkriswagner

dsn_test.go
notkriswagner: Yes thats correct. `a` is the account name, `aofnadsf` is the external ID and this is a global URL so there is no connection to org-name (yet)

PR #283, opened 2020-02-07
SNOW-127316 Add doc to use private key in Golang

ChTimTsubasa:
### Description
I do not mention any go library to use to decrypt PKCS8 key since we don't want to give an impression to the user that the library's security level is validated by Snowflake.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing


snowstakeda:
Added @SnowflakeMarkGilkey for consistency with other drivers.

ChTimTsubasa:
@smtakeda Since this is only doc change. Can we directly merge this into master?

PR #285, opened 2020-02-20
Remove raw response body logging during okta and external browser authentication

ChTimTsubasa:
### Description
No session/master token is found

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


snowstakeda:
@ankit-bhatnagar167 , this would be another emergency release.

ChTimTsubasa:
@snowstakeda @ankitbhatnagarsf This is patch is not as emergency as the ODBC/.NET since golang does not expose id/session/master token in the first place.

@ankitbhatnagarsf Wondering if you can let me know the version this PR should goes in so that I can create the prep branch?

PR #293, opened 2020-04-23
Added "flag" code and -log_dir option for logging

sanderiam:
### Description
Added notes about the required import of the "flag" module and the code any go application needs for the logging. Also added documentation for the -log_dir option to put the logging into a file either instead of or along with the STDERR output.


sfc-gh-stakeda:
#293 merged the change. Thanks @sanderiam !

PR #298, opened 2020-05-16
SNOW-160669 add -e and pipefail option to wss.sh

sfc-gh-stakeda:
### Description
Add -e to wss.sh to fail immediate on error

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


scripts/wss.sh
```diff
@@ -2,6 +2,8 @@
 #
 # Run whitesource for Golang
 #
+set -e
+set -o pipefail
```
sfc-gh-stakeda, line 6: This will detect failure and exit right away.

PR #312, opened 2020-06-24
SNOW-144296 Multi Statement Support

sfc-gh-jbahk:
### Description

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -87,7 +87,12 @@ func (sc *snowflakeConn) exec(
 			}
 		}
 	}
+	multiCount := ctx.Value(MultiStatementCount)
+	if multiCount != nil {
+		req.Parameters = map[string]interface{}{MultiStatementCount: multiCount}
+	}
 	glog.V(2).Infof("bindings: %v", req.Bindings)
+	glog.V(2).Infof("parameters: %v", req.Parameters)
 
 	headers := make(map[string]string)
 	headers["Content-Type"] = headerContentTypeApplicationJSON
--- a/connection.go
+++ b/connection.go
@@ -7,11 +7,13 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/json"
+	"fmt"
 	"github.com/google/uuid"
 	"net/url"
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"time"
 )
 
 const (
@@ -50,7 +52,7 @@ func (sc *snowflakeConn) isDml(v int64) bool {
 
 func (sc *snowflakeConn) exec(
 	ctx context.Context,
-	query string, noResult bool, isInternal bool, parameters []driver.NamedValue) (*execResponse, error) {
+	query string, noResult bool, isInternal bool, bindings []driver.NamedValue) (*execResponse, error) {
 	var err error
 	counter := atomic.AddUint64(&sc.SequenceCounter, 1) // query sequence counter
 
@@ -62,18 +64,18 @@ func (sc *snowflakeConn) exec(
 	req.IsInternal = isInternal
 	tsmode := "TIMESTAMP_NTZ"
 	idx := 1
-	if len(parameters) > 0 {
-		req.Bindings = make(map[string]execBindParameter, len(parameters))
-		for i, n := 0, len(parameters); i < n; i++ {
-			t := goTypeToSnowflake(parameters[i].Value, tsmode)
+	if len(bindings) > 0 {
+		req.Bindings = make(map[string]execBindParameter, len(bindings))
+		for i, n := 0, len(bindings); i < n; i++ {
+			t := goTypeToSnowflake(bindings[i].Value, tsmode)
 			glog.V(2).Infof("tmode: %v\n", t)
 			if t == "CHANGE_TYPE" {
-				tsmode, err = dataTypeMode(parameters[i].Value)
+				tsmode, err = dataTypeMode(bindings[i].Value)
 				if err != nil {
 					return nil, err
 				}
 			} else {
-				v1, err := valueToString(parameters[i].Value, tsmode)
+				v1, err := valueToString(bindings[i].Value, tsmode)
 				if err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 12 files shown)


dsn.go
```diff
@@ -335,6 +335,21 @@ func ParseDSN(dsn string) (cfg *Config, err error) {
 	return cfg, nil
 }
 
+// AddConnParameter adds a specified connection parameter to dsn
+func AddConnParameter(dsn string, key string, value interface{}) (string, error) {
```
sfc-gh-stakeda, line 339: Do you want to expose this function to the users? If not, let's change the name start with a lower case.

sfc-gh-jbahk: I think this is the only way to allow multistatement by the users -- we want them to manually specify the number of statements they want to execute, rigth?

sfc-gh-stakeda: This function tweaks the DSN string, which should be created by:
- Manually create it from scratch by the user.
- Call `DSN` to convert from`Config` 

In the first case, it is up to the user about how to create it.
In the second case, you should consider adding the new parameter to `Config` or directly use `Params` to add it.

So I don't see why we need this function.

statement_test.go
```diff
@@ -0,0 +1,184 @@
...
+		t.Fatalf("failed to open db. %v, err: %v", dsn, err)
+	}
+	defer db.Close()
+
+	multiStmtQuery := "select 123;\n" +
+		"select 456;\n" +
+		"select 789;\n" +
+		"select '000';"
+
+	rows, _ := db.Query(multiStmtQuery)
```
sfc-gh-stakeda, line 59: Can you please revisit the Query interface why it cannot take extra custom parameter? I thought you could add it. If not, did you consider using `Context`?

sfc-gh-stakeda: The goal is to set the statement level parameter for multistatement support. Did you explore any other option?

statement_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+import (
+	"database/sql"
+	"testing"
+)
+
+func TestMultiStatementTransaction(t *testing.T) {
+	var db *sql.DB
+	var err error
+	ndsn, err := AddConnParameter(dsn, MultiStatementCount, "0")
```
sfc-gh-stakeda, line 11: If`AddConnParameter` is only for test, that's ok but please change the name.

statement_test.go
sfc-gh-jbahk: Upon research, I will move forward with context. The Query interface cannot take other parameters, making it an invalid interface. We can create a wrapper function such as QueryContextMulti() of some sort, however.

dsn.go
sfc-gh-jbahk: I will delete this function.

sfc-gh-jbahk:
Multistatement tests currently not passing on GitHub (passes locally) due to the server side changes not being deployed yet. Will re-run once available.

statement_test.go
```diff
@@ -0,0 +1,296 @@
...
+		err = rows.Scan(&v)
+		if err != nil {
+			t.Errorf("failed to scan: %#v", err)
+		}
+		if v != expected {
+			t.Fatalf("failed to fetch. value: %v", v)
+		}
+		expected++
+	}
+}
```
sfc-gh-stakeda, line 296: Overall positive test cases look ok.
Can you add negative tests? For example, setting 3 statements and run 2 or 4 queries and get the error?

connection.go
```diff
@@ -344,3 +400,89 @@ func (sc *snowflakeConn) stopHeartBeat() {
...
+		return nil, nil
+	}
+	queryIDs := strings.Split(IDs, ",")
+	resultTypes := strings.Split(types, ",")
+	if len(queryIDs) != len(resultTypes) {
+		return []childResult{}, &SnowflakeError{
+			Number:  ErrNoChildrenReturned,
+			Message: errMsgFailedToRetrieveChild,
+		}
+	}
```
sfc-gh-stakeda, line 432: When could this happen? Is it a user or system error? If this is a user error, the error message should be actionable.

statement_test.go
sfc-gh-stakeda: Add the test cases
- Set the user/account's MultiStatement parameter to 0 and run multiple queries.

connection.go
sfc-gh-stakeda: If this is a system error, dump the debug information to the log and/or include them in the error message so that the users can contact to the support with the information.

connection.go
```diff
@@ -48,9 +52,16 @@ func (sc *snowflakeConn) isDml(v int64) bool {
 	return false
 }
 
+func (sc *snowflakeConn) isMultiStmt(data execResponseData) bool {
+	if data.StatementTypeID == statementTypeIDMulti && data.RowType[0].Name == "multiple statement execution" {
+		return true
+	}
+	return false
```
sfc-gh-stakeda, line 59: nit; Can this be oneliner?
```
return data.StatementTypeID == statementTypeIDMulti && data.RowType[0].Name == "multiple statement execution"
```

connection.go
```diff
@@ -264,6 +293,33 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
...
+				rows.ChunkDownloader = populateChunkDownloader(ctx, sc, childData.Data)
+				nextChunkDownloader = rows.ChunkDownloader
+				firstResultSet = true
+			} else {
+				nextChunkDownloader.NextDownloader = populateChunkDownloader(ctx, sc, childData.Data)
+				nextChunkDownloader = nextChunkDownloader.NextDownloader
+			}
+		}
+	}
+
```
sfc-gh-stakeda, line 322: nit; can this be in a separate function?

connection.go
```diff
@@ -344,3 +400,89 @@ func (sc *snowflakeConn) stopHeartBeat() {
...
+	for i, id := range queryIDs {
+		res[i] = childResult{id, resultTypes[i]}
+	}
+	return res, nil
+}
+
+func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string) (execResponse, error) {
+	headers := make(map[string]string)
+	headers["Content-Type"] = headerContentTypeApplicationJSON
+	headers["accept"] = headerAcceptTypeApplicationSnowflake // TODO v1.1: change to JSON in case of PUT/GET
```
sfc-gh-stakeda, line 443: Is this comment valid?

sfc-gh-jbahk: The headers were a part of the query fetching function that I copied. I'm not sure if it's valid (presumably it was left as a TODO) but I left it incase someone wanted to revisit the header. I can look into it and keep or remove it. Would that be acceptable?

sfc-gh-stakeda: If this function is purely to get result set, you won't need to  change this header for PUT and GET command. This comment is for `query-request` call

sfc-gh-jbahk: Got it. I'll remove it

connection.go
sfc-gh-jbahk: Yes, though the function call to it will be a little ugly. Would you recommend I do the same for `ExecContext()`as well to mirror this?

connection.go
sfc-gh-jbahk: This would be a system error where the server side didn't return the proper number of children to begin with. I'll add it to the log and error message for the users to see.

connection.go
sfc-gh-stakeda: A question is code readability. In general, deeply indented code is often hard to read. Not critical. If adding a function is ugly, I'm ok too.

PR #325, opened 2020-08-17
SNOW-183510 Update context behavior function signatures

sfc-gh-jbahk:
### Description
Added
- Update chunk download helper function signature to return error and handle error
- Parity in function signatures within restful.go / restful object
- Handle request timeout errors for restful object
- Add and expose the query ID for DMLs as a part of Snowflake result

Not added
- New describeContext function (sets parameter describeOnly) at the statement level. Requires exposing this function by making Snowflake Statement object public as well.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -235,7 +235,9 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 		glog.V(2).Infof("number of updated rows: %#v", updatedRows)
 		return &snowflakeResult{
 			affectedRows: updatedRows,
-			insertID:     -1}, nil // last insert id is not supported by Snowflake
+			insertID:     -1,
+			queryID:      sc.QueryID,
+		}, nil // last insert id is not supported by Snowflake
 	} else if sc.isMultiStmt(data.Data) {
 		childResults := getChildResults(data.Data.ResultIDs, data.Data.ResultTypes)
 		for _, child := range childResults {
--- a/connection.go
+++ b/connection.go
@@ -60,7 +60,11 @@ func (sc *snowflakeConn) isMultiStmt(data execResponseData) bool {
 
 func (sc *snowflakeConn) exec(
 	ctx context.Context,
-	query string, noResult bool, isInternal bool, bindings []driver.NamedValue) (*execResponse, error) {
+	query string,
+	noResult bool,
+	isInternal bool,
+	bindings []driver.NamedValue) (
+	*execResponse, error) {
 	var err error
 	counter := atomic.AddUint64(&sc.SequenceCounter, 1) // query sequence counter
 
@@ -257,7 +261,9 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 		glog.V(2).Infof("number of updated rows: %#v", updatedRows)
 		return &snowflakeResult{
 			affectedRows: updatedRows,
-			insertID:     -1}, nil
+			insertID:     -1,
+			queryID:      sc.QueryID,
+		}, nil
 	}
 	glog.V(2).Info("DDL")
 	return driver.ResultNoRows, nil
@@ -297,6 +303,7 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
 			RowSetBase64: data.Data.RowSetBase64,
 		},
 	}
+	rows.queryID = sc.QueryID
 
 	if sc.isMultiStmt(data.Data) {
 		childResults := getChildResults(data.Data.ResultIDs, data.Data.ResultTypes)
--- a/connection.go
+++ b/connection.go
@@ -192,7 +192,7 @@ func (sc *snowflakeConn) Close() (err error) {
 	glog.V(2).Infoln("Close")
 	sc.stopHeartBeat()
 
-	err = sc.rest.FuncCloseSession(sc.rest)
+	err = sc.rest.FuncCloseSession(context.TODO(), sc.rest)
 	if err != nil {
 		glog.V(2).Info(err)
 	}
--- a/connection.go
+++ b/connection.go
@@ -138,7 +138,7 @@ func (sc *snowflakeConn) exec(
 	}
 	glog.V(2).Infof("Success: %v, Code: %v", data.Success, code)
 	if !data.Success {
-		return data, &SnowflakeError{
+		return nil, &SnowflakeError{
 			Number:   code,
 			SQLState: data.Data.SQLState,
 			Message:  data.Message,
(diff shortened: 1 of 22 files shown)


rows.go
```diff
@@ -400,116 +410,98 @@ func downloadChunkHelper(ctx context.Context, scd *snowflakeChunkDownloader, idx
...
+		glog.Flush()
+		return &SnowflakeError{
+			Number:      ErrFailedToGetChunk,
+			SQLState:    SQLStateConnectionFailure,
+			Message:     errMsgFailedToGetChunk,
+			MessageArgs: []interface{}{idx},
 		}
-		st := &largeResultSetReader{
-			status: 0,
-			body:   source,
```
sfc-gh-stakeda, line 431: @sfc-gh-jbahk I recall this is required for Azure Blob support. Can you check the history?
https://github.com/snowflakedb/gosnowflake/commit/2264cec1

sfc-gh-stakeda:
@sfc-gh-jbahk can you elaborate how query id should be exposed?

sfc-gh-jbahk:
@sfc-gh-stakeda sure, we essentially just add a `queryID` field for the user to be able to call on their `SnowflakeResult` object. It requires simply adding an extra field to the rows (for select queries) and the snowflakeresults (for dml) and creating a function that retrieves it.

rows.go
sfc-gh-jbahk: I moved the core functionality into a separate method because the method was getting too long. The `decodeChunk` method takes care of decoding (whether it be JSON, Arrow or the custom JSON decoder). The functionality should be the same.

sfc-gh-stakeda: hmm, it is interesting... The GH action is passing for GCP/GCS. @sfc-gh-jbahk , can you check if this is no longer required? The problem was since GCS didn't set the header properly, the clients could not get the right content type.

sfc-gh-stakeda: @sfc-gh-jbahk , if you don't have access to GCS account, let me know. You could use the local dev gcsaccount.

sfc-gh-stakeda:
@sfc-gh-jbahk , can you find the best way to expose the query id?

rows.go
sfc-gh-stakeda: ah, ok, so it is checking the Gzip header. I took wrong.

sfc-gh-stakeda:
@sfc-gh-jbahk https://github.com/sigmacomputing/gosnowflake/commit/9a539cac11808dddcab96afcb49c9a40f55fb222 is missing.

sfc-gh-jbahk:
@sfc-gh-stakeda That commit is already implemented here: [https://github.com/snowflakedb/gosnowflake/blob/SNOW-183510-update-context-behavior-function-signatures/retry.go#L282-L291](url)

sfc-gh-stakeda:
ok, thanks for the confirmation.
@sfc-gh-jbahk can you please summarize what you merged and not merged here?

sfc-gh-jbahk:
@sfc-gh-stakeda I updated the PR description.

sfc-gh-stakeda:
The describe functionality can be mapped to https://golang.org/pkg/database/sql/#Conn.PrepareContext, which returns `Stmt`. Let's create another ticket for PrepareContext implementation.

sfc-gh-stakeda:
How can the application specify the timeout? Can you add an example as a test case?

sfc-gh-jbahk:
@sfc-gh-stakeda The tests are the ones modified in restful_test.go (`time.Second` is the controlled variable) regarding specifying the timeout.

statement_test.go
```diff
@@ -439,3 +440,36 @@ func TestMultiStatementCountMismatch(t *testing.T) {
...
+
+	ctx := context.TODO()
+	conn, _ := db.Conn(ctx)
+
+	err = conn.Raw(func(x interface{}) error {
+		stmt, err := x.(driver.ConnPrepareContext).PrepareContext(ctx, "select 1")
+		if err != nil {
+			return err
+		}
+		rows, err := stmt.(driver.StmtQueryContext).QueryContext(ctx, nil)
```
sfc-gh-stakeda, line 461: This test is good. What will happen if `QueryContext` fails? For example, a SQL syntax error should return an error code along with a query id.

sfc-gh-jbahk: If `QueryContext` fails, there will be no associated query ID since the entire query would fail. An example of that would be a SQL syntax, but it doesn't execute in the backend so there is no query ID to speak of.

sfc-gh-stakeda: There must be a query_id for failed queries. Otherwise the customers cannot locate which query failed. Can you find out the query id in the response from query-request? It is supported by Python Connection, btw.

sfc-gh-jbahk: Oh, I see. It hasn't been done in this driver it looks like - we just return the error immediately. I'll look to see how we can extract that information, but it most likely will bring about a behavior change (we return `nil` at the moment as a result along with an error, but we'll have to actually return an object).

sfc-gh-jbahk: So yes, I can return a row/result, but it replaces the existing return value of `nil`. In the case that the customers check the nullity of the returned object instead of the error returned, this will cause an issue.

sfc-gh-jbahk: @sfc-gh-stakeda I just pushed a prototype of what it'd look like if we were to return a failed SQL query object with the corresponding query ID.

sfc-gh-stakeda: can you include the query id in SnowflakeError?

statement_test.go
```diff
@@ -460,14 +460,32 @@ func TestGetQueryID(t *testing.T) {
...
+			return err
+		}
+		rows, err = stmt.(driver.StmtQueryContext).QueryContext(ctx, nil)
+		if err == nil {
+			t.Fatal("should have failed to execute query")
+		}
+		if driverErr, ok := err.(*SnowflakeError); ok {
+			if driverErr.QueryID == "" {
+				t.Fatal("should have an associated query ID")
+			}
```
sfc-gh-stakeda, line 482: can you check the error code as well?

sfc-gh-stakeda: If a syntax error occurs, the error code should be returned from the GS.

connection.go
```diff
@@ -230,7 +230,7 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 	if err != nil {
 		glog.V(2).Infof("error: %v", err)
 		if data != nil {
-			return &snowflakeResult{queryID: data.Data.QueryID}, err
+			return nil, &SnowflakeError{Message: err.Error(), QueryID: data.Data.QueryID}
```
sfc-gh-stakeda, line 233: SnowflakeError must include `Number` and `SQLState` as well for all cases.

PR #331, opened 2020-09-09
SNOW-160505 Async Support

sfc-gh-jbahk:
### Description
This PR allows asynchronous querying by enabling the AsyncExec parameter to be set via the driver, which, when sent to the server, will not block waiting for the entire query to finish.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -77,7 +77,7 @@ func (sc *snowflakeConn) exec(
 			req.Parameters = map[string]interface{}{string(MultiStatementCount): key}
 		}
 	}
-	glog.V(2).Infof("parameters: %v", req.Parameters)
+	logger.WithContext(ctx).Infof("parameters: %v", req.Parameters)
 
 	tsmode := "TIMESTAMP_NTZ"
 	idx := 1
@@ -109,16 +109,7 @@ func (sc *snowflakeConn) exec(
 			}
 		}
 	}
-<<<<<<< HEAD
-	multiCount := ctx.Value(MultiStatementCount)
-	if multiCount != nil {
-		req.Parameters = map[string]interface{}{string(MultiStatementCount): multiCount}
-	}
 	logger.WithContext(ctx).Infof("bindings: %v", req.Bindings)
-	logger.WithContext(ctx).Infof("parameters: %v", req.Parameters)
-=======
-	glog.V(2).Infof("bindings: %v", req.Bindings)
->>>>>>> 13a25c6... add async support
 
 	headers := make(map[string]string)
 	headers["Content-Type"] = headerContentTypeApplicationJSON
--- a/connection.go
+++ b/connection.go
@@ -26,12 +26,6 @@ const (
 	statementTypeIDDelete           = statementTypeIDDml + int64(0x300)
 	statementTypeIDMerge            = statementTypeIDDml + int64(0x400)
 	statementTypeIDMultiTableInsert = statementTypeIDDml + int64(0x500)
-
-	statementTypeIDDdl     = int64(0x6000)
-	statementTypeIDCreate  = statementTypeIDDdl + int64(0x101)
-	statementTypeIDComment = statementTypeIDDdl + int64(0x200)
-	statementTypeIDDrop    = statementTypeIDDdl + int64(0x300)
-	statementTypeIDAlter   = statementTypeIDDdl + int64(0x401)
 )
 
 const (
@@ -245,15 +239,11 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 	if sc.rest == nil {
 		return nil, driver.ErrBadConn
 	}
-	internal, err := isInternal(ctx)
-	if err != nil {
-		return nil, err
-	}
 	noResult, err := isAsyncMode(ctx)
 	if err != nil {
 		return nil, err
 	}
-	data, err := sc.exec(ctx, query, noResult, internal, args)
+	data, err := sc.exec(ctx, query, noResult, false, args)
 	if err != nil {
 		logger.WithContext(ctx).Infof("error: %v", err)
 		if data != nil {
@@ -340,15 +330,11 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
 		return nil, driver.ErrBadConn
 	}
 
-	internal, err := isInternal(ctx)
-	if err != nil {
-		return nil, err
-	}
 	noResult, err := isAsyncMode(ctx)
 	if err != nil {
 		return nil, err
 	}
-	data, err := sc.exec(ctx, query, noResult, internal, args)
+	data, err := sc.exec(ctx, query, noResult, false, args)
 	if err != nil {
 		logger.WithContext(ctx).Errorf("error: %v", err)
 		if data != nil {
@@ -447,16 +433,11 @@ func (sc *snowflakeConn) Ping(ctx context.Context) error {
 	if sc.rest == nil {
 		return driver.ErrBadConn
 	}
-
-	internal, err := isInternal(ctx)
-	if err != nil {
-		return err
-	}
 	noResult, err := isAsyncMode(ctx)
 	if err != nil {
 		return err
 	}
-	_, err = sc.exec(ctx, "SELECT 1", noResult, internal, []driver.NamedValue{})
+	_, err = sc.exec(ctx, "SELECT 1", noResult, false, []driver.NamedValue{})
 	return err
 }
 
@@ -581,18 +562,6 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string) 
 		return nil, err
 	}
 	return respd, nil
-}
-
-func isInternal(ctx context.Context) (bool, error) {
-	val := ctx.Value(IsInternal)
-	if val == nil {
-		return false, nil
-	}
-	boolVal, ok := val.(bool)
-	if !ok {
-		return false, fmt.Errorf("failed to cast val %+v to bool", val)
-	}
-	return boolVal, nil
 }
 
 func isAsyncMode(ctx context.Context) (bool, error) {
--- a/connection.go
+++ b/connection.go
@@ -26,6 +26,12 @@ const (
 	statementTypeIDDelete           = statementTypeIDDml + int64(0x300)
 	statementTypeIDMerge            = statementTypeIDDml + int64(0x400)
 	statementTypeIDMultiTableInsert = statementTypeIDDml + int64(0x500)
+
+	statementTypeIDDdl     = int64(0x6000)
+	statementTypeIDCreate  = statementTypeIDDdl + int64(0x101)
+	statementTypeIDComment = statementTypeIDDdl + int64(0x200)
+	statementTypeIDDrop    = statementTypeIDDdl + int64(0x300)
+	statementTypeIDAlter   = statementTypeIDDdl + int64(0x401)
 )
 
 const (
@@ -45,13 +51,7 @@ type snowflakeConn struct {
 
 // isDml returns true if the statement type code is in the range of DML.
 func (sc *snowflakeConn) isDml(v int64) bool {
-	switch v {
-	case statementTypeIDDml, statementTypeIDInsert,
-		statementTypeIDUpdate, statementTypeIDDelete,
-		statementTypeIDMerge, statementTypeIDMultiTableInsert:
-		return true
-	}
-	return false
+	return statementTypeIDDml <= v && v <= statementTypeIDMultiTableInsert
 }
 
 // isMultiStmt returns true if the statement type code is of type multistatement
@@ -73,9 +73,18 @@ func (sc *snowflakeConn) exec(
 	req := execRequest{
 		SQLText:    query,
 		AsyncExec:  noResult,
+		IsInternal: isInternal,
 		SequenceID: counter,
 	}
-	req.IsInternal = isInternal
+	params := []paramKey{MultiStatementCount}
+	for _, k := range params {
+		key := ctx.Value(k)
+		if key != nil {
+			req.Parameters = map[string]interface{}{string(MultiStatementCount): key}
+		}
+	}
+	glog.V(2).Infof("parameters: %v", req.Parameters)
+
 	tsmode := "TIMESTAMP_NTZ"
 	idx := 1
 	if len(bindings) > 0 {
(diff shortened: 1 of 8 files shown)


connection.go
```diff
@@ -26,6 +26,12 @@ const (
 	statementTypeIDDelete           = statementTypeIDDml + int64(0x300)
 	statementTypeIDMerge            = statementTypeIDDml + int64(0x400)
 	statementTypeIDMultiTableInsert = statementTypeIDDml + int64(0x500)
+
+	statementTypeIDDdl     = int64(0x6000)
+	statementTypeIDCreate  = statementTypeIDDdl + int64(0x101)
+	statementTypeIDComment = statementTypeIDDdl + int64(0x200)
+	statementTypeIDDrop    = statementTypeIDDdl + int64(0x300)
+	statementTypeIDAlter   = statementTypeIDDdl + int64(0x401)
```
sfc-gh-stakeda, line 34: why you need these?

sfc-gh-stakeda:
@sfc-gh-jbahk , please add more comments to the PR or the code itself to explain what change you are making to help the reviewers get context quickly.

connection.go
sfc-gh-jbahk: It was requested in an open issue.

sfc-gh-stakeda: It would be the best if you add this when it is required.

restful.go
```diff
@@ -199,7 +203,31 @@ func postRestfulQueryHelper(
 
 		var resultURL string
 		isSessionRenewed := false
+		noResult, _ := isAsyncMode(ctx)
```
sfc-gh-cshi, line 206: Should Restful as a wrapper on RESTFUL req/resp (transport layer, provide some hooks - 'Func'  at the most) be aware of that it has to write things to Rows or Result or even Rows/Result data structures that are specific to golang's sql module? Fill and return a wrapper object (e.g. QueryHandler) to include queryID, status, and statusChannel should be enough.  I think the original code that aimed to return queryId  only in async manner is simpler and more flexible.

sfc-gh-jbahk: As discussed in the call today, the issue with returning a query handler object is that the user has no way to retrieve the `rows` or `result` object after. We can build a way for that to be possible, then our initial method of hanging on calls like `rows.Next()` is deemed moot, since that call depends on the async retrieval being completed. If we return a query handler and use the query ID to retrieve the result, the retrieval process will be kicked off only when the customer queries again. This method is more inline with existing code and APIs

sfc-gh-cshi: I thought I already answered the question on Rows.Next() and we agreed to hanging on Rows.Next() is the simple way to code.  What was the concern raised on blocking at Rows.Next()? I think we agree that to make ExcuteContext and QueryContext async is the goal, and this PR is to allow first query to return immediately, but if future callers want to get queryId for this query they have to wait for the queryId channel to return something.

rows.go
```diff
@@ -197,6 +213,14 @@ func (rows *snowflakeRows) NextResultSet() error {
 	return rows.ChunkDownloader.nextResultSet()
 }
 
+func (rows *snowflakeRows) checkAsyncQueryStatus() {
+	// if async query, block until query is finished
+	if rows.status == QueryStatusInProgress {
+		rows.status = <-rows.statusChannel
+		close(rows.statusChannel)
```
sfc-gh-anezvigin, line 220: In general it's a good idea to close a channel on the writer side, not the reader side. Reading from a closed channel gives you a zero-value but writing to a closed channel panics. It's easier to spot a write on a closed channel if you do the closing around the place where you do the writing.

rows.go
```diff
@@ -197,6 +213,14 @@ func (rows *snowflakeRows) NextResultSet() error {
 	return rows.ChunkDownloader.nextResultSet()
 }
 
+func (rows *snowflakeRows) checkAsyncQueryStatus() {
```
sfc-gh-anezvigin, line 216: I would call this something like `waitForQueryStatusIfNeeded()` to make it clear at the call site that this function might block.

result.go
```diff
@@ -1,26 +1,58 @@
...
-func (res *snowflakeResult) QueryID() string {
+func (res *snowflakeResult) GetQueryID() string {
 	return res.queryID
 }
+
+func (res *snowflakeResult) GetStatus() queryStatus {
+	return res.status
+}
+
+func (res *snowflakeResult) checkAsyncExecStatus() {
```
sfc-gh-anezvigin, line 52: Same comments as in rows.go

restful.go
```diff
@@ -199,7 +203,31 @@ func postRestfulQueryHelper(
...
+				res.queryID = respd.Data.QueryID
+				res.status = QueryStatusInProgress
+				res.statusChannel = make(chan queryStatus)
+				respd.Data.AsyncResult = res
+			case "query":
+				rows.queryID = respd.Data.QueryID
+				rows.status = QueryStatusInProgress
+				rows.statusChannel = make(chan queryStatus)
+				respd.Data.AsyncRows = rows
+			default:
```
sfc-gh-anezvigin, line 224: Should this default case fail query execution? The queryID, status, and statusChannel not being set feels like undefined behavior once the goroutine is kicked off below.

restful.go
```diff
@@ -433,3 +461,69 @@ func cancelQuery(ctx context.Context, sr *snowflakeRestful, requestID string, ti
...
+		return
+	}
+
+	respd := execResponse{} // reset the response
+	err = json.NewDecoder(resp.Body).Decode(&respd)
+	resp.Body.Close()
+	if err != nil {
+		logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
+		return
+	}
```
sfc-gh-anezvigin, line 487: What's actually going to happen when this error is handled? Will rows.Next() block indefinitely?

You may wanna update statusChannel to not just receive a `queryStatus string` but a struct like:

```
type queryEvent struct {
  status queryStatus
  err error
}
```

Then update all of the checkAsyncExecStatus() functions to this signature:

```
checkAsyncExecStatus() error
```

Then update places like Next() to do something like this:

```
func (rows *snowflakeRows) Next(dest []driver.Value) (err error) {
    if err := rows.checkAsyncQueryStatus(); err != nil {
        return err
    }
```

That's just one idea though. The error handling flow would need to be tested (manually and I suggest writing unit tests for it as well).

statement_test.go
```diff
@@ -497,3 +498,43 @@ func TestGetQueryID(t *testing.T) {
 		t.Fatalf("failed to prepare statement. err: %v", err)
 	}
 }
+
+func TestAsyncMode(t *testing.T) {
```
sfc-gh-anezvigin, line 502: What does code coverage look like? Feels like there aren't enough test cases.

sfc-gh-anezvigin (changes requested):
Overall approach looks good. Left some notes.

restful.go
sfc-gh-jbahk: This makes sense. I'll update it as such.

restful.go
sfc-gh-jbahk: Good point. I'll short circuit this and return the returned &respd object before the goroutine is kicked off.

statement_test.go
sfc-gh-jbahk: It's at 80.7% with the push I just made.

statement_test.go
```diff
@@ -497,3 +498,91 @@ func TestGetQueryID(t *testing.T) {
...
```
sfc-gh-cshi, line 502: I would suggest to add test cases in which several queries within the same session or in different sessions will be filed asynchronously, and some results of late queries could return earlier.

connection.go
```diff
@@ -73,9 +69,18 @@ func (sc *snowflakeConn) exec(
...
 		SequenceID: counter,
 	}
-	req.IsInternal = isInternal
+	params := []paramKey{MultiStatementCount}
+	for _, k := range params {
+		key := ctx.Value(k)
+		if key != nil {
+			req.Parameters = map[string]interface{}{string(MultiStatementCount): key}
+		}
+	}
```
sfc-gh-anezvigin, line 81: I can see what the for loop is trying to do (save on some typing later via the params slice), but this doesn't seem necessary. The following is much clearer albeit more boilerplate-y:

```
req := execRequest{
  // ...
  Parameters: map[string]interface{}{},
  // ...
}

if key := ctx.Value(MultiStatementCount); key != nil {
  req.Parameters[string(MultiStatementCount)] = key
}
if key := ctx.Value(SomeOtherThing); key != nil {
  req.Parameters[string(SomeOtherThing)] = key
}
if key := ctx.Value(YetAnotherThing); key != nil {
  req.Parameters[string(YetAnotherThing)] = key
}
```

connection.go
```diff
@@ -547,6 +570,155 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
...
+			rows.statusChannel <- queryEvent{
+				QueryFailed,
+				&SnowflakeError{
+					Number:  -1,
+					Message: err.Error(),
+					QueryID: rows.queryID,
+				},
+			}
+			close(rows.statusChannel)
+		}
```
sfc-gh-anezvigin, line 619: Can be DRY'd up significantly. The only thing different between both branches is the query ID and the channel. It's a non-trivial point because someone might update the error message for one SnowflakeError{} definition but forget the other one.

sfc-gh-anezvigin: Actually what you could do is just define the channel at the very top of the function like:

```
var statusChannel chan queryEvent
if resType == "exec" {
  statusChannel = ...
} else {
  statusChannel = ...
}
```

This way you can do away with a lot of the if statements with duplicate logic in them.

connection.go
```diff
@@ -547,6 +570,155 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
...
+		return
+	}
+
+	respd := execResponse{}
+	err = json.NewDecoder(resp.Body).Decode(&respd)
+	resp.Body.Close()
```
sfc-gh-anezvigin, line 625: There's a small risk of a memory leak here, but I think it can be easily mitigated this way after calling FuncGet() above:

```
if resp.Body != nil {
  defer resp.Body.Close()
}
```

I _think_ (please verify) that it's okay to call resp.Body.Close() multiple times so you can still leave this close call in here.

The memory leak risk is there's a lot of code between this close call and the response body being created. If anything panics along the way (recovered later by the user) or if FuncGet() doesn't guarantee that `err != nil && resp = nil` -- you might have a leak that's hard to track down.

connection.go
```diff
@@ -547,6 +570,155 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
...
+	if respd.Code != "" {
+		code, err = strconv.Atoi(respd.Code)
+		if err != nil {
+			code = -1
+		}
+	} else {
+		code = -1
+	}
+
+	if resType, _ := ctx.Value(resultType).(string); resType == "exec" {
```
sfc-gh-anezvigin, line 662: Do we need to do this casting every time? Can it be done at the top of the func?

connection.go
```diff
@@ -547,6 +570,155 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
...
+				&SnowflakeError{
+					Number:   code,
+					SQLState: respd.Data.SQLState,
+					Message:  respd.Message,
+					QueryID:  respd.Data.QueryID,
+				},
+			}
+		}
+		close(rows.statusChannel)
+	}
```
sfc-gh-anezvigin, line 719: What is the else case here? You might as well panic so that channels don't block indefinitely. BTW if `restType` mapped to a constant of some kind it would decrease the likelihood of setting the wrong thing by accident. There are parts of the code that are doing this:

```
ctx = context.WithValue(ctx, resultType, "exec")
```

When it could be doing this:

```
ctx = setQueryResultType(ctx, ExecResultType)
```

With a complementary:

```
getQueryResultType(ctx) // ExecResultType | RowsResultType
```

result.go
```diff
@@ -1,26 +1,76 @@
...
+		event := <-res.statusChannel
+		res.status = event.status
+		if event.err != nil {
+			res.err = event.err
+			return res.err
+		}
+		res.status = event.status
+	} else if res.status == QueryFailed {
+		return res.err
+	}
```
sfc-gh-anezvigin, line 74: Looking at this function made me realize... why not just make `res.statusChannel` more like `res.errorChannel` and have it be `chan error`? The only two things we send on this channel are complete events and errors. So if we receive `nil` on the channel that just means complete. Doing it this way minimizes the amount of states we need to worry about.

statement_test.go
sfc-gh-anezvigin: Got it, thanks. The main thing I'm worried about is making sure the edge cases are covered in tests and coverage helps highlight those.

sfc-gh-anezvigin (commented):
Left a few more comments.

result.go
sfc-gh-jbahk: This makes perfect sense. Will do.

connection.go
sfc-gh-jbahk: I confirmed that it's okay to call `.Close()` on a ReadCloser.

connection.go
```diff
@@ -232,8 +235,13 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 	if sc.rest == nil {
 		return nil, driver.ErrBadConn
 	}
-	// TODO: handle noResult and isInternal
-	data, err := sc.exec(ctx, query, false, false, args)
+	noResult, err := isAsyncMode(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// TODO handle isInternal
```
sfc-gh-anezvigin, line 242: Aren't we handling isInternal already? Above?

sfc-gh-anezvigin (approved):

connection.go
sfc-gh-jbahk: The feature isn't in place yet -- we just pass in the isInternal value but it's false by default currently.

connection.go
```diff
@@ -547,6 +581,116 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
...
+		sfError.QueryID = rows.queryID
+	}
+	defer close(errChannel)
+	headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, sr.Token)
+	resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
+	if err != nil {
+		logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
+		sfError.Message = err.Error()
+		errChannel <- sfError
+		close(errChannel)
```
Yifeng-Sigma, line 624: channel might close twice here

PR #332, opened 2020-09-11
SNOW-184545 Prepared Statements

sfc-gh-jbahk:
### Description
Enable prepared statements (and prepared statements with context) for Golang.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


statement_test.go
```diff
@@ -488,12 +484,38 @@ func TestGetQueryID(t *testing.T) {
...
-			t.Fatal("rows should be empty")
-		}
 		return nil
 	})
 	if err != nil {
 		t.Fatalf("failed to prepare statement. err: %v", err)
 	}
 }
+
+func TestPreparedStatement(t *testing.T) {
```
sfc-gh-stakeda, line 494: can you add comments so that the reviews can understand what is the purpose of tests and how you are validating?

transaction.go
```diff
@@ -27,7 +27,7 @@ func (tx *snowflakeTx) Rollback() (err error) {
 	if tx.sc == nil || tx.sc.rest == nil {
 		return driver.ErrBadConn
 	}
-	_, err = tx.sc.exec(context.TODO(), "ROLLBACK", false, false, nil)
+	_, err = tx.sc.exec(context.TODO(), "ROLLBACK", false, false, false, nil)
```
sfc-gh-stakeda, line 30: If think you either add comments to the boolean parameters or change the method to clarify what parameters are. In Java, we usually create a builder to set each of the parameter individually with names and make a function call in the end.

sfc-gh-stakeda: My point is there is two many `false` in the parameters. It is not clear which one is for what by seeing this line.

statement_test.go
```diff
@@ -7,10 +7,40 @@ import (
...
+	s, _ := db.Prepare("insert into test values (?, ?)")
+	defer s.Close()
+	s.Exec(intArray, strArray)
+
+	rows, _ := db.Query("select * from test order by 1")
+	defer rows.Close()
+	var v1 int
+	var v2 string
+	for rows.Next() {
+		err := rows.Scan(&v1, &v2)
```
sfc-gh-cshi, line 37: Should the test verify the returned row number and values?

sfc-gh-cshi (approved):

PR #341, opened 2020-10-19
SNOW-206590 allow specifying the snowflake request id via context

sfc-gh-dwu:
### Description
Allow specific request ID to be set via the context key `SNOWFLAKE_REQUEST_ID`.
If the key is not specified or value is empty, fallback to existing behavior of generating UUID.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/retry.go
+++ b/retry.go
@@ -51,9 +51,9 @@ type requestGUIDReplacer interface {
 
 // Get the request ID from the context if specified, otherwise generate one
 func getRequestID(ctx context.Context) string {
-	requestId, ok := ctx.Value(SnowflakeRequestIDKey).(string)
-	if ok && requestId != "" {
-		return requestId
+	requestID, ok := ctx.Value(SnowflakeRequestIDKey).(string)
+	if ok && requestID != "" {
+		return requestID
 	}
 	return uuid.New().String()
 }
--- a/retry.go
+++ b/retry.go
@@ -26,6 +26,11 @@ func init() {
 	random = rand.New(rand.NewSource(time.Now().UnixNano()))
 }
 
+type contextKey string
+
+// SnowflakeRequestIDKey is optional context key to specify request id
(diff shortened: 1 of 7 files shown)


retry.go
```diff
@@ -44,6 +49,15 @@ type requestGUIDReplacer interface {
...
 }
 
+// Get the request ID from the context if specified, otherwise generate one
+func getRequestID(ctx context.Context) string {
+	requestID, ok := ctx.Value(SnowflakeRequestIDKey).(string)
+	if ok && requestID != "" {
+		return requestID
+	}
+	return uuid.New().String()
+}
```
sfc-gh-anezvigin, line 59: Instead of making `SnowflakeRequestIDKey` public, suggest you make it private and add a public function for setting the request ID. Example:

```
func WithRequestID(ctx context.Context, requestID string) context.Context {
  return context.WithValue(ctx, snowflakeRequestIDKey, requestID)
}
```

Usage outside of gosnowflake:

```
ctx = gosnowflake.WithRequestID(context.Background(), "my-request-id")
```

retry.go
```diff
@@ -44,6 +49,15 @@ type requestGUIDReplacer interface {
 	replace() *url.URL
 }
 
+// Get the request ID from the context if specified, otherwise generate one
+func getRequestID(ctx context.Context) string {
```
sfc-gh-anezvigin, line 53: This going to generate a new request ID every time it's called which isn't idiomatic for context getters. You should be able to call a context getter from anywhere and get the most recent value.

So for example if you did this:

```
func someRestfulFunc(ctx context.Context) {
  requestID := getRequestID(ctx) // returns "foo"
  someOtherRestfulFunc(ctx)
}

func someOtherRestfulFunc(ctx context.Context) {
  requestID := getRequestID(ctx) // uh oh, returns a new request ID
}
```

PR #342, opened 2020-10-19
SNOW-206590 allow specifying the snowflake request id via context

sfc-gh-dwu:
### Description
Allow specific request ID to be set via the context key `SNOWFLAKE_REQUEST_ID`.
If the key is not specified or value is empty, fallback to existing behavior of generating UUID.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


retry.go
```diff
@@ -44,6 +49,15 @@ type requestGUIDReplacer interface {
...
 }
 
+// Get the request ID from the context if specified, otherwise generate one
+func getRequestID(ctx context.Context) string {
+	requestID, ok := ctx.Value(SnowflakeRequestIDKey).(string)
+	if ok && requestID != "" {
+		return requestID
+	}
+	return uuid.New().String()
+}
```
sfc-gh-anezvigin, line 59: Instead of making `SnowflakeRequestIDKey` public, suggest you make it private and add a public function for setting the request ID. Example:

```
func WithRequestID(ctx context.Context, requestID string) context.Context {
  return context.WithValue(ctx, snowflakeRequestIDKey, requestID)
}
```

Usage outside of gosnowflake:

```
ctx = gosnowflake.WithRequestID(context.Background(), "my-request-id")
```

retry.go
```diff
@@ -44,6 +49,15 @@ type requestGUIDReplacer interface {
 	replace() *url.URL
 }
 
+// Get the request ID from the context if specified, otherwise generate one
+func getRequestID(ctx context.Context) string {
```
sfc-gh-anezvigin, line 53: This going to generate a new request ID every time it's called which isn't idiomatic for context getters. You should be able to call a context getter from anywhere and get the most recent value.

So for example if you did this:

```
func someRestfulFunc(ctx context.Context) {
  requestID := getRequestID(ctx) // returns "foo"
  someOtherRestfulFunc(ctx)
}

func someOtherRestfulFunc(ctx context.Context) {
  requestID := getRequestID(ctx) // uh oh, returns a new request ID
}
```

restful.go
```diff
@@ -56,14 +56,14 @@ type snowflakeRestful struct {
...
+	FuncPostQuery       func(context.Context, *snowflakeRestful, *url.Values, map[string]string, []byte, time.Duration, string) (*execResponse, error)
+	FuncPostQueryHelper func(context.Context, *snowflakeRestful, *url.Values, map[string]string, []byte, time.Duration, string) (*execResponse, error)
 	FuncPost            func(context.Context, *snowflakeRestful, *url.URL, map[string]string, []byte, time.Duration, bool) (*http.Response, error)
 	FuncGet             func(context.Context, *snowflakeRestful, *url.URL, map[string]string, time.Duration) (*http.Response, error)
 	FuncRenewSession    func(context.Context, *snowflakeRestful, time.Duration) error
 	FuncPostAuth        func(context.Context, *snowflakeRestful, *url.Values, map[string]string, []byte, time.Duration) (*authResponse, error)
 	FuncCloseSession    func(context.Context, *snowflakeRestful, time.Duration) error
-	FuncCancelQuery     func(context.Context, *snowflakeRestful, *uuid.UUID, time.Duration) error
+	FuncCancelQuery     func(context.Context, *snowflakeRestful, string, time.Duration) error
```
sfc-gh-anezvigin, line 66: Now that the request ID lives in the context, you can probably just get rid of the request ID argument and rely only on the context object for getting the request ID. Otherwise it's unclear which to use -- getRequestID() or the argument?

util.go
```diff
@@ -3,10 +3,31 @@
...
 )
 
+type contextKey string
+
+// snowflakeRequestIDKey is optional context key to specify request id
+const snowflakeRequestIDKey contextKey = "SNOWFLAKE_REQUEST_ID"
+
+// WithRequestID returns a new context with the specified snowflake request id
+func WithRequestID(ctx context.Context, requestID string) context.Context {
+	return context.WithValue(ctx, snowflakeRequestIDKey, requestID)
```
sfc-gh-jbahk, line 19: Might I ask why not pass in the requestID as a type UUID (correct me if I'm wrong, but I think you can convert a string to UUID type) without having to change the function signature for every function that uses a requestID?

restful.go
sfc-gh-dwu: Internally, the functions that have `requestID` explicitly as an argument all use that `requestID` instead of grabbing the request ID from the context.

The three function that take `requestID` explicitly are `FuncPostQuery/FuncPostQueryHelper/FuncCancel`
The FuncPost* need the request ID because internally they have some fancy re-try logic that keeps the request ID stable across re-try, but generates random request ID for renewal and cancel.

To keep the existing behavior intact when `SNOWFLAKE_REQUEST_ID` is not set in the context, it's still simpler to keep the current argument intact and not stamp a request ID into the context if not specified.

util.go
sfc-gh-dwu: According to the REST api documentation, the `requestID` does not have a hard requirement of being a UUID (it's just a string field)
https://docs.snowflake.com/en/user-guide/data-load-snowpipe-rest-apis.html

The upstream callers may also end up stamping in non-UUID request IDs, for example, our Snowflake Golang App server internally doesn't use UUID format. So allowing `string` matches the API doc better and is more flexible for the clients.

retry.go
sfc-gh-dwu: renamed method to `getOrGenerateRequestIDFromContext`

util.go
sfc-gh-jbahk: I see. That makes sense.

sfc-gh-dwu:
@sfc-gh-jbahk Do you think we need to add documentation calling out to this new `WithRequestID` mechanism in the GoDocs?

sfc-gh-jchu:
> @sfc-gh-jbahk Do you think we need to add documentation calling out to this new `WithRequestID` mechanism in the GoDocs?

@sfc-gh-jbahk do you have a suggestion?

sfc-gh-jbahk:
@sfc-gh-dwu @sfc-gh-jchu Hi, sorry, my notifications for GitHub is messed up so I'm just seeing this. Yes, I would recommend adding new documentation. Please coordinate with @sfc-gh-mgilkey.

sfc-gh-dwu:
Updated doc and pinged @sfc-gh-mgilkey for review, thanks.

doc.go
```diff
@@ -195,6 +195,14 @@ Note: The logger may be changed in the future for better logging. Currently if
 the applications use the same parameters as glog, you cannot collect both
 application and driver logs at the same time.
 
+Query request ID
+
+Specific query request ID can be set in the context and will be passed through
```
sfc-gh-mgilkey, line 200: Trivial: Start with "A", i.e. "A specific query request ID can be ...".

sfc-gh-dwu: Yes I can fix, anything else? Thanks

util.go
```diff
@@ -3,10 +3,31 @@
 package gosnowflake
 
 import (
+	"context"
 	"database/sql/driver"
+	"github.com/google/uuid"
 	"time"
 )
 
+type contextKey string
```
sfc-gh-anezvigin, line 12: Seems unnecessary/premature.

sfc-gh-anezvigin (approved):
Implementation LGTM.

util.go
sfc-gh-dwu: One of the linters complained about using bare `string` and it said to make a specific type to avoid collisions

sfc-gh-dwu: GoLint "should not use basic type string as key in context"

sfc-gh-anezvigin: Oh, interesting. I didn't that of that. Disregard my comment. Thanks linter!

sfc-gh-jbahk (approved):

PR #343, opened 2020-10-21
SNOW-209090 Support opening connection with Config struct

sfc-gh-dwu:
Description
Add the ability to open a connection using a specific Config struct.
Over time, additional non-string representation config options will be implemented,
so we would like the ability to specify the Config directly vs always stringify into a DSN.

Testing
All existing unit tests hit new code path.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2020 Snowflake Computing Inc. All right reserved.
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+// InternalSnowflakeDriver is the interface for an internal Snowflake driver
+type InternalSnowflakeDriver interface {
+	Open(dsn string) (driver.Conn, error)
+	OpenWithConfig(ctx context.Context, config Config) (driver.Conn, error)
+}
+
+// Connector creates Driver with the specified Config
+type Connector struct {
+	driver InternalSnowflakeDriver
+	cfg    Config
+}
+
+// NewConnector creates a new connector with the given SnowflakeDriver and Config.
+func NewConnector(driver InternalSnowflakeDriver, config Config) Connector {
+	return Connector{driver, config}
+}
+
+// Connect creates a new connection.
+func (t Connector) Connect(ctx context.Context) (driver.Conn, error) {
+	cfg := t.cfg
+	err := fillMissingConfigParameters(&cfg)
+	if err != nil {
+		return nil, err
+	}
+	return t.driver.OpenWithConfig(ctx, cfg)
+}
+
+// Driver creates a new driver.
+func (t Connector) Driver() driver.Driver {
+	return t.driver
+}
--- a/connector_test.go
+++ b/connector_test.go
@@ -1,0 +1,49 @@
+// Copyright (c) 2020 Snowflake Computing Inc. All right reserved.
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+	"reflect"
+	"testing"
+)
+
+type noopTestDriver struct {
+	config Config
+	conn   *snowflakeConn
+}
+
+func (d noopTestDriver) Open(_ string) (driver.Conn, error) {
+	return nil, nil
+}
+
+func (d noopTestDriver) OpenWithConfig(_ context.Context, config Config) (driver.Conn, error) {
+	d.config = config
+	return d.conn, nil
+}
+
+func TestConnector(t *testing.T) {
+	conn := snowflakeConn{}
+	mock := noopTestDriver{conn: &conn}
+	createDSN("UTC")
+	config, err := ParseDSN(dsn)
+	if err != nil {
+		t.Fatalf("Failed to parse dsn %s", dsn)
+	}
+	connector := NewConnector(mock, *config)
+	connection, err := connector.Connect(context.Background())
+	if err != nil {
+		t.Fatalf("Connect error %s", err)
+	}
+	if connection != &conn {
+		t.Fatalf("Connection mismatch %s", connection)
+	}
+	fillMissingConfigParameters(config)
+	if reflect.DeepEqual(config, mock.config) {
+		t.Fatalf("Config should be equal, expected %v, actual %v", config, mock.config)
+	}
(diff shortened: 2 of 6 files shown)


driver.go
```diff
@@ -15,16 +15,22 @@ type SnowflakeDriver struct{}
...
+	cfg, err := ParseDSN(dsn)
 	if err != nil {
-		sc.cleanup()
 		return nil, err
 	}
+	return d.OpenWithConfig(ctx, *cfg)
+}
+
+// OpenWithConfig creates a new connection with the given config.
+func (d SnowflakeDriver) OpenWithConfig(ctx context.Context, config Config) (driver.Conn, error) {
```
sfc-gh-anezvigin, line 27: For this to be useful outside the driver you'll also need to implement the driver.Connector interface. See: https://golang.org/pkg/database/sql/#OpenDB

This is how we connect to Snowflake using a connector:
https://github.com/snowflakedb/snapps/blob/master/src/nulib/snowflake.go#L186

Also I think this'll need tests. Specifically around making sure that you can create identical connections using `sql.Open()` and `sql.OpenDB()`.

sfc-gh-dwu: Ah, I see, let me make it support that interface.

sfc-gh-anezvigin: JFYI we needed to add a Copy() method to the config for it to work properly. You may not need this but the thing to be careful of is if the config has any pointers or container types like maps or slices.

See:
https://github.com/snowflakedb/snapps/blob/master/src/nulib/gosnowflake/driver.go#L34
https://github.com/snowflakedb/snapps/blob/master/src/nulib/gosnowflake/dsn.go#L60

sfc-gh-dwu: Yep, I realized the `Config` struct has all public fields, some of which are mutable.
I feel like concepts like configuration are meant to be immutable, so perhaps its better to just make all fields private, and only let people use a builder to create instances of config. That way, external users are guaranteed not to mutate the config while the driver is running.

connector.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2020 Snowflake Computing Inc. All right reserved.
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+// SnowflakeDriverInterface is the interface for a Snowflake driver
```
sfc-gh-anezvigin, line 10: I think the naming can be made less redundant since there shouldn't be conflicts within the same package.

Here's what I'm thinking:

```go
type Driver interface { // currently SnowflakeDriverInterface
  Open(...)
  OpenWithConfig(...)
}

type Connector struct { // currently SnowflakeConnector
  driver Driver
}

// We can't actually change this due to backwards compatibility. "SnowflakeDriver" is
// redundant and now we're stuck with the name unless we bump a major driver version.
type RestfulDriver struct {} // currently SnowflakeDriver
```

connector_test.go
```diff
@@ -0,0 +1,50 @@
...
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+	"reflect"
+	"testing"
+)
+
+type snowflakeMockDriver struct {
```
sfc-gh-anezvigin, line 12: Can we call this `noopTestDriver` or something similar? We may be introducing a mock driver in a future PR and it could be confused with this.

doc.go
```diff
@@ -125,6 +125,8 @@ A complete connection string looks similar to the following:
 Session-level parameters can also be set by using the SQL command "ALTER SESSION"
 (https://docs.snowflake.com/en/sql-reference/sql/alter-session.html).
 
+Alternatively, use OpenWithConfig() function to create a database handle with the specified Config.
```
sfc-gh-anezvigin, line 128: We should add a test that uses OpenWithConfig() just like a user of the library would. Essentially testing the publicly exposed interface.

connector.go
sfc-gh-dwu: Yep, if everyone is onboard with renaming the existing struct.
This would probably require a major version bump to not break back-compat

connector_test.go
sfc-gh-dwu: Yes

connector.go
sfc-gh-anezvigin: Alternatively we could keep `SnowflakeDriver` as is. `SnowflakeConnector` and `SnowflakeDriverInterface` are both new in this PR so no compat issue with them.

sfc-gh-dwu: Kept the existing names as is, moved all new interface definition to connector.go
Renamed types to be shorter to be more idiomatic Golang

connector_test.go
sfc-gh-dwu: renamed

doc.go
sfc-gh-dwu: added test

connector.go
```diff
@@ -0,0 +1,40 @@
...
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+// Driver is the interface for a Snowflake driver
+type Driver interface {
```
sfc-gh-cshi, line 11: Don't think it is a good idea while importing database/sql/driver we create another 'Driver' interface here. Could you just use a wrapper to SnowflakeDriver outside of gosnowflake package?

sfc-gh-anezvigin: > Don't think it is a good idea while importing database/sql/driver we create another 'Driver' interface here.

Why is that a problem?

sfc-gh-dwu: made this Driver private

connector.go
```diff
@@ -0,0 +1,40 @@
...
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+// InternalSnowflakeDriver is the interface for an internal Snowflake driver
+type InternalSnowflakeDriver interface {
```
sfc-gh-dwu, line 11: I tried to move this interface definition into a private package `gosnowflake/private` but because this need to refer back to `Config`, there was a cyclical dependency unless we also move `Config`, but that would break existing back compat.

So I prefixed this as `InternalSnowflakeDriver`

sfc-gh-anezvigin (approved):

PR #354, opened 2020-12-01
SNOW-209092 Skip Session Close

sfc-gh-jbahk:
### Description
Added a control flow to keep client session alive if parameter enabled.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -219,7 +219,7 @@ func (sc *snowflakeConn) Close() (err error) {
 	sc.stopHeartBeat()
 
 	if !sc.cfg.KeepSessionAlive {
-		err = sc.rest.FuncCloseSession(context.TODO(), sc.rest, sc.rest.RequestTimeout)
+		err = sc.rest.FuncCloseSession(sc.ctx, sc.rest, sc.rest.RequestTimeout)
 		if err != nil {
 			logger.Error(err)
 		}
--- a/connection.go
+++ b/connection.go
@@ -218,9 +218,11 @@ func (sc *snowflakeConn) Close() (err error) {
 	logger.WithContext(sc.ctx).Infoln("Close")
 	sc.stopHeartBeat()
 
-	err = sc.rest.FuncCloseSession(sc.ctx, sc.rest, sc.rest.RequestTimeout)
-	if err != nil {
-		logger.Error(err)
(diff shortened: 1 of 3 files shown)


sfc-gh-anezvigin:
My concern with using `isClientSessionKeepAliveEnabled()` is that if you turn on keep-alive, there's no way to close the session even if you wanted to.

sfc-gh-anezvigin:
Also this is a pretty big behavior change for existing users of the driver that are turning off keep alive. Their sessions will stay open on the GS side.

sfc-gh-jbahk:
I was thinking that users already enable the `client_session_keep_alive` to have the session persist and turn it off, followed by Close() to properly close the session. What do you think?

sfc-gh-cshi:
Seems we don't have end to end test for this session persistence change for go driver. Should we have one?

sfc-gh-anezvigin:
> I was thinking that users already enable the client_session_keep_alive to have the session persist and turn it off, followed by Close() to properly close the session. What do you think?

Problem is that Close() does more than just optionally close the snowflake session. Go's SQL interface uses it to clean up resources (eg: open channels, etc) that it uses to manage the query lifecycle. The user must call Close() when using the driver in long running processes.

So the following scenario would create a resource leak:

- Long running process (eg: API server)
- Sets client_session_keep_alive to false
- Does _not_ call Close() on the sql.DB object

sfc-gh-anezvigin (approved):
👍

connection.go
```diff
@@ -203,9 +203,11 @@ func (sc *snowflakeConn) Close() (err error) {
 	logger.WithContext(sc.ctx).Infoln("Close")
 	sc.stopHeartBeat()
 
-	err = sc.rest.FuncCloseSession(sc.ctx, sc.rest, sc.rest.RequestTimeout)
-	if err != nil {
-		logger.Error(err)
+	if !sc.cfg.KeepSessionAlive {
+		err = sc.rest.FuncCloseSession(context.TODO(), sc.rest, sc.rest.RequestTimeout)
```
sfc-gh-dwu, line 207: `context.Todo()` or should this be `sc.ctx`

sfc-gh-dwu (approved):
lgtm with 1 ctx change

PR #360, opened 2020-12-07
SNOW-209087: Update log framework to use logrus. Allow replacing logg…

sfc-gh-cshi:
…er from outside of driver

    glog does not support structured logs and the current default logger is printing no logs. To switch to printing logs requires
    another build flag which is not suitable for enterprise(production) usage.
    This change is to use logrus to replace glog, so we could actually have 'Error', 'Warning', 'Info' and 'Debug' levels
    in log messages, structured logs.
    Other changes include setting logging levels accordingly in all logging places.
    Add cmd/logger. Add required vendors for logrus

Testing: Local 'make' tests all pass

### Description
Introduce SFLogger interface to expose FieldLogger from logrus with additional APIs needed. Replace glog logging to use proper log level defined in logrus.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-stakeda (approved):
Looks good to me. Just resolve the merge conflict.

sfc-gh-jbahk:
@sfc-gh-cshi did we get this vendor approved by security?

sfc-gh-stakeda:
One suggestion. After merging this, it would probably be better bumping up the minor version to include this change, because this is not a trivial bug fix according to https://semver.org/. It is up you how you translate the change, though. Please work with @ankit-bhatnagar167 . Thanks.

sfc-gh-anezvigin:
The logger is still global which means we won't be able to push state through the call stack. The way we solve this problem is by attaching our structured logger state to the `context.Context` object and extracting in places such as `retryHTTP()`.

sfc-gh-cshi:
> The logger is still global which means we won't be able to push state through the call stack. The way we solve this problem is by attaching our structured logger state to the `context.Context` object and extracting in places such as `retryHTTP()`.

Adding and passing context to logs will be addressed in another PR/JIRA this week. This PR now is for log framework update, not a trivial one already.

PR #362, opened 2020-12-22
SNOW-239637 Logging with context

sfc-gh-cshi:
### Description
Add WithContext support for logging to record context fields. Add session ID in context.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/chunk_arrow.go
+++ b/chunk_arrow.go
@@ -18,7 +18,7 @@ type arrowResultChunk struct {
 }
 
 func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType) ([]chunkRowType, error) {
-	logger.Info("Arrow Decoder")
+	logger.Debug("Arrow Decoder")
 
 	var chunkRows []chunkRowType
 
--- a/connection.go
+++ b/connection.go
@@ -35,6 +35,7 @@ const (
 )
 
 type snowflakeConn struct {
+	ctx             context.Context
 	cfg             *Config
 	rest            *snowflakeRestful
 	SequenceCounter uint64
@@ -81,7 +82,7 @@ func (sc *snowflakeConn) exec(
 		req.Bindings = make(map[string]execBindParameter, len(bindings))
 		for i, n := 0, len(bindings); i < n; i++ {
 			t := goTypeToSnowflake(bindings[i].Value, tsmode)
-			logger.Infof("tmode: %v\n", t)
+			logger.WithContext(ctx).Debugf("tmode: %v\n", t)
 			if t == "CHANGE_TYPE" {
 				tsmode, err = dataTypeMode(bindings[i].Value)
 				if err != nil {
@@ -109,8 +110,8 @@ func (sc *snowflakeConn) exec(
 	if multiCount != nil {
 		req.Parameters = map[string]interface{}{string(MultiStatementCount): multiCount}
 	}
-	logger.Infof("bindings: %v", req.Bindings)
-	logger.Infof("parameters: %v", req.Parameters)
+	logger.WithContext(ctx).Infof("bindings: %v", req.Bindings)
+	logger.WithContext(ctx).Infof("parameters: %v", req.Parameters)
 
 	headers := make(map[string]string)
 	headers["Content-Type"] = headerContentTypeApplicationJSON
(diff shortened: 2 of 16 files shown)


driver.go
```diff
@@ -105,21 +106,12 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
 	}
 
 	sc.populateSessionParameters(authData.Parameters)
+	sc.ctx = context.WithValue(sc.ctx, SFSessionIDKey, authData.SessionID)
```
sfc-gh-stakeda, line 109: This is a good idea. Aside from this, perhaps it will be good to have the query id too so that each log line can have the query id instead to match queries and logs. The same idea can apply to all other drivers.

sfc-gh-stakeda: Note that enhancement is not necessarily be in this PR, though.

connection.go
```diff
@@ -302,26 +303,26 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
...
-		logger.Infof("number of updated rows: %#v", updatedRows)
+		logger.WithContext(ctx).Infof("number of updated rows: %#v", updatedRows)
 		return &snowflakeResult{
 			affectedRows: updatedRows,
 			insertID:     -1,
 			queryID:      sc.QueryID,
 		}, nil
 	}
-	logger.Info("DDL")
+	logger.Debug("DDL")
```
sfc-gh-stakeda, line 313: Should this be `logger.WithContext(ctx).Debug` ?

location.go
```diff
@@ -83,7 +83,7 @@ func init() {
 	timezones = make(map[int]*time.Location, 48)
 	// pre-generate all common timezones
 	for i := -720; i <= 720; i += 30 {
-		logger.Infof("offset: %v", i)
+		logger.Debugf("offset: %v", i)
```
sfc-gh-stakeda, line 86: If ctx is missing, what will happen?

rows.go
```diff
@@ -46,7 +46,7 @@ type snowflakeRows struct {
 }
 
 func (rows *snowflakeRows) Close() (err error) {
-	logger.Infoln("Rows.Close")
+	logger.Debugln("Rows.Close")
```
sfc-gh-stakeda, line 49: Could `rows` get a context?

location.go
sfc-gh-cshi: Then there will not be context info in log message. I think it should be fine for some log messages in gosnowflake like this one.

driver.go
sfc-gh-cshi: The query Id was obtained too late only after we got response from a request. For now I would think we could correlate query ID with session ID to find the query related logs. Or We could add a local query ID when we start a query. These IDs I think we could sync with the request ID concept numeracy is introducing and we could easily add late in QueryWithContext and ExecWithContext.

rows.go
sfc-gh-cshi: Update. Use context from rows.sc.ctx.

connection.go
sfc-gh-cshi: Not sure if we even need this log message. For now I just change it to Debug level. If we don't see any benefit from this log message in the future - providing progress or event info etc. we could remove it.

sfc-gh-stakeda: That's good point. If no log is required, it should be removed. I believe this was added for more or less debugging. This doesn't look useful.

chunk_arrow.go
```diff
@@ -18,7 +18,7 @@ type arrowResultChunk struct {
 }
 
 func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType) ([]chunkRowType, error) {
-	logger.Info("Arrow Decoder")
+	logger.Debug("Arrow Decoder")
```
sfc-gh-stakeda, line 21: could this have the context too?

sfc-gh-stakeda:
Overall looks good to me. This PR reminds me that the importance of log with context. Other drivers can follow to have more meaning information, session id, request id, query id, for better diagnosis.

sfc-gh-stakeda (approved):

chunk_arrow.go
sfc-gh-cshi: Context is not yet available in arrowResultChunk struct. And entering an internal function is trivial. So I will leave this to future if someone feel they need this log.

PR #363, opened 2020-12-22
SNOW-242502 added jira auto closure

sfc-gh-mkeller:
Automated Jira closure for GH issues.
Same as: snowflakedb/snowflake-sqlalchemy#205
See SNOW-251256 as an example.


sfc-gh-cshi (approved):
LGTM. Seems to me this is the same change in other repos.

PR #364, opened 2021-01-01
SNOW-209089 Add HTTP transport override to support future token accessor

sfc-gh-dwu:
SNOW-209089

### Description
In order to support manual auth token accessor / management, we
need a clean way to intercept HTTP requests and responses and manually
set AUTH header, retry, and reauth.

Add an optional value to the config so that an override Http transport
can be specified. When the Transport is specified, the REST client will
use this transport instead of the default snowflake transport.

Testing
Added unit test to verify that config transport is indeed the one
used by the snowflake restful http client

This is built upon https://github.com/snowflakedb/gosnowflake/pull/343 which allows
opening connection with custom Config

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connector.go
+++ b/connector.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2020 Snowflake Computing Inc. All right reserved.
+
+package gosnowflake
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+// InternalSnowflakeDriver is the interface for an internal Snowflake driver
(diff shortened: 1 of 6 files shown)


sfc-gh-anezvigin (approved):

sfc-gh-cshi (approved):
Looks goos to me

sfc-gh-anezvigin:
image

PR #369, opened 2021-01-19
SNOW-254781: Adding precommit secret scanning hook

sfc-gh-nsfard:
SNOW-254781: Adding precommit secret scanning hook


sfc-gh-sshankar (approved):
Approval

PR #370, opened 2021-01-26
SNOW-212952 Emit Query ID

sfc-gh-jbahk:
### Description
Interface to capture query ID as soon as it's ready using a channel passed through a context.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/statement_test.go
+++ b/statement_test.go
@@ -649,12 +649,10 @@ func TestEmitQueryID(t *testing.T) {
 
 	dest := make([]driver.Value, 2)
 
-	go func(ch chan string) {
-		queryID := <- queryIDChan
-		if queryID == "" {
-			t.Fatal("expected a nonempty query ID")
-		}
-	}(queryIDChan)
+	var queryID string
+	go func(qID *string, ch chan string) {
+		*qID = <-queryIDChan
+	}(&queryID, queryIDChan)
 
 	err = conn.Raw(func(x interface{}) error {
 		stmt, err := x.(driver.ConnPrepareContext).PrepareContext(ctx, fmt.Sprintf("SELECT SEQ8(), RANDSTR(1000, RANDOM()) FROM TABLE(GENERATOR(ROWCOUNT=>%v))", numrows))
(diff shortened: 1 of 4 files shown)


statement_test.go
```diff
@@ -630,3 +631,65 @@ func retrieveRows(rows *sql.Rows, ch chan string) {
...
+		defer rows.Close()
+
+		if err != nil {
+			return err
+		}
+		if queryID = rows.(SnowflakeResult).GetQueryID(); queryID == "" {
+			// if query ID not available, wait using channel passed in
+			queryID = <-queryIDChan
+		}
+		if queryID == "" {
```
sfc-gh-anezvigin, line 671: This won't always test if the channel is used to emit the query ID, right? Is there any way to test that specific case?

Idea: Have the test launch a goroutine before running the query and waiting blocking on the channel there. That's how we're using it internally.

statement_test.go
```diff
@@ -630,3 +631,67 @@ func retrieveRows(rows *sql.Rows, ch chan string) {
...
+			if err != nil {
+				if err == io.EOF {
+					break
+				} else {
+					t.Fatalf("unexpected error: %v", err)
+				}
+			}
+			cnt++
+		}
+		queryID := <- goRoutineChan
```
sfc-gh-cshi, line 685: If no queryId is returned, what is the error the caller has to handle? A test case on such a error situation will be helpful.

statement_test.go
```diff
@@ -630,3 +631,67 @@ func retrieveRows(rows *sql.Rows, ch chan string) {
...
+		if err != nil {
+			return err
+		}
+		rows, err := stmt.(driver.StmtQueryContext).QueryContext(ctx, nil)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		if err != nil {
```
sfc-gh-cshi, line 669: check the same err again?

sfc-gh-cshi (approved):

PR #371, opened 2021-01-27
feat: allow users to configure the logger

amine7536:
### Description
The default logger is using a configuration that might not suit every user. 

This MR adds the ability for users to use their own Logrus logger instead of the default one.

Example :
```go
mylogger := logrus.New()
mylogger.SetOutput(os.Stdout) // change output
mylogger.Add(mySentryHook) // add hooks
mylogger.SetFormatter(&logrus.JSONFormatter{}) // use json formatter

logger := sf.CreateLogger(mylogger)
sf.SetLogger(&logger)
```

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/log.go
+++ b/log.go
@@ -3,11 +3,12 @@ package gosnowflake
 import (
 	"context"
 	"fmt"
-	rlog "github.com/sirupsen/logrus"
 	"io"
 	"path"
 	"runtime"
 	"time"
+
+	rlog "github.com/sirupsen/logrus"
 )
 
 //SFSessionIDKey is context key of session id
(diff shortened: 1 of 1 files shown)


sfc-gh-cshi (commented):
Do not see the need to have this CreateLogger function INSIDE gosnowflake while there is SetLogger. The idea for defaultLogger was there will always be a default logger, but with SetLogger, you could always replace the default one to any logger you want as long as it exposes the same functions.

sfc-gh-jbahk:
Closing this as the thread is stale.

PR #372, opened 2021-01-28
SNOW-209084 Stream Based Chunk Downloader

sfc-gh-jbahk:
### Description
Replace traditional worker based downloader with stream based downloader to be used for JSON format queries. The worker based downloader is still the default for Arrow out of necessity.

Moved a bulk of the chunk downloading/fetching logic as well as the structs to a separate file for modularity. Added multiple unit tests involving downloading and fetching using the stream based downloader. Added integration test enabling stream based downloader. Accessible via context returned from `WithStreamDownloader`.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


chunk_downloader.go
```diff
@@ -0,0 +1,533 @@
...
+		scd.CurrentChunkSize = firstArrowChunk.rowCount
+		if err != nil {
+			return err
+		}
+	}
+
+	// start downloading chunks if exists
+	chunkMetaLen := len(scd.ChunkMetas)
+	if chunkMetaLen > 0 {
+		logger.Debugf("MaxChunkDownloadWorkers: %v", MaxChunkDownloadWorkers)
```
sfc-gh-anezvigin, line 147: Should these loggers all have contexts?

chunk_downloader.go
```diff
@@ -0,0 +1,533 @@
...
+	"math/rand"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+type snowflakeChunkDownloader struct {
```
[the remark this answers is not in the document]
sfc-gh-jbahk: The interface doesn't specify the member variables and it would result in having to use setters and getters to access or modify them as well as implement multiple functions. The internal fork doesn't take into account the worker based downloader so I think it makes sense to put them together.

chunk_downloader.go
```diff
@@ -0,0 +1,533 @@
...
+	sc                 *snowflakeConn
+	ctx                context.Context
```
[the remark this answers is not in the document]
sfc-gh-jbahk: Storing the context object was a feature since, and used in, the worker based downloader.

chunk_downloader.go
sfc-gh-jbahk: Yes, the context seems necessary to log them, but it's not entirely crucial. I can remove or change them.

chunk_downloader.go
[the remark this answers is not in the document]
sfc-gh-anezvigin: The internal one used to have both (that's what the interface allowed for). The only reason we deleted the worker based one is because we didn't have good test coverage for it. I'll message you with details.

> The interface doesn't specify the member variables and it would result in having to use setters and getters to access or modify them as well as implement multiple functions. 

All of the methods where you do `if useStreamDownloader()` can be interface methods.

chunk_downloader.go
sfc-gh-anezvigin: Not having a context object drops call stack information from the logger.

chunk_downloader.go
```diff
@@ -0,0 +1,593 @@
...
+	go func() {
+		var readErr = io.EOF
+
+		defer func() {
+			scd.readErr = readErr
+			close(scd.rowStream)
+		}()
+
+		defer func() {
+			if r := recover(); r != nil {
```
sfc-gh-cshi, line 435: What is this defer function for, can these two defer func merge to one?

util.go
```diff
@@ -12,14 +12,44 @@ import (
 
 type contextKey string
 
-// SnowflakeRequestIDKey is optional context key to specify request id
-const SnowflakeRequestIDKey contextKey = "SNOWFLAKE_REQUEST_ID"
+const (
+	// MultiStatementCount controls the number of queries to execute in a single API call
+	MultiStatementCount paramKey = "MULTI_STATEMENT_COUNT"
+	// AsyncMode tells the server to not block the request on executing the entire query
+	asyncMode paramKey = "ASYNC_MODE_QUERY"
```
sfc-gh-cshi, line 19: Should asyncMode be public so the users of gosnowflake could set this param?

sfc-gh-anezvigin: I think they should all be private with getter/setter functions to access them. Not sure if that reasoning works with the stuff that @sfc-gh-dwu added though.

chunk_downloader.go
sfc-gh-anezvigin: Yea, it doesn't hurt to merge them.

chunk_downloader.go
```diff
@@ -0,0 +1,593 @@
...
+
+func (scd *streamChunkDownloader) hasNextResultSet() bool {
+	return scd.readErr == nil
+}
+
+func (scd *streamChunkDownloader) nextResultSet() error {
+	return scd.readErr
+}
+
+func (scd *streamChunkDownloader) start() error {
```
sfc-gh-anezvigin, line 425: This function is missing a lot of the timing logs we have in the previous implementation. We rely on those logs to track timings and performance. Can they be brought back so we can hook our logger into them?

sfc-gh-anezvigin: @sfc-gh-cshi FYI this was a hard requirement for the structured logging integration. We need to be able to drop timing logs into the downloader.

sfc-gh-anezvigin (approved):

PR #379, opened 2021-02-09
SNOW-212951 Fetch query result with query Id

sfc-gh-cshi:
With a query Id provided in context to fetch the results of that query.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -416,7 +416,7 @@ func prepareJWTToken(config *Config) (string, error) {
 	return tokenString, err
 }
 
-//authenticate with sc.cfg
+// Authenticate with sc.cfg
 func authenticateWithConfig(sc *snowflakeConn) error {
 	var authData *authResponseMain
 	var samlResponse []byte
--- a/auth.go
+++ b/auth.go
@@ -415,3 +415,52 @@ func prepareJWTToken(config *Config) (string, error) {
 
 	return tokenString, err
 }
+
+//authenticate with sc.cfg
+func authenticateWithConfig(sc *snowflakeConn) error {
+	var authData *authResponseMain
+	var samlResponse []byte
+	var proofKey []byte
+	var err error
+	logger.Infof("Authenticating via %v", sc.cfg.Authenticator.String())
+	switch sc.cfg.Authenticator {
+	case AuthTypeExternalBrowser:
+		samlResponse, proofKey, err = authenticateByExternalBrowser(
+			sc.ctx,
+			sc.rest,
+			sc.cfg.Authenticator.String(),
+			sc.cfg.Application,
+			sc.cfg.Account,
+			sc.cfg.User,
+			sc.cfg.Password)
+		if err != nil {
+			sc.cleanup()
+			return err
+		}
+	case AuthTypeOkta:
+		samlResponse, err = authenticateBySAML(
+			sc.ctx,
+			sc.rest,
+			sc.cfg.OktaURL,
+			sc.cfg.Application,
+			sc.cfg.Account,
+			sc.cfg.User,
+			sc.cfg.Password)
+		if err != nil {
+			sc.cleanup()
+			return err
+		}
+	}
+	authData, err = authenticate(
+		sc.ctx,
+		sc,
+		samlResponse,
+		proofKey)
+	if err != nil {
+		sc.cleanup()
+		return err
+	}
+	sc.populateSessionParameters(authData.Parameters)
+	sc.ctx = context.WithValue(sc.ctx, SFSessionIDKey, authData.SessionID)
+	return nil
+}
--- a/chunk_downloader.go
+++ b/chunk_downloader.go
@@ -30,9 +30,9 @@ type chunkDownloader interface {
 	reset()
 	getChunkMetas() []execResponseChunk
 	getQueryResultFormat() resultFormat
+	getRowType() []execResponseRowType
 	setNextChunkDownloader(downloader chunkDownloader)
 	getNextChunkDownloader() chunkDownloader
-	getRowType() []execResponseRowType
 }
 
 type snowflakeChunkDownloader struct {
--- a/connection.go
+++ b/connection.go
@@ -8,7 +8,9 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -61,6 +63,9 @@ type snowflakeConn struct {
 	SQLState        string
 	internal        InternalClient
 }
+
+var queryIDPattern = `[\w\-_]+`
+var queryIDRegexp = regexp.MustCompile(queryIDPattern)
 
 // isDml returns true if the statement type code is in the range of DML.
 func (sc *snowflakeConn) isDml(v int64) bool {
@@ -303,6 +308,29 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 }
 
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.queryContextInternal(ctx, query, args)
+	}
+
+	// so there is a queryID in context for which we want to fetch the result
+	if !queryIDRegexp.MatchString(qid) {
+		return nil, &SnowflakeError{
+			Number:  ErrQueryIDFormat,
+			Message: "Invalid QID",
+			QueryID: qid}
+	}
+	// first we will check the status of this particular query to find out if there is result to fetch
+	err := sc.checkQueryStatus(ctx, qid)
+	if err == nil || (err != nil && err.(*SnowflakeError).Number == ErrQueryIsRunning) {
+		// the query is running. Rows object will be returned from here.
+		return sc.buildRowsForRunningQuery(ctx, qid)
+	}
+
+	return nil, err
+}
+
+func (sc *snowflakeConn) queryContextInternal(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
 	logger.WithContext(ctx).Infof("Query: %#v, %v", query, args)
 	if sc.rest == nil {
 		return nil, driver.ErrBadConn
@@ -338,7 +366,6 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
 
 	rows := new(snowflakeRows)
 	rows.sc = sc
-	rows.ChunkDownloader = populateChunkDownloader(ctx, sc, data.Data)
 	rows.queryID = sc.QueryID
 
 	if sc.isMultiStmt(&data.Data) {
@@ -346,6 +373,8 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		rows.addDownloader(populateChunkDownloader(ctx, sc, data.Data))
 	}
 
 	rows.ChunkDownloader.start()
@@ -463,7 +492,7 @@ func (sc *snowflakeConn) handleMultiExec(ctx context.Context, data execResponseD
 	childResults := getChildResults(data.ResultIDs, data.ResultTypes)
 	for _, child := range childResults {
 		resultPath := fmt.Sprintf("/queries/%s/result", child.id)
-		childData, err := sc.getQueryResult(ctx, resultPath)
+		childData, err := sc.getQueryResultResp(ctx, resultPath)
 		if err != nil {
 			logger.Errorf("error: %v", err)
 			code, err := strconv.Atoi(childData.Code)
@@ -509,35 +538,11 @@ func (sc *snowflakeConn) handleMultiExec(ctx context.Context, data execResponseD
 
 func (sc *snowflakeConn) handleMultiQuery(ctx context.Context, data execResponseData, rows *snowflakeRows) error {
 	childResults := getChildResults(data.ResultIDs, data.ResultTypes)
-	var nextChunkDownloader chunkDownloader
-	firstResultSet := false
 
 	for _, child := range childResults {
-		resultPath := fmt.Sprintf("/queries/%s/result", child.id)
-		childData, err := sc.getQueryResult(ctx, resultPath)
+		err := sc.rowsForRunningQuery(ctx, child.id, rows)
 		if err != nil {
-			logger.WithContext(ctx).Errorf("error: %v", err)
-			if childData != nil {
-				code, err := strconv.Atoi(childData.Code)
-				if err != nil {
-					return err
-				}
-				return &SnowflakeError{
-					Number:   code,
-					SQLState: childData.Data.SQLState,
-					Message:  err.Error(),
-					QueryID:  childData.Data.QueryID}
-			}
 			return err
-		}
-		if !firstResultSet {
-			// populate rows.ChunkDownloader with the first child
-			rows.ChunkDownloader = populateChunkDownloader(ctx, sc, childData.Data)
-			nextChunkDownloader = rows.ChunkDownloader
-			firstResultSet = true
-		} else {
-			nextChunkDownloader.setNextChunkDownloader(populateChunkDownloader(ctx, sc, childData.Data))
-			nextChunkDownloader = nextChunkDownloader.getNextChunkDownloader()
 		}
 	}
 	return nil
@@ -581,7 +586,7 @@ func getChildResults(IDs string, types string) []childResult {
 	return res
 }
 
-func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string) (*execResponse, error) {
+func (sc *snowflakeConn) getQueryResultResp(ctx context.Context, resultPath string) (*execResponse, error) {
 	headers := getHeaders()
 	if serviceName, ok := sc.cfg.Params[serviceName]; ok {
 		headers[httpHeaderServiceName] = *serviceName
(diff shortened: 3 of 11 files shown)


connection.go
```diff
@@ -232,6 +232,17 @@ func (sc *snowflakeConn) Prepare(query string) (driver.Stmt, error) {
 }
 
 func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	qid, _ := getResumeQueryId(ctx)
+	if qid == "" {
+		return sc.execContextInternal(ctx, query, args)
+	}
+	actualQuery := "select * from table(result_scan(" + qid + "))"
```
sfc-gh-anezvigin, line 239: Interesting approach. What are the performance characteristics of doing this vs jumping straight to using the `/queries/<query_id>/result` endpoint?

sfc-gh-cshi: /queries/<query_id>/result is more for an in progress query, and this one is to receive result of a completed query within last 24 hours.

sfc-gh-jbahk: An endpoint to consider would be `/monitoring/queries/<query-id>`, that denotes the status of an ongoing query as well.

sfc-gh-anezvigin: @sfc-gh-cshi We use the query ID for in-progress queries. If we need to use the result_scan query we can do it ourselves by just submitting the SQL?

connection.go
```diff
@@ -329,6 +340,16 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 }
 
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.queryContextInternal(ctx, query, args)
+	}
+	actualQuery := fmt.Sprintf("select * from table(result_scan('%s'))", qid)
```
sfc-gh-turbaszek, line 347: Probably we can assume that `qid` will be a safe value, but maybe it would be good to avoid SQL injection by for example doing a basic regex validation?

sfc-gh-cshi: The driver is submitting a query on user's behavior, and they can explore SQL injection in UI too if they want to. So we cannot protect user's own data from user's own misbehavior. The security question we snowflake want to answer is, will there be any SQL that could escape the authorization mechanism we have when there is a malicious user? I think that is not addressed on client side anyway.

sfc-gh-anezvigin: Agreed that we should protect against injection.

> The driver is submitting a query on user's behavior, and they can explore SQL injection in UI too if they want to.

The difference here is that the user doesn't have a chance to sanitize the data themselves. The correct way to sanitize it is a bind variable which the user can't do. We can also just verify that the value is a UUID -- why not?

sfc-gh-turbaszek: Users hurting themselves is one thing (we still should protect them as long as we can) but men in the middle is another thing. I'm not that too familiar with `gosnowflake` so I can be wrong,  but plain queries always makes me uncomfortable.

sfc-gh-jbahk (changes requested):
I think there are a lot of changes here that are unnecessary and not related to this PR. Please keep the PR here minimal to the changes that are relevant.

driver.go
```diff
@@ -29,95 +27,14 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
...
-			sc.cfg.Account,
-			sc.cfg.User,
-			sc.cfg.Password)
-		if err != nil {
-			sc.cleanup()
-			return nil, err
-		}
+	sc, err := buildSnowFlakeConn(ctx, config)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
sfc-gh-cshi: these are needed in tests.

restful.go
```diff
@@ -56,10 +61,11 @@ type snowflakeRestful struct {
 	HeartBeat   *heartbeat
 
 	Connection          *snowflakeConn
+
 	FuncPostQuery       func(context.Context, *snowflakeRestful, *url.Values, map[string]string, []byte, time.Duration, uuid.UUID, *Config) (*execResponse, error)
 	FuncPostQueryHelper func(context.Context, *snowflakeRestful, *url.Values, map[string]string, []byte, time.Duration, uuid.UUID, *Config) (*execResponse, error)
-	FuncPost            func(context.Context, *snowflakeRestful, *url.URL, map[string]string, []byte, time.Duration, bool) (*http.Response, error)
-	FuncGet             func(context.Context, *snowflakeRestful, *url.URL, map[string]string, time.Duration) (*http.Response, error)
+	FuncPost            FuncPostType
+	FuncGet             FuncGetType
```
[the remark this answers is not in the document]
sfc-gh-cshi: these are helpful because I am passing a func of this type as function parameter.

rows.go
```diff
@@ -30,6 +30,7 @@ var (
 type snowflakeRows struct {
 	sc              *snowflakeConn
 	ChunkDownloader chunkDownloader
+	tailChunkDownloader chunkDownloader
```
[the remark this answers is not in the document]
sfc-gh-cshi: How chunkDownloader are organized and maintained belong to rows.

rows.go
```diff
@@ -207,3 +208,14 @@ func (rows *snowflakeRows) waitForAsyncQueryStatus() error {
 	}
 	return nil
 }
+
+func (rows * snowflakeRows) addDownloader(newDL chunkDownloader) error {
```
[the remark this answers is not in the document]
sfc-gh-cshi: so other modules do not have to know the details about how downloaders are organized within rows.

chunk_downloader.go
```diff
@@ -407,15 +407,16 @@ func populateJSONRowSet(dst []chunkRowType, src [][]*string) {
...
-	RowSet         rowSetType
+	ctx            	context.Context
+	id             	int64
+	fetcher        	streamChunkFetcher
+	readErr        	error
+	rowStream      	chan []*string
+	Total          	int64
+	ChunkMetas     	[]execResponseChunk
+	NextDownloader 	chunkDownloader
+	Prev 			chunkDownloader
```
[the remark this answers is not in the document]
sfc-gh-cshi: removed

statement_test.go
```diff
@@ -58,46 +57,63 @@ func TestGetQueryID(t *testing.T) {
...
-	defer db.Close()
+	return db
+}
 
-	queryIDChan := make(chan string, 1)
-	numrows := 100000
-	ctx := WithAsyncMode(context.Background())
-	ctx = WithQueryIDChan(ctx, queryIDChan)
+// End-to-end test to fetch result with queryID
+func TestE2EFetchResultByID(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-jbahk: Can you explain? This has not been resolved.

rows.go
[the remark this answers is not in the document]
sfc-gh-jbahk: Yes, and each of them have a pointer to the next so a tail chunk downloader is not necessary

rows.go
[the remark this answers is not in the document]
sfc-gh-jbahk: I don't think is particularly relevant, and the downloader logic is not available anyway since these functions remain unexported. I am dealing with chunk downloads at the moment so please revert them back.

rows.go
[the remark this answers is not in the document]
sfc-gh-cshi: tail is to save the time to search the tail of the link to add next downloader since the number of chunk could be hundreds.

rows.go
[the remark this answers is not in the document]
sfc-gh-cshi: > I think there are a lot of changes here that are unnecessary and not related to this PR. Please keep the PR here minimal to the changes that are relevant.

Some changes are to better reuse code to avoid duplicating the same code everywhere.
addDownloader() can easily evolve to new logic/data structure without modifying other the callers. The two main processes in driver.go (construct conn object and use sc.cfg for authentication) are the same, the effects of these two functions are very straightforward and should be modularized - and needed when we want to use them in different places

statement_test.go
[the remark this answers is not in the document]
sfc-gh-cshi: This was added back. They were lost by the rebase automatic resolution.

connection.go
```diff
@@ -301,7 +314,37 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
...
 
+// QueryContext when there is
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.queryContextInternal(ctx, query, args)
+	}
+
+	// so there is a queryID in context for which we want to fetch the result
+	pattern := `[\w\-_]+`
```
sfc-gh-anezvigin, line 325: should be regexp.MustCompile() and made global

sfc-gh-cshi: addressed

sfc-gh-jbahk (approved):
LGTM, please rebase, fix the conflicts and run `make` and `go fmt` before merging.

sfc-gh-dwu:
What is the behavior of running `result_scan` over a in-progress QID that still hasn't had completed?
Does it:
1. error out due to QID not completed
2. block until completion
3. read in-progress result

I would like to clarify whether the approach of using `select * from result_scan(qid)` solves the need we have with resuming read result of in-progress query.

connection.go
```diff
@@ -304,7 +320,37 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
...
 
+// QueryContext when there is
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.queryContextInternal(ctx, query, args)
+	}
+
+	// so there is a queryID in context for which we want to fetch the result
+
```
sfc-gh-dwu, line 331: extra line

connection.go
```diff
@@ -304,7 +320,37 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
...
+	// first we will check the status of this particular query to find out if there is result to fetch
+	err := sc.checkQueryStatus(ctx, qid)
+	if err != nil {
+		if err.(*SnowflakeError).Number == ErrQueryIsRunning {
+			return sc.buildRowsForRunningQuery(ctx, qid)
+		}
+		return nil, err
+	}
+
+	//for a finished query
```
sfc-gh-dwu, line 347: space between `//for`

sfc-gh-dwu:
Nvm, looks like we distinguish between in-progress query and completed queries.
Completed queries use `result_scan` and in progress uses `buildRowsForRunningQuery`

Only remaining question would be what if we hit race condition and call `buildRowsForRunningQuery` for a query that is completed?

auth.go
```diff
@@ -396,3 +396,52 @@ func prepareJWTToken(config *Config) (string, error) {
 
 	return tokenString, err
 }
+
+//authenticate with sc.cfg
```
sfc-gh-dwu, line 400: Looks like this method:
1. call `authenticateByExternalBrowser` or `authenticateBySaml`
2. then call `authenticate`
3. then set the auth response session information into the connection's session param and context

What do you think about calling this `authenticateSession(...)` or `initializeSession`?

```
// Authenticate and initialize the the session parameters
func authenticateSession(sc *snowflakeConn) error { ... }
```

connection.go
```diff
@@ -295,7 +311,31 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 	return driver.ResultNoRows, nil
 }
 
+// QueryContext when there is
```
sfc-gh-dwu, line 314: comment is incomplete

connection.go
```diff
@@ -605,6 +622,112 @@ func (sc *snowflakeConn) getQueryResult(ctx context.Context, resultPath string)
 	return respd, nil
 }
 
+// checkQueryStatus return error - nil means the query completed successfully - there is query result to fetch.
+// when check the status too soon there may not be status available, an ErrQueryStatus will be returned.
```
sfc-gh-dwu, line 626: A query goes through multiple stages:
- first, it can be queued for execution waiting on warehouse
- when it is executing, it may not have result yet
- then query can complete with result or error

Is `ErrQueryStatus` returned when the query is currently queued for execution?
And is `ErrQueryIsRunning 279301` when it is executing and have no result yet?

I think we can be more specific with the comment `when check the status too soon...`

util.go
```diff
@@ -24,6 +24,10 @@ const (
 	SnowflakeRequestIDKey contextKey = "SNOWFLAKE_REQUEST_ID"
 	// streamChunkDownload determines whether to use a stream based chunk downloader
 	streamChunkDownload paramKey = "STREAM_CHUNK_DOWNLOAD"
+	//FetchResultByID the queryID of query result to fetch
+	FetchResultByID paramKey = "SF_FETCH_RESULT_BY_ID"
```
sfc-gh-dwu, line 28: the query id is always a UUID and has to be a properly formatted UUID (cus on GS side we generate it as UUID)
I noticed that we store this as a `str` in the context value, can we store it as a `UUID` so that we dont even need to do the regexp validation?

util.go
```diff
@@ -24,6 +24,10 @@ const (
 	SnowflakeRequestIDKey contextKey = "SNOWFLAKE_REQUEST_ID"
 	// streamChunkDownload determines whether to use a stream based chunk downloader
 	streamChunkDownload paramKey = "STREAM_CHUNK_DOWNLOAD"
+	//FetchResultByID the queryID of query result to fetch
```
sfc-gh-dwu, line 27: space after `//`

connection_test.go
```diff
@@ -186,3 +190,118 @@ func TestCloseIgnoreSessionGone(t *testing.T) {
...
+	}
+	if customget != nil {
+		getMethods := new(getMethodSwitch)
+		getMethods.origin = sc.rest.FuncGet
+		getMethods.new = customget
+		ctx = context.WithValue(ctx, debugCustomFunc, getMethods)
+		sc.rest.FuncGet = customget
+	}
+
+	//ctx = WithAsyncMode(ctx)
```
sfc-gh-dwu, line 270: dead code

connection_test.go
```diff
@@ -186,3 +190,118 @@ func TestCloseIgnoreSessionGone(t *testing.T) {
...
+	rows1, err := sc.QueryContext(ctx, "select min(c1) as ms, sum(c1) from ut_conn group by (c1 % 10) order by ms", nil)
+	if err != nil {
+		t.Fatalf("Query failed: %v", err)
+	}
+
+	//time.Sleep(2 * time.Second)
```
sfc-gh-dwu, line 276: dead code

util.go
```diff
@@ -24,6 +24,10 @@ const (
...
+	FetchResultByID paramKey = "SF_FETCH_RESULT_BY_ID"
+	// debugGetFunc carry debug func to override internal GET
+	debugCustomFunc paramKey = "DEBUG_GET_FUNC"
```
sfc-gh-dwu, line 30: there should be a cleaner way to do this, or to intercept by using a HTTP round tripper transport defined in the config

connection.go
```diff
@@ -244,6 +249,17 @@ func (sc *snowflakeConn) PrepareContext(ctx context.Context, query string) (driv
 }
 
 func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.execContextInternal(ctx, query, args)
+	}
+	actualQuery := "select * from table(result_scan(" + qid + "))"
+	newCtx := context.WithValue(ctx, FetchResultByID, "")
+	return sc.execContextInternal(newCtx, actualQuery, args)
```
sfc-gh-dwu, line 258: the API behavior is going to be a little weird if:
- client passes in a SQL query string
- client also passes in non-empty `args`
- client defines a resume query ID in context

We take the resume query id and feed client's passed in `args` into our substituted SQL `select * from table(...`, this feels like it could be trouble. The way that we do the `qid` string interpolation feels like it could also open us up for potential security hole because it doesn't go through any SQL sanitization.

`ExecContext` is meant for executing SQL and not expecting any result rows, in this case, do we need to expose the QID resuming capability if the SQL is already meant to NOT produce any results?

@sfc-gh-anezvigin if we don't need QID resuming capability from `ExecContext`, I'd advocate keeping our change surface area smaller, and NOT exposing the ability for a client to feed a resume QID into `ExecContext`.
Especially since here we're also resuming by issuing yet another job to do `select * from table(result_scan(...))` which effectively generates a brand new separate job, just to resume a prior SQL that has no result set.

sfc-gh-dwu (changes requested):
1. The `debugCustomFunc` way of injecting in and intercepting HTTP calls seems unclean, can we try to find a better way via http round tripper in a config override?
2. Some dead code in the test file and comment cleanup
3. Resume QID should always be UUID, can we use UUID vs str with a regex filter to enforce that the QID is valid
4. `ExecContext` will have unexpected behavior if client feed in named values for SQL substitution AND give us a resume QID

auth.go
sfc-gh-cshi: > Looks like this method:
> 
> 1. call `authenticateByExternalBrowser` or `authenticateBySaml`
> 2. then call `authenticate`
> 3. then set the auth response session information into the connection's session param and context
> 
> What do you think about calling this `authenticateSession(...)` or `initializeSession`?
> The comment can also be better formatted and more descriptive
> 
> ```
> // Authenticate and initialize the the session parameters
> func authenticateSession(sc *snowflakeConn) error { ... }
> ```

This method is doing authenticate with config in sc.cfg, so the user of this function will know they could provide config in sc.cfg to call this method to do authentication. Also I feel that initSession is too general.

auth.go
```diff
@@ -415,3 +415,52 @@ func prepareJWTToken(config *Config) (string, error) {
...
```
sfc-gh-dwu, line 419: `// Authenticate the connection with its config`

connection.go
```diff
@@ -303,6 +308,29 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 }
 
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
+	if qid == "" {
+		return sc.queryContextInternal(ctx, query, args)
+	}
+
+	// so there is a queryID in context for which we want to fetch the result
+	if !queryIDRegexp.MatchString(qid) {
```
sfc-gh-dwu, line 317: `getResumeQueryID` can return a `err`, right now, the `err` is ignored, but we can move this UUID regular expression checking into the `getResumeQueryID` and return the err there.

connection.go
```diff
@@ -303,6 +308,29 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 }
 
 func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	qid, _ := getResumeQueryID(ctx)
```
sfc-gh-dwu, line 311: `_` err is being dropped, we can check it

rows.go
```diff
@@ -209,3 +210,14 @@ func (rows *snowflakeRows) waitForAsyncQueryStatus() error {
 	}
 	return nil
 }
+
+func (rows *snowflakeRows) addDownloader(newDL chunkDownloader) error {
```
sfc-gh-dwu, line 214: interface does not need to return `err` if all code path are returning `nil`
noticed that all callers are also not checking err return

connection.go
```diff
@@ -338,14 +366,15 @@ func (sc *snowflakeConn) QueryContext(ctx context.Context, query string, args []
 
 	rows := new(snowflakeRows)
 	rows.sc = sc
-	rows.ChunkDownloader = populateChunkDownloader(ctx, sc, data.Data)
 	rows.queryID = sc.QueryID
 
 	if sc.isMultiStmt(&data.Data) {
 		err := sc.handleMultiQuery(ctx, data.Data, rows)
```
sfc-gh-dwu, line 372: `handleMultiQuery` sets up the chain of chunk downloaders
the behavior was a bit different from before, where we always created a `ChunkDownloader` and then called `handleMultiQuery` which overwrote the previous chunk downloader

Can we add method comment for `handleMultiQuery` to explain that it sets up the chain of chunk downloaders for multiple query results

connection.go
```diff
@@ -463,7 +492,7 @@ func (sc *snowflakeConn) handleMultiExec(ctx context.Context, data execResponseD
 	childResults := getChildResults(data.ResultIDs, data.ResultTypes)
 	for _, child := range childResults {
 		resultPath := fmt.Sprintf("/queries/%s/result", child.id)
```
sfc-gh-dwu, line 494: url template `/queries/%s/result` is used at least twice, we can pull it out to a private string constant at the top of this file

connection_test.go
```diff
@@ -186,3 +190,105 @@ func TestCloseIgnoreSessionGone(t *testing.T) {
...
+	return customGetQuery(ctx, rest, fullURL, vals, duration, jsonStr)
+}
+
+// this function is going to: 1, create a table, 2, query on this table,
+//      3, fetch result of query in step 2, mock running status and error status of that query.
+func fetchResultByQueryID(t *testing.T, customget FuncGetType, expectedFetchErr *SnowflakeError) error {
+
+	config, _ := ParseDSN(dsn)
+	ctx := context.Background()
+	sc, err := buildSnowFlakeConn(ctx, *config)
```
sfc-gh-dwu, line 240: I noticed that this test function also does:
- buildSnowflakeConn
- authenticateWithConfig
- startHeartBeat

These seem to be doing the same as `driver.OpenWithConfig()`, so is this equivalent to:
```
conn, err := driver.OpenWithConfig(...)
sc := conn.(snowflakeConn)
if customget != nil {
  sc.rest.FuncGet = customget
}
```

If these two are equivalent, then I would prefer we just call `OpenWithConfig()` because that's what the normal client code path uses

sfc-gh-cshi: For connection_test or other tests, heart beat is not needed so I prefer keeping it current way and remove the heart beat part.

connection.go
sfc-gh-cshi: will do

connection.go
sfc-gh-cshi: comments added

rows.go
sfc-gh-cshi: addressed

sfc-gh-dwu (approved):

PR #382, opened 2021-02-17
SNOW-209100 UnExport Fields

sfc-gh-jbahk:
### Description
UnExporting fields

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-dwu:
Is there a test that shows + verifies how this is intended to be used?
The code LGTM, just wonder if a test can be added to demonstrate usage + have test coverage.

client_test.go
```diff
@@ -0,0 +1,43 @@
...
+	sf := new(SnowflakeDriver)
+	c, _ := sf.Open(dsn)
+	conn := c.(*snowflakeConn)
+	httpClient.SetConfig(conn.cfg)
+
+	_, err := httpClient.Post(testPath, getHeaders(), nil, defaultClientTimeout)
+	if err != nil {
+		t.Error(err)
+	}
+}
```
sfc-gh-dwu, line 43: no newline at EOF, does gofmt not add the new line EOF?

retry.go
```diff
@@ -229,6 +231,10 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
 		for k, v := range r.headers {
 			req.Header.Set(k, v)
 		}
+		urlStrLen := len(r.fullURL.String())
+		if urlStrLen >= len(testPath) && r.fullURL.String()[urlStrLen-len(testPath):] == testPath {
```
sfc-gh-dwu, line 235: this is a bit janky, what are alternative ways that do not involve hardcoding a `test path` sentinel value in the production code path?

client_test.go
sfc-gh-jbahk: It does not. I can add that in.

retry.go
sfc-gh-jbahk: I agree, but I couldn't think of a proper GET request that would just bypass this. Any ideas?

sfc-gh-dwu: Would http round tripper fix this? If we defined our own dummy roundtripper that just did nothing and returned no error

client.go
```diff
@@ -0,0 +1,87 @@
...
+func (cli *HTTPClient) SetConfig(config *Config) {
+	cli.cfg = config
+	cli.st = SnowflakeTransport
+	if config.Transporter == nil {
+		if config.InsecureMode {
+			cli.st = snowflakeInsecureTransport
+		} else {
+			ocspResponseCacheLock.Lock()
+			atomic.StoreUint32((*uint32)(&ocspFailOpen), uint32(config.OCSPFailOpen))
+			ocspResponseCacheLock.Unlock()
```
sfc-gh-dwu, line 47: I see we do `lock()` and then `unlock()` but in between we already use `atomic`, so is it still necessary to do the mutex locking if the storeUint32 is already atomic?

sfc-gh-jbahk: I had thought so too, but looking at this: https://github.com/snowflakedb/gosnowflake/blob/master/driver.go#L45, I thought it would be better safe than sorry.

sfc-gh-dwu (approved):
Add EOF on that one file that seem to be missing EOF.
LGTM, thanks!

client.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+)
+
+// InternalClient is implemented by HTTPClient
+type InternalClient interface {
+	Get(context.Context, *url.URL, map[string]string, time.Duration) (*http.Response, error)
+	Post(context.Context, *url.URL, map[string]string, []byte, time.Duration, bool) (*http.Response, error)
+}
+
+type hTTPClient struct {
```
sfc-gh-anezvigin, line 18: Should be `httpClient`

sfc-gh-anezvigin (approved):
Left one minor comment regarding variable naming. LGTM as well.

sfc-gh-dwu:
For reference, when `sql.Open` will return the DB object, which can create wrapper DB Conn structs.
The `snowflakeConn` is actually wrapped inside this struct, and there's a method `Raw()` that exposes the actual `snowflakeConn` as just `interface{}`

This mechanism allows us to use reflection on the interface to retrieve the unexported `internal` client field
image

PR #383, opened 2021-02-22
SNOW-187551 Support Bulk Array Binding

sfc-gh-jbahk:
### Description
Support for bulk array binding for Golang driver using streaming PUTs using the local file system. This can be enabled by setting the parameter `CLIENT_STAGE_ARRAY_BINDING_THRESHOLD` and binding the column arrays via the `Array()` function, e.g. `db.Exec("insert into tbl (?, ?, ?, ?)", Array(intArray), Array(fltArray), Array(boolArray), Array(strArray))`. [Note that the command would work the same even if the parameter is not set] When the binding threshold is reached, we query a PUT command to GS, and process the passed columns and stream them to the local stage. Partial implementation of PUT.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-stakeda:
Please add more context to the PR. Did you already implement PUT command? How much did you test it? It is reliable enough to be used for bulk array binding? Which area do you want me to review?

sfc-gh-stakeda:
no test case is included.

sfc-gh-jbahk:
@sfc-gh-stakeda sorry, I should have added you as a reviewer once I'm finished with this PR. This is still a draft and my next few commits will implement the streaming PUT command.

sfc-gh-jbahk:
Streaming PUT writes to the local stage for array binding and does not require the cloud providers. I have another branch working on the full implementation of PUT (and the cloud providers) so please excuse the lint suppression for unused code.

encrypt_util.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		chunk := make([]byte, blockSize)
+		mode.CryptBlocks(chunk, blockSizeCipher)
+		out.Write(chunk)
+	}
+
+	// encrypt key with QRMK
+	var encKek []byte
+	block, _ = aes.NewCipher(decodedKey)
+	stream := cipher.NewCFBEncrypter(block, ivData) // TODO look into different encryptor
```
sfc-gh-stakeda, line 74: What do you mean by this comment? If you have todo, please file a JIRA ticket.

storage_client.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2021 Snowflake Computing Inc. All right reserved.
+
+package gosnowflake
+
+//lint:file-ignore U1000 Ignore all unused code
```
sfc-gh-stakeda, line 5: why do you want to ignore unused code?

util.go
```diff
@@ -23,6 +28,8 @@ const (
 	SnowflakeRequestIDKey contextKey = "SNOWFLAKE_REQUEST_ID"
 	// streamChunkDownload determines whether to use a stream based chunk downloader
 	streamChunkDownload paramKey = "STREAM_CHUNK_DOWNLOAD"
+
+	useOpenSSLOnlyKey = "SF_USE_OPENSSL_ONLY"
```
sfc-gh-stakeda, line 32: I don't know about this flag. Can you elaborate what this is?

local_util.go
```diff
@@ -0,0 +1,56 @@
+package gosnowflake
```
sfc-gh-stakeda, line 1: Ensure add copyright headers to all files

file_transfer_agent.go
```diff
@@ -0,0 +1,687 @@
...
+		panic("not implemented") // TODO
+	}
+
+	client := sfa.getStorageClient(sfa.stageLocationType)
+	client.uploadOneFileWithRetry(meta)
+	return meta
+}
+
+func (sfa *snowflakeFileTransferAgent) downloadFilesParallel(fileMetas []fileMetadata) {
+	// TODO
```
sfc-gh-stakeda, line 562: I believe this means you have not implemented GET command yet. Can you please file a JIRA and put the links (which is accessible internal only, though)?

sfc-gh-stakeda:
Please note at this moment `CLIENT_STAGE_ARRAY_BINDING_THRESHOLD` cannot be updated by the customers.

sfc-gh-stakeda (commented):
I believe you should split changes into three parts: 1) type converter refactoring, 2) PUT command implementation, 3) Bulk array binding implementation so that we can do focused reviews. In general, lots of moving stuff in one PR is not recommended.

binding_test.go
```diff
@@ -0,0 +1,388 @@
...
+			t.Error("failed to query")
+		}
+	})
+}
+
+// TestBindingBulkArray tests array binding via streaming PUT using the staging
+// process
+func TestBindingBulkArray(t *testing.T) {
+	runTests(t, dsn, func(dbt *DBTest) {
+		dbt.mustExec("ALTER SESSION SET CLIENT_STAGE_ARRAY_BINDING_THRESHOLD = 1")
```
sfc-gh-stakeda, line 328: This is not feasible for public user account. Unless you run this tests internally, can you use the actual `CLIENT_STAGE_ARRAY_BINDING_THRESHOLD` value?

file_transfer_agent.go
```diff
@@ -0,0 +1,687 @@
...
+}
+
+func (sfa *snowflakeFileTransferAgent) downloadOneFile(meta fileMetadata) fileMetadata {
+	// TODO
+	return fileMetadata{}
+}
+
+func (sfa *snowflakeFileTransferAgent) getStorageClient(stageLocationType cloudType) storageUtil {
+	if stageLocationType == local {
+		return &localUtil{}
```
sfc-gh-stakeda, line 576: ah, ok local only...

file_transfer_agent.go
```diff
@@ -0,0 +1,687 @@
...
+	} else if stageLocationType == s3 {
+		return nil // TODO
+	} else if stageLocationType == azure {
+		return nil // TODO
+	} else if stageLocationType == gcs {
+		return nil // TODO
```
sfc-gh-stakeda, line 582: For reference, JDBC and Python Connector use the native SDK for AWS and Azure,  but GCP implementation directly accesses the REST API using general HTTP client. It is up to you but I believe you may want to consider direct REST APIs use for all cloud providers for consistency and flexibility without native SDK dependencies. The downside is you'll have to handle all cases that the SDK does like corner cases.

connection.go
```diff
@@ -87,41 +87,75 @@ func (sc *snowflakeConn) exec(
...
-		req.Bindings = make(map[string]execBindParameter, len(bindings))
-		for i, n := 0, len(bindings); i < n; i++ {
-			t := goTypeToSnowflake(bindings[i].Value, tsmode)
-			logger.WithContext(ctx).Debugf("tmode: %v\n", t)
-			if t == "CHANGE_TYPE" {
-				tsmode, err = dataTypeMode(bindings[i].Value)
-				if err != nil {
-					return nil, err
+	requestID := getOrGenerateRequestIDFromContext(ctx)
+	uploader := bindUploader{
```
sfc-gh-cshi, line 91: 'uploader' is not needed for other exec branches when there is no bind. This also means that arrayBindValueCount and isArrayBind do not have to be part of binUploader. creation of uploader and below added logic should be put under a 'if' branch to reduce unnecessary work for not-bind-upload workflows.

file_util.go
```diff
@@ -0,0 +1,154 @@
...
+	w := gzip.NewWriter(&c)
+	w.Write(*buf) // write buf to gzip writer
+	w.Close()
+	return &c, c.Len()
+}
+
+func (util *snowflakeFileUtil) getDigestAndSize(src **bytes.Buffer) (string, int64) {
+	useOpenSSL := getUseOpenSSL()
+	var m hash.Hash
+	if !useOpenSSL {
```
sfc-gh-cshi, line 35: seems if useOpenSSL is true, digest is not computed at all so this method could directly return nil, src.len

bind_uploader.go
```diff
@@ -0,0 +1,271 @@
...
+	"strconv"
+	"strings"
+)
+
+const (
+	stageName       = "SYSTEM$BIND"
+	createStageStmt = "CREATE TEMPORARY STAGE " + stageName + " file_format=" +
+		"(type=csv field_optionally_enclosed_by='\"')"
+
+	inputStreamBufferSize      = 1024 * 1024 * 10
```
sfc-gh-hchaturvedi, line 21: Suggest adding comments for this. How did we come up with this any references etc.

encrypt_util.go
```diff
@@ -0,0 +1,144 @@
...
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
+	"strconv"
+)
+
+const (
+	blockSize = int(aes.BlockSize / 8)
```
sfc-gh-hchaturvedi, line 18: Can you add comments what are we doing here?

connection.go
sfc-gh-jbahk: Makes sense.

encrypt_util.go
sfc-gh-jbahk: Turns out the the AES package's BlockSize is actually in bytes, which is what I wanted to achieve.

sfc-gh-hchaturvedi: Yup that's what I thought and hence asked.

encrypt_util.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (em *encryptMetadata) getSecureRandom(byteLength int) []byte {
+	token := make([]byte, byteLength)
+	rand.Read(token)
+	return token
+}
+
+func pkcs5Padding(src []byte, blockSize int) []byte {
+	padding := blockSize - len(src)%blockSize
```
sfc-gh-hchaturvedi, line 135: Good job on ensuring a full blocksize of padding if src length is multiple of blockSize!

nit : change from `padding` to `padLength` or something similar.

encrypt_util.go
```diff
@@ -0,0 +1,144 @@
...
+	padtext := bytes.Repeat([]byte{byte(padding)}, padding)
+	return append(src, padtext...)
+}
+
+func pkcs5UnPadding(src []byte) []byte {
+	length := len(src)
+	unpadding := int(src[length-1])
```
sfc-gh-hchaturvedi, line 142: nit: change to `paddedBytesLength` or something similar

encrypt_util.go
```diff
@@ -0,0 +1,144 @@
...
+	em := new(encryptMetadata)
+	ivData := em.getSecureRandom(blockSize)
+	fileKey := em.getSecureRandom(keySize)
+
+	block, err := aes.NewCipher(fileKey)
+	if err != nil {
+		panic(err)
+	}
+	mode := cipher.NewCBCEncrypter(block, ivData)
+	cipherText := make([]byte, aes.BlockSize+src.Len())
```
sfc-gh-hchaturvedi, line 43: Add a comment : 
`sizeOf(initializationVector) + sizeOf(src)`

file_util.go
```diff
@@ -0,0 +1,142 @@
...
+	contentLength      int64
+	encryptionMetadata encryptMetadata
+}
+
+func getByteBufferContent(src **bytes.Buffer) *[]byte {
+	var b bytes.Buffer
+	tee := io.TeeReader(*src, &b) // read src to buf
+	buf, err := ioutil.ReadAll(tee)
+	if err != nil {
+		return nil
```
sfc-gh-hchaturvedi, line 116: Shouldn't we be logging this?

sfc-gh-jbahk: Not really, since this merely returns the byte content within the bytes.Buffer. Golang's bytes.Buffer class does not have a "seek" method so I thought the best alternative was to read and return the contents when dealing with streams.

encrypt_util.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	mode := cipher.NewCBCEncrypter(block, ivData)
+	cipherText := make([]byte, aes.BlockSize+src.Len()) // sizeOf(initializationVector) + sizeOf(src)
+
+	padded := false
+	for {
+		buf := getByteBufferContent(&src)
+		if len(*buf) == 0 {
+			break
+		} else if len(*buf)&aes.BlockSize != 0 {
```
sfc-gh-hchaturvedi, line 46: Should this be `%` and not `&` ?

sfc-gh-hchaturvedi: If it _should_ be `%` then this logic would mean that you pad ONLY if you have residual bytes or only if srcLen is not a multiple of aes.BlockSize whereas you should be padding either ways. In case the srcLen is a multiple of block size you should add one whole block of padding.

sfc-gh-jbahk: You are correct. We are indeed padding only if we have residual bytes that is not a multiple of aes.BlockSize, but why would we have to pad if the stream length is a multiple of the aes.BlockSize?

file_util.go
sfc-gh-hchaturvedi: No I meant that if there's an error during the read, shouldn't we be logging that?

encrypt_util.go
sfc-gh-hchaturvedi: Consider the case when the last byte is 0x01 do you consider it a padding or part of the data?

encrypt_util.go
```diff
@@ -0,0 +1,132 @@
...
+	return json.Unmarshal(data, &ew.snowflakeFileEncryption)
+}
+
+type encryptMetadata struct {
+	key     string
+	iv      string
+	matdesc string
+}
+
+func encryptStream(sfe *snowflakeFileEncryption, src *bytes.Buffer, out *bytes.Buffer, chunkSize int) *encryptMetadata {
```
sfc-gh-hchaturvedi, line 46: A small doc string explaining what this functions is employing would be nice. Something on the lines of 
`... this function encrypts a stream buffer using AES128 block cipher in CBC mode with PKCS5 padding.`

sfc-gh-hchaturvedi:
@sfc-gh-jbahk have we looked into any GO lang APIs that abstract all the padding and encryption out? As in we just give the buffer and we get the cipherText out?

sfc-gh-jbahk:
@sfc-gh-hchaturvedi the `crypto/aes` and `crypto/cipher` packages are the de facto standard packages for Golang.

file_util.go
```diff
@@ -0,0 +1,142 @@
...
+	src, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		panic(err)
+	}
+	buf := bytes.NewBuffer(src)
+	return util.getDigestAndSize(&buf)
+}
+
+// file metadata for PUT/GET
+type fileMetadata struct {
```
sfc-gh-cshi, line 52: I know this is how other driver implemented fileMetadata, but some items are of 'the file to transfer', some are related to 'the transferring', and some are related to the 'transfer worker'. While reorg the code could be hard, I would suggest to check this point: when there is 'sfa', do we still need six 'callback's in fileMeta?

sfc-gh-jbahk: I see. Do you have a suggestion?

sfc-gh-jbahk: I will look into the re-org when I start implementing the S3 PUT

sfc-gh-cshi: I think you could remove callbacks in first commit. They are not used anywhere in this commit. And in next commit you could use sfa.callback directly.

bind_uploader.go
```diff
@@ -3,10 +3,256 @@
...
+		return nil, &SnowflakeError{
+			Number:  ErrBindSerialization,
+			Message: "no binds found in the first column",
+		}
+	}
+
+	_, arr := snowflakeArrayToString(&columns[0])
+	numRows := len(arr)
+	for i := 0; i < numColumns; i++ {
+		_, arr = snowflakeArrayToString(&columns[i])
```
sfc-gh-cshi, line 153: should  we save 'arr' returned snowflakeArrayToString to avoid doing it again in loop at line 169? Also if we have done it for index 0, this loop on line 152 should start from index 1.

sfc-gh-jbahk: Fixed.

sfc-gh-cshi: saving arr to avoid call snowflakeArrayToString again in line 170?

sfc-gh-cshi (approved):

PR #386, opened 2021-02-26
SNOW-209081 Allow bypass authentication by re-using token from accessor

sfc-gh-dwu:
SNOW-209081

### Description
Description
Introduce a no-op authentication type that bypasses the actual authentication
by grabbing the auth token and session from the existing token accessor.

Testing
add new unit test case to verify that post auth http call is not made when auth type is token accessor

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -39,6 +39,8 @@ const (
 	AuthTypeOkta
 	// AuthTypeJwt is to use Jwt to perform authentication
 	AuthTypeJwt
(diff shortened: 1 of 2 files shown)


sfc-gh-anezvigin (approved):
Looks good. Let's give @sfc-gh-sbaron a chance to review as well, to make sure it's suitable for internal uses.

sfc-gh-sbaron (approved):
Looks good!

PR #394, opened 2021-03-04
SNOW-29352 Add Cloud Vendors

sfc-gh-jbahk:
### Description
Add cloud vendors for PUT/GET command.
Because of the dependency management native to Golang wherein the checksum produced in go.sum is recognized by the secret scanner as a "high entropy string", we default to snapps' secret scanner in the .pre-commit-config.yaml.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-hchaturvedi:
@sfc-gh-jbahk  the `snapps-secret-scanner` is going to replace the regular one from now on then?

sfc-gh-jbahk:
@sfc-gh-hchaturvedi yes. Confirmed with @sfc-gh-nsfard that this was the thing to do.

sfc-gh-hchaturvedi (approved):
LGTM!

sfc-gh-nsfard:
@sfc-gh-hchaturvedi @sfc-gh-jbahk confirming i did talk to Jaewan about this. Y'all should customize your own secret scanner if this one doesn't meet your needs in the future. Here's instructions for how to do that: https://snowflakecomputing.atlassian.net/wiki/spaces/CLO/pages/1212940297/Pre-Commit+Secret+Scanning+Instructions#How-to-Customize-the-Secret-Scanners

denyska:
This makes library hard to use in the projects that already depend on other versions of cloud SDKs: i.e. if project already depends on AWS SDK1 then bringing AWS SDK2 will create a bit of dependency confusion. Transitive dependency contribute to the problem as well.

On top of it adding AWS/Azure/GCP dependencies to the projects that don't need them is overreach (security , compliance , transitive dependencies, etc). Instead, have you considered extracted  cloud specific operations into another repos so they can included al-a-carte?

Reducing number of dependencies makes library easier to adopt.

PR #395, opened 2021-03-05
SNOW-294718 Fix OCSP Error on Init Bug

sfc-gh-jbahk:
### Description
Create ocsp_response_cache.json file if it doesn't exist.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -720,7 +720,7 @@ func initOCSPCache() {
 	ocspResponseCacheLock = &sync.RWMutex{}
 
 	logger.Infof("reading OCSP Response cache file. %v\n", cacheFileName)
-	f, err := os.Open(cacheFileName)
+	f, err := os.OpenFile(cacheFileName, os.O_CREATE|os.O_RDONLY, os.ModePerm)
 	if err != nil {
 		logger.Errorf("failed to open. Ignored. %v\n", err)
 		return
--- a/ocsp_test.go
+++ b/ocsp_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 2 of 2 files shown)


ocsp.go
```diff
@@ -720,7 +720,7 @@ func initOCSPCache() {
 	ocspResponseCacheLock = &sync.RWMutex{}
 
 	logger.Infof("reading OCSP Response cache file. %v\n", cacheFileName)
-	f, err := os.Open(cacheFileName)
+	f, err := os.OpenFile(cacheFileName, os.O_CREATE|os.O_RDONLY, os.ModePerm)
```
sfc-gh-hchaturvedi, line 723: Should this be created in write mode?

sfc-gh-hchaturvedi: Also please change this to a warning. We shouldn't be throwing errors if ocsp cache dosn't exist locally..

sfc-gh-jbahk: I believe we are trying to read from the file.

sfc-gh-hchaturvedi: Sounds good. I'd still suggest update the loggere.Error to something similar to what Python has - https://github.com/snowflakedb/snowflake-connector-python/blob/f168a26a251361bac2fd8822636085e7e20132a9/src/snowflake/connector/ocsp_snowflake.py#L501

sfc-gh-hchaturvedi: Here was my concern when I commented on this - https://golang.org/pkg/os/#example_OpenFile .

> If the file does not exist, and the O_CREATE flag is passed, it is created with mode perm (before umask)

Is it possible that a file is created with READONLY permissions which then the code later cannot write to?

sfc-gh-jbahk: Nope, it just means that it returns a reader such that we read contents from it, and cannot write to it using the same file pointer. If we want to write to it, we can open the file again with a WRITEONLY pointer. In fact, we can pass in a READ/WRITE flag to do both, but I didn't see the point of having a write flag we wouldn't use.

sfc-gh-hchaturvedi: Makes sense.

sfc-gh-hchaturvedi (approved):
LGTM - just update the log.

sfc-gh-hchaturvedi:
Forgot about this - but can you add a test to ensure a file creation?

PR #398, opened 2021-03-09
SNOW-291628 Synchronize the session renewal when token expires

sfc-gh-dwu:
### Description
When a request comes back as session expired, we issue a
session renewal and rotate the token in the token accessor.

Multiple requests can hit the same session expired error at
the same time, and trigger multiple renewals of the same stale token.

Use a mutex to synchronize the token renewal, and only issue
one renewal for the stale token.

Testing

New tests with multiple goroutines added to verify concurrency.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/util.go
+++ b/util.go
@@ -120,7 +120,7 @@ type TokenAccessor interface {
 	GetTokens() (token string, masterToken string, sessionID int)
 	SetTokens(token string, masterToken string, sessionID int)
 	Lock() error
-	Unlock() error
+	Unlock()
 }
 
 type simpleTokenAccessor struct {
@@ -140,9 +140,8 @@ func (sta *simpleTokenAccessor) Lock() error {
 	return nil
 }
 
-func (sta *simpleTokenAccessor) Unlock() error {
+func (sta *simpleTokenAccessor) Unlock() {
 	sta.accessorLock.Unlock()
-	return nil
 }
 
 func (sta *simpleTokenAccessor) GetTokens() (token string, masterToken string, sessionID int) {
--- a/util.go
+++ b/util.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"database/sql/driver"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


sfc-gh-dwu:
Going to move the mutex lock & unlock into the token accessor interface so that the driver user can control synchronizing across servers for renewals.

sfc-gh-sbaron (commented):
Code changes look good!

Generally I try to avoid tests that throw a bunch of go routines at a race condition, because you don't really know what order things execute in, so you can't tell whether you're testing what you mean to, and if a new bug crops up, it can end up as a flaky test rather than failing consistently

Especially for TestUnitTokenAccessorRenewSession, wondering if there's a way to break it up into more discrete unit tests, something like:
1) With no competition, renewExpiredSessionToken renews the session token
2) With no competition, if the token in the accessor has changed, renewExpiredSessionToken doesn't renew the session token
3) If one goroutine is already holding the accessor lock, another will wait until it's done. This one is kind of hard with mutexes, but I think if you make a Test Accessor that uses something else for synchronization, it could be doable 

Might be overkill, but wanted to throw it out there

sfc-gh-jbahk (approved):
LGTM!

util.go
```diff
@@ -118,23 +119,41 @@ func toNamedValues(values []driver.Value) []driver.NamedValue {
 type TokenAccessor interface {
 	GetTokens() (token string, masterToken string, sessionID int)
 	SetTokens(token string, masterToken string, sessionID int)
+	Lock() error
+	Unlock() error
```
sfc-gh-anezvigin, line 123: We should either handle the error when unlocking in renewExpiredSessionToken() or remove the error return type.

util_test.go
```diff
@@ -44,6 +47,34 @@ func TestSimpleTokenAccessor(t *testing.T) {
...
+		go func() {
+			// set a random session and token
+			session := rand.Int()
+			sessionStr := strconv.Itoa(session)
+			accessor.SetTokens("t"+sessionStr, "m"+sessionStr, session)
+
+			// read back session and token and verify that invariant still holds
+			token, masterToken, session := accessor.GetTokens()
+			sessionStr = strconv.Itoa(session)
+			if "t"+sessionStr != token || "m"+sessionStr != masterToken {
```
sfc-gh-anezvigin, line 65: This is only testing that the master token and session token are set atomically, which is an implementation detail of a particular accessor. I think a more robust test would be to ensure that only a single session renewal occurs when multiple goroutines encounter expired sessions, but I don't know how hard that would be to mock out.

sfc-gh-anezvigin (approved):

sfc-gh-anezvigin:
Left a couple thoughts, otherwise LGTM.

util.go
sfc-gh-dwu: prefer to remove the error return type in the unlock, will remove

util_test.go
sfc-gh-dwu: The other unit test, `TestUnitTokenAccessorRenewSession` is calling with multiple goroutines competing to issue renewal, and verifying at the end that only 1 renewal actually happened.

sfc-gh-dwu:
Added new tests to verify:
- with no contention, renewals can happen
- with controlled locking, if token is swapped while renewal goroutine is locked, it will check and not renew because it became stale
- contention with many goroutines trying to compete for renewal, only one wins

PR #404, opened 2021-03-25
SNOW-294154 S3 PUT Support Additional Dependencies

sfc-gh-jbahk:
### Description
Support PUT command (stream and file) using S3 SDK, uploading to S3 stage and in turn copying to table.

### Checklist
- [x] Code compiles correctly
- [x] Run ``make fmt`` to fix inconsistent formats
- [x] Run ``make lint`` to get lint errors and fix all of them
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

PR #409, opened 2021-04-12
SNOW-294154 S3 PUT Support

sfc-gh-jbahk:
### Description
S3 PUT and streaming PUT implementation.
PUT documentation: https://docs.snowflake.com/en/sql-reference/sql/put.html
Streaming PUT: use WithFileStream() method to pass file pointer to be read for PUT
Dependencies: https://github.com/snowflakedb/gosnowflake/pull/404

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -6,8 +6,14 @@ package gosnowflake
 
 import (
 	"bytes"
+	"context"
+	"database/sql/driver"
+	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
+	"math"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -16,16 +22,20 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/smithy-go"
 )
 
 type cloudType string
 type commandType string
 
 const (
-	fileProtocol      = "file://"
-	dataSizeThreshold = 64 * 1024 * 1024
-	injectWaitPut     = 0
-	isWindows         = runtime.GOOS == "windows"
+	fileProtocol              = "file://"
+	dataSizeThreshold int64   = 64 * 1024 * 1024
+	injectWaitPut             = 0
+	isWindows                 = runtime.GOOS == "windows"
+	mb                float64 = 1024.0 * 1024.0
 )
 
 const (
@@ -49,7 +59,6 @@ const (
 	errStatus resultStatus = iota
 	uploaded
 	downloaded
-	collision
 	skipped
 	renewToken
 	renewPresignedURL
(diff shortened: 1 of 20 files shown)


storage_client.go
```diff
@@ -13,9 +23,138 @@ type storageUtil interface {
...
+	for i := 0; i < 10; i++ {
+		// retry
+		err := rsu.uploadOneFile(meta)
+		if err != nil {
+			return err
+		}
+		if meta.resStatus == uploaded {
+			for j := 0; j < 10; j++ {
+				utilClass.getFileHeader(meta, meta.dstFileName)
+				if meta.resStatus == notFoundFile {
```
sfc-gh-cshi, line 135: Is there a test case covering this 'if meta.resStatus == notFoundFile' code path? If getFileHeader is done successfully then line 139 is always executed then retryFailed set to true?

file_transfer_agent.go
```diff
@@ -464,45 +640,54 @@ func (sfa *snowflakeFileTransferAgent) uploadFilesParallel(fileMetas []*fileMeta
...
 			for i, meta := range targetMeta {
 				wg.Add(1)
-				go func(k int, m *fileMetadata, errChan chan error) {
+				go func(k int, m *fileMetadata) {
 					defer wg.Done()
-					var uploadErr error
-					results[k], uploadErr = sfa.uploadOneFile(m)
-					errChan <- uploadErr
-				}(i, meta, errChan)
+					results[k], errors[k] = sfa.uploadOneFile(m)
```
sfc-gh-cshi, line 651: For this async implementation, since we are not using any thread here, if there are 1000 files to upload, in the middle the upload failed because token expires, then what will happen? Seems the client will keep sending all these 1000X12M files, then rely on the logic at the end (line 656) to collect error statuses to retry. Should we try to be more efficient here, e.g. at the beginning of uploadOneFile, check if other execution of this uploadOneFile failed with renewToken or renewPresignedURL, then the current upload should not continue. I would suggest to add a test to mock a time out case, and for GCS, a renewPresignedURL so you could test the code path here more thoroughly and compare the performance.

sfc-gh-cshi (changes requested):

storage_client.go
```diff
@@ -13,9 +23,138 @@ type storageUtil interface {
...
+				}
+				retryFailed = true
+				break
+			}
+			if retryFailed {
+				// not found, retry with outer loop
+				continue
+			}
+		}
+		retryFailed = true
```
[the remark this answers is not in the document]
sfc-gh-jbahk: I'm not sure what you mean here. If the retry fails, then we set the result status as error.

s3_util.go
```diff
@@ -2,58 +2,227 @@
...
+				return &fileHeader{
+					digest:             "",
+					contentLength:      0,
+					encryptionMetadata: nil,
+				}
+			} else if ae.ErrorCode() == expiredToken {
+				meta.resStatus = renewToken
+				return nil
+			}
+			meta.resStatus = errStatus
```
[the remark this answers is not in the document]
sfc-gh-jbahk: Good point. I'll add it to `lastError`.

file_transfer_agent.go
sfc-gh-jbahk: The thought was that we try to upload regardless and if it fails because of an expired client or anything of similar nature where it requires the driver to take action before retrying, we'd try to fix the problem before trying to re-upload again. If we short circuit the upload, wouldn't that lead to not uploading files we otherwise could have?

sfc-gh-cshi: error reported for expired token is something that will fail every upload after the expiration. So this is not random thing. We should pause the uploads that are not started, refresh all clients (these clients are created at the same time so they should be expired at this point) then resume the uploads. This is the logic in other drivers. If you prefer to defer the efficiently handling expired token late I think that is also fine.

storage_client.go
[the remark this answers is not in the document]
sfc-gh-jbahk: Fixed

storage_client.go
sfc-gh-jbahk: Fixed

file_transfer_agent.go
```diff
@@ -274,28 +323,80 @@ func (sfa *snowflakeFileTransferAgent) initFileMetadata() error {
...
 
 		if len(sfa.encryptionMaterial) > 0 {
 			for _, meta := range sfa.fileMetadata {
 				meta.encryptionMaterial = sfa.encryptionMaterial[0]
 			}
 		}
+	} else if sfa.commandType == downloadCommand {
+		for _, fileName := range sfa.srcFiles {
+			if len(fileName) > 0 {
+				firstPathSep := strings.Index(fileName, "/")
```
sfc-gh-cshi, line 363: You are using the first path sep to get the file name?, and does '/' work for windows?

sfc-gh-jbahk: We are parsing the result from the GS side which is in this format.

storage_client.go
```diff
@@ -13,9 +23,137 @@ type storageUtil interface {
...
+		}
+		if retry {
+			continue
+		} else {
+			break
+		}
+	}
+	if retry {
+		// could not upload file even after retry
+		meta.resStatus = errStatus
```
sfc-gh-cshi, line 151: If we got one resStatus = errStatus, we should fail the whole PUT right? But I don't see anywhere you check this 'meta.errStatus' outside, only renewToken and renewnPresignedUrl were checked and handled.

sfc-gh-jbahk: As mentioned before, I think we still want to go through with the PUT despite one (or multiple) files fail and return the status of that to the users. The users can then attempt to PUT the files failed files after. If they try to upload the same batch of files, it'll simply return skipped.

sfc-gh-cshi (approved):
Would like to have the code coverage report if possible.

sfc-gh-jbahk:
@sfc-gh-cshi it's at 80%. I'll provide a more final number after the Azure/GCP PUT tests + bulk array binding tests.

PR #421, opened 2021-05-10
SNOW-292862 Additional Array Bind Types

sfc-gh-jbahk:
### Description
Extension of array binds. Adding support for timestamp, time, date and byte array bindings.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -18,6 +18,8 @@ import (
 	"github.com/apache/arrow/go/arrow/decimal128"
 )
 
+const format = "2006-01-02 15:04:05.999999999"
+
 // goTypeToSnowflake translates Go data type to Snowflake data type.
 func goTypeToSnowflake(v driver.Value, tsmode snowflakeType) snowflakeType {
 	switch t := v.(type) {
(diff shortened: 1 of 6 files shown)


prepared_statement_test.go
```diff
@@ -9,17 +9,18 @@ import (
 // TestPreparedStatement creates a basic prepared statement, inserting values
 // after the statement has been prepared
 func TestPreparedStatement(t *testing.T) {
-
 	runTests(t, dsn, func(dbt *DBTest) {
-		dbt.mustExec(createTableSQL)
```
sfc-gh-cshi, line 14: Any reason replace the createTableSQL and insertSQL below? They look like the same

converter.go
```diff
@@ -536,55 +563,142 @@ func Array(a interface{}) interface{} {
...
+			x = x.Add(time.Second * time.Duration(offset))
+			v := fmt.Sprintf("%d", x.Unix()*1000)
+			arr = append(arr, &v)
+		}
+	case reflect.TypeOf(&timeArray{}):
+		t = timeType
+		a := nv.Value.(*timeArray)
+		for _, x := range *a {
+			var v string
+			if stream {
```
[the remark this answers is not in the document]
sfc-gh-jbahk: It's because how GS parses time when it's passed from a stream (aka via a file) and when it's passed directly as an array is different, so the conversions are different.

prepared_statement_test.go
sfc-gh-jbahk: These query constants have now been modified for the tests in bindings_test.go

sfc-gh-cshi (approved):

PR #426, opened 2021-05-11
Fetch and return monitoring data for slow queries

GregOwen:
### Description
Why: when a query is slow, we'd like to have additional information to help us analyze why (we could get this information via a separate request to the QUERY_HISTORY table, but we'd prefer to have it automatically)

This PR:
1. Adds new `monitoringResponse` and `QueryMonitoringData` types to represent the monitoring data we want to return
2. Adds a `monitoring` field and `Monitoring` public getter to `snowflakeResult` and `snowflakeRows` to expose this data
3. For all `ExecContext` and `QueryContext` calls, if the query takes longer than a fixed timeout to execute, fetch the monitoring data for the query and add it to the result

Open questions:
1. There's overlap between `monitoringResponse/QueryMonitoringData` and the existing types `statusResponse/retStatus`: both represent the result returned by a query to `/monitoring/queries/$qid`, but return different views on that data. How should we reconcile these types to avoid duplication?
2. Similarly, there's overlap between `connection.go:monitoring()` and the existing `connection.go:checkQueryStatus()`. How much do we want to reuse code between these two?
3. Currently, the "slow query" threshold is hardcoded at 5 seconds. How should we allow clients to configure this? Make it a `Context` property? (Once we decide on that, I'll add a test)

### Checklist
- [X] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -9,6 +9,7 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -371,6 +372,7 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 				Message:  err.Error(),
 				QueryID:  data.Data.QueryID}
 		}
+		return nil, err
 	}
 
 	// if async exec, return result object right away
@@ -479,7 +481,7 @@ func (sc *snowflakeConn) queryContextInternal(ctx context.Context, query string,
 	}
 
 	rows.ChunkDownloader.start()
-	return rows, nil
+	return rows, err
 }
 
 func (sc *snowflakeConn) Prepare(query string) (driver.Stmt, error) {
--- a/connection.go
+++ b/connection.go
@@ -226,34 +226,33 @@ func (sc *snowflakeConn) monitoring(qid string, runtimeSec time.Duration) (*Quer
 		return nil, nil
 	}
 
-	fullURL := fmt.Sprintf(
-		"%s://%s:%d/monitoring/queries/%s",
-		sc.rest.Protocol, sc.rest.Host, sc.rest.Port, qid,
-	)
-
 	// Bound the GET request to 1 second in the absolute worst case.
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
 
 	headers := make(map[string]string)
-	headers["accept"] = "application/json"
-	headers["User-Agent"] = userAgent
-	headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, sc.rest.Token)
+	param := make(url.Values)
+	param.Add(requestGUIDKey, uuid.New().String())
+	if tok, _, _ := sc.rest.TokenAccessor.GetTokens(); tok != "" {
+		headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, tok)
+	}
+	resultPath := fmt.Sprintf("/monitoring/queries/%s", qid)
+	url := sc.rest.getFullURL(resultPath, &param)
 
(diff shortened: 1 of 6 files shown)


GregOwen:
@sfc-gh-jbahk here's the first of the PRs I mentioned. There are some pieces that I expect we'll need to clean up before we can merge (I left these in "Open questions" in the PR comment); let me know what you think and I'll revise.

sfc-gh-jbahk:
@GregOwen I'll defer this to @sfc-gh-cshi who developed the query monitoring feature.

result.go
```diff
@@ -69,3 +71,7 @@ func (res *snowflakeResult) waitForAsyncExecStatus() error {
 	}
 	return nil
 }
+
+func (res *snowflakeResult) Monitoring() *QueryMonitoringData {
```
sfc-gh-cshi, line 75: GetStatus and Monitoring both use monitoring/.. REST API. These two can share some code. And we prefer fetching monitoring data only when status or monitoring is needed, and we should not return the same saved monitoring when Monitoring is called. What do you think?

sfc-gh-cshi (changes requested):

result.go
GregOwen: > These two can share some code.

Sounds good. How would you like to handle the data structures (`QueryMonitoringData` and `retStatus`)? I think we should keep them separate since they represent different views on the same underlying data (the only field they have in common is `Status`), but I can also merge them into a single struct if you'd prefer.

> we prefer fetching monitoring data only when status or monitoring is needed

Sounds good.

> we should not return the same saved monitoring when Monitoring is called

To make sure I understand, it sounds like you'd prefer each call to `Monitoring` to re-fetch the monitoring data. Is there a reason you don't want to cache the monitoring information and reuse it? Do we expect the monitoring data for a given query to change between calls to `Monitoring`? All of the fields in `QueryMonitoringData` look like they should be fixed once a query has been run.

GregOwen: As a follow-on question, how would we actually implement the re-fetching? At the time that the client calls `.Monitoring()` on a `SnowflakeResult`, we no longer have access to the `snowflakeConn` object that we need to call `FuncGet()`.

GregOwen: After looking at this more closely, I'm not sure there's a good way to expose the monitoring information, since it's fundamentally not part of the golang sql interface. If we return the monitoring data as part of the result (as in the current implementation of this PR), the client needs a way to get access to the query result as a `SnowflakeResult` rather than a generic `Rows` object. If we expose monitoring data via a top-level method on `snowflakeConn`, the client needs a way to get access to the `snowflakeConn` as a `snowflakeConn` rather than a generic `DB` object. In either case, we're exposing something that isn't currently part of the public interface.

GregOwen:
@sfc-gh-cshi  I think we may need to abandon this PR, since it fundamentally requires exposing functionality that is not part of the Go sql interface.

connection.go
```diff
@@ -329,13 +386,24 @@ func (sc *snowflakeConn) ExecContext(ctx context.Context, query string, args []d
 			return nil, err
 		}
 		logger.WithContext(ctx).Debugf("number of updated rows: %#v", updatedRows)
-		return &snowflakeResult{
+		rows := &snowflakeResult{
```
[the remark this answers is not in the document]
GregOwen: Changed this to only add the results when monitoring has been requested by the client

rows.go
```diff
@@ -140,6 +141,10 @@ func (rows *snowflakeRows) GetQueryID() string {
 	return rows.queryID
 }
 
+func (rows *snowflakeRows) Monitoring() *QueryMonitoringData {
```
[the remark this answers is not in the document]
GregOwen: Added a test case

sfc-gh-jbahk:
@GregOwen can you tell me if https://github.com/snowflakedb/gosnowflake/pull/467 would address your needs?

sfc-gh-jbahk:
Closing this as a stale branch. Please reopen if need as you see fit.

PR #434, opened 2021-05-28
Fix StaticCheck Error

sfc-gh-jbahk:
### Description
make lint failing due to staticcheck update regarding ineffective assignment

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connector_test.go
+++ b/connector_test.go
@@ -14,11 +14,11 @@ type noopTestDriver struct {
 	conn   *snowflakeConn
 }
 
-func (d noopTestDriver) Open(_ string) (driver.Conn, error) {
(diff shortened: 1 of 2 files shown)


sfc-gh-mgilkey (commented):
I don't know enough about the underlying functions and parameters to know whether these corrections are correct. For example, I don't know what the "noopTestDriver" is, so I don't know whether it should be passed as a pointer. 

I'd prefer someone with more knowledge review this.

If you cannot find someone better-qualified than I am, then I'll dig deeper and see if I can form an intelligent opinion.

sfc-gh-cshi (approved):

PR #442, opened 2021-06-21
SNOW-365916 Add Privatelink Revocation Check

sfc-gh-jbahk:
### Description
Add privatelink revocation check when building connection.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
@@ -755,6 +756,18 @@ func (sc *snowflakeConn) buildRowsForRunningQuery(ctx context.Context, qid strin
 	return rows, err
 }
 
+func (sc *snowflakeConn) setupOCSPPrivatelink(app string, host string) error {
+	ocspCacheServer := fmt.Sprintf("http://ocsp.%v/ocsp_response_cache.json", host)
+	if err := os.Setenv(cacheServerURLEnv, ocspCacheServer); err != nil {
+		return err
+	}
+	ocspRetryHost := fmt.Sprintf("http://ocsp.%v/retry/", host) + "%v/%v"
+	if err := os.Setenv(ocspRetryURLEnv, ocspRetryHost); err != nil {
+		return err
+	}
+	return nil
+}
+
 func isAsyncMode(ctx context.Context) bool {
 	val := ctx.Value(asyncMode)
 	if val == nil {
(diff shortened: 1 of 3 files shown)


connection.go
```diff
@@ -758,6 +759,14 @@ func (sc *snowflakeConn) buildRowsForRunningQuery(ctx context.Context, qid strin
 	return rows, err
 }
 
+func (sc *snowflakeConn) setupOCSPPrivatelink(app string, host string) error {
```
sfc-gh-hchaturvedi, line 762: This change is not enough. It's not only the cache server URL that changes but also the actual OCSP responder URL that changes as well. 

`ocspRetryHost := fmt.Sprintf("http://ocsp.%v/retry/\%v/\%v", host)`
`if err: = os.Setenv("ocspRetryURLEnv, ocspRetryHost); err != nil {`
`            return err`
`}`

The above change would export an environment variable `ocspRetryURLEnv` with the value something like : 
`http://ocsp.abc.privatelink.snowflakecomputing.com/retry/%v/%v`

Once this is done, in ocsp.go within `getRevocationStatus` you would want to change line `489` to first check if `ocspRetryURLEnv` is set. If it is, you should extract that value and create retry URL from it like so : 

` hostname := fmt.Sprintf(ocspRetryURLEnv, url.Parse(ocspHost).Hostname(), ocspReq)`

this would set the ocspRetryURL to something like : 
`http://ocsp.abc.privatelink.snowflakecomputing.com/retry/msocsp.com/<Base64 OCSP request>`

connection.go
```diff
@@ -755,6 +756,18 @@ func (sc *snowflakeConn) buildRowsForRunningQuery(ctx context.Context, qid strin
...
+	ocspCacheServer := fmt.Sprintf("http://ocsp.%v/ocsp_response_cache.json", host)
+	if err := os.Setenv(cacheServerURLEnv, ocspCacheServer); err != nil {
+		return err
+	}
+	ocspRetryHost := fmt.Sprintf("http://ocsp.%v/retry/", host) + "%v/%v"
```
sfc-gh-hchaturvedi, line 764: do you need some kind of escaping here for `"%v/%v"` ?

sfc-gh-jbahk: Nope! It's treated as a literal there.

connection_test.go
```diff
@@ -314,3 +315,20 @@ func fetchResultByQueryID(t *testing.T, customget FuncGetType, expectedFetchErr
...
+		Password: "testpassword",
+		Host:     "testaccount.us-east-1.privatelink.snowflakecomputing.com",
+	})
+	if err != nil {
+		t.Error(err)
+	}
+	ocspURL := os.Getenv(cacheServerURLEnv)
+	expectedURL := "http://ocsp.testaccount.us-east-1.privatelink.snowflakecomputing.com/ocsp_response_cache.json"
+	if ocspURL != expectedURL {
+		t.Errorf("expected: %v, got: %v", expectedURL, ocspURL)
```
sfc-gh-hchaturvedi, line 332: For completion check the retry url as well. Also is there a way you can test the retry url generated with the fields populated?

sfc-gh-hchaturvedi (approved):
LGTM

sfc-gh-hchaturvedi:
Please make sure the tests pass before merging.

PR #443, opened 2021-06-21
SNOW-81453 Inband Telemetry

sfc-gh-jbahk:
### Description
Create inband telemetry.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/telemetry.go
+++ b/telemetry.go
@@ -38,8 +38,8 @@ func (st *snowflakeTelemetry) addLog(data *telemetryData) error {
 		return fmt.Errorf("telemetry disabled; not adding log")
 	}
 	st.mutex.Lock()
-	defer st.mutex.Unlock()
 	st.logs = append(st.logs, data)
+	st.mutex.Unlock()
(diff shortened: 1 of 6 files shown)


telemetry.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (st *snowflakeTelemetry) sendBatch() error {
+	if !st.enabled {
+		return fmt.Errorf("telemetry disabled; not sending log")
+	}
+	type telemetry struct {
+		Logs []*telemetryData `json:"logs"`
+	}
+	s := &telemetry{st.logs}
```
sfc-gh-cshi, line 58: sendBatch can be called from outside and not just from addLog, so there should be a lock protection on st.log. Otherwise there might be some logs added but get lost since at the end of this sendBatch, st.logs is cleaned. 

Should we consider doing a sendBatch when the connection is closed? Otherwise we are expecting other developers to take care when to sendBatch.

sfc-gh-cshi (approved):

PR #444, opened 2021-06-25
SNOW-302103 Disable OCSP Cache Errors in Logs

sfc-gh-jbahk:
### Description
Demoted OCSP cache errors to warnings to declutter logs.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -266,6 +266,9 @@ func decodeCertIDKey(k *certIDKey) string {
 }
 
 func checkOCSPResponseCache(encodedCertID *certIDKey, subject, issuer *x509.Certificate) *ocspStatus {
+	if strings.EqualFold(os.Getenv(cacheServerEnabledEnv), "false") {
+		return &ocspStatus{code: ocspNoServer}
+	}
 	ocspResponseCacheLock.RLock()
 	gotValueFromCache := ocspResponseCache[*encodedCertID]
 	ocspResponseCacheLock.RUnlock()
@@ -641,7 +644,6 @@ func validateWithCache(subject, issuer *x509.Certificate) (*ocspStatus, []byte, 
 
 func downloadOCSPCacheServer() {
 	if strings.EqualFold(os.Getenv(cacheServerEnabledEnv), "false") {
-		logger.Infof("cache server disabled. skipping downloading OCSP Cache.")
 		return
 	}
 	ocspCacheServerURL := os.Getenv(cacheServerURLEnv)
@@ -716,6 +718,9 @@ func overrideCacheDir() {
 
 // initOCSPCache initializes OCSP Response cache file.
 func initOCSPCache() {
+	if strings.EqualFold(os.Getenv(cacheServerEnabledEnv), "false") {
+		return
+	}
 	ocspResponseCache = make(map[certIDKey][]interface{})
 	ocspResponseCacheLock = &sync.RWMutex{}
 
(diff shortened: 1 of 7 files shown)


sfc-gh-hchaturvedi:
I see the errors are mostly converted to Warnings. Is this equivalent to DEBUG logs in other drivers or can you log them at a DEBUG level as well?

sfc-gh-hchaturvedi:
Also I think if you're trying to address the problem of the driver spewing cache related logs even if the cache is disabled, we should focus on why the cache is still being processed even when it is disabled. With your fix we might remove logs but the problem of the cache being accessed even after being disabled would still exist.

ocsp.go
```diff
@@ -850,12 +852,11 @@ func writeOCSPCacheFile() {
...
 	j, err := json.Marshal(buf)
 	if err != nil {
-		logger.Errorf("failed to convert OCSP Response cache to JSON. ignored.")
+		logger.Debugf("failed to convert OCSP Response cache to JSON. ignored.")
 		return
 	}
-	err = ioutil.WriteFile(cacheFileName, j, 0644)
-	if err != nil {
-		logger.Errorf("failed to write OCSP Response cache. err: %v. ignored.\n", err)
+	if err = ioutil.WriteFile(cacheFileName, j, 0644); err != nil {
```
sfc-gh-hchaturvedi, line 858: May be put a check here if the cache is enabled, if not we don't really need to write it and create a DEBUG log either.

ocsp.go
```diff
@@ -815,28 +815,30 @@ func extractOCSPCacheResponseValue(cacheValue []interface{}, subject, issuer *x5
 // writeOCSPCacheFile writes a OCSP Response cache file. This is called if all revocation status is success.
 // lock file is used to mitigate race condition with other process.
 func writeOCSPCacheFile() {
+	if strings.EqualFold(os.Getenv(cacheServerEnabledEnv), "false") {
+		logger.Infof("skipping writing OCSP Cache.")
```
sfc-gh-hchaturvedi, line 819: I would make this a little more prominent, because this is going to be the only line that would signify that the cache is turned off. Also turning the cache off can have major impacts so this should be made a little more prominent.

ocsp.go
sfc-gh-jbahk: If the cache is disabled, this code path won't be hit as per the beginning of the function.

ocsp.go
sfc-gh-jbahk: Noted.

sfc-gh-hchaturvedi:
Our main focus should be to guard all occurrences of cache usage with an if block to prevent access if cache has been disabled. Please go through ocsp.go and add the guarding code - this includes fetching OCSP responses from cache.

sfc-gh-hchaturvedi:
A good test can be - once you have disabled cache you should not be seeing any cache related logs at ANY level - info or debug. I'd suggest you to do that test once before next review.

sfc-gh-hchaturvedi:
Looks like in your current fix the ocsp response cache would still be referred even if it is disabled.  Look for function `validateWithCache` and trace how it's being called.

sfc-gh-hchaturvedi:
> Looks like in your current fix the ocsp response cache would still be referred even if it is disabled. Look for function `validateWithCache` and trace how it's being called.

This still applies. Please review the code for all cache access.

sfc-gh-hchaturvedi:
Looks like all areas are covered now. However I see a different problem - we have put `info` level logs everywhere ocsp cache is being accessed - so every time a driver hits OCSP code it will riddle the logs with these `skipping cache` log statements. 

Let us change every such log statement BUT the first one at the entry of this codepath to DEBUG level.  That should suffice IMO.

sfc-gh-hchaturvedi (approved):
LGTM.

PR #446, opened 2021-06-29
Make snowflakeConn public to allow efficient async queries from lambda functions

rleighton:
### Description
We use lambda functions to implement an async query interface. Roughly the flow is:
 1. make async query, return query id to client
 2. client polls for status of query using query id
 3. when done, client uses query id to get results/errors

Step 2) above is currently  implemented using a sql query which is:
* Slow
* Uses compile/planning resources that can significantly load a cluster when there are many queries in flight

This driver has a method that implements step 2) called  `checkQueryStatus`  but it is private. By making this method public step 2) can be done w/out executing a sql query.

This change was tested in our application.

### Checklist
- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -263,7 +263,7 @@ func getHeaders() map[string]string {
 // Used to authenticate the user with Snowflake.
 func authenticate(
 	ctx context.Context,
-	sc *snowflakeConn,
+	sc *SnowflakeConn,
 	samlResponse []byte,
 	proofKey []byte,
 ) (resp *authResponseMain, err error) {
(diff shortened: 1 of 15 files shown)


rleighton:
Any feedback? This is a really small change and allows for robust implementation via lambda functions.

rleighton:
@sfc-gh-hchaturvedi Note that the python api exposes this method. 

 https://docs.snowflake.com/en/user-guide/python-connector-api.html#get_query_status

rleighton:
Another note, rather than having `checkQueryStatus` return nil if a query is complete w/out error, return a struct with the data from `query_history_by_user`

sfc-gh-jbahk:
Hi @rleighton , thanks for your feedback. At this time we don't intend on making the snowflakeConn struct public for security reasons. We do offer a way to retrieve the query ID before it has finished querying and have any operation with its result (either Row or Result) to hang until the specific function has completed running. Please let me know if this suits your needs.

rleighton:
What is needed is a fast call , given a query id, that will tell you if the
query is done.

Another option, cleaner than exposing the struct is an interface with the
extra method. Are you ok with that?

rleighton:
@sfc-gh-jbahk  Closing in favor of https://github.com/snowflakedb/gosnowflake/pull/458

PR #448, opened 2021-06-29
SNOW-373921: Update to use newer account format name for the new Orga…

sfc-gh-mgilkey:
…nizations feature in Snowflake.

Because this is an area I don't know well, and because we need reviewers with knowledge of both Golang and the new Organizations feature, I would like ALL of you (Doug, Natasha, and Jaewan) to review this.

### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -80,9 +80,8 @@ The following connection parameters are supported:
 
 	* passcode: Specifies the passcode provided by Duo when using multi-factor authentication (MFA) for login.
 
-	* passcodeInPassword: false by default. Set to true if the MFA passcode is
-		embedded in the login password. Appends the MFA passcode to the end of the
-		password.
+	* passcodeInPassword: false by default. Set to true if the MFA passcode is embedded
+		in the login password. Appends the MFA passcode to the end of the password.
 
 	* loginTimeout: Specifies the timeout, in seconds, for login. The default
 		is 60 seconds. The login request gives up after the timeout length if the
--- a/doc.go
+++ b/doc.go
@@ -24,23 +24,20 @@ Use the Open() function to create a database handle with connection parameters:
 
 	db, err := sql.Open("snowflake", "<connection string>")
 
-The Go Snowflake Driver supports the following connection syntaxes (or data source name (DSN) formats).
-The first three lines use the newer format (supported by the Organizations feature of Snowflake).
-The last three lines use the older format.
+The Go Snowflake Driver supports the following connection syntaxes (or data source name (DSN) formats):
 
-	* username[:password]@<organization_name>-<account_name>/dbname/schemaname[?param1=value&...&paramN=valueN]
-	* username[:password]@<organization_name>-<account_name>/dbname/[?param1=value&...&paramN=valueN]
-	* username[:password]@hostname:port/dbname/schemaname?account=<organization_name>-<account_name>[&param1=value&...&paramN=valueN]
-
-	* username[:password]@accountname/dbname/schemaname[?param1=value&...&paramN=valueN]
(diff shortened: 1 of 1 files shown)


natalia-ormiston (approved):
LGTM.

doc.go
```diff
@@ -24,7 +24,13 @@ Use the Open() function to create a database handle with connection parameters:
 
 	db, err := sql.Open("snowflake", "<connection string>")
 
-The Go Snowflake Driver supports the following connection syntaxes (or data source name (DSN) formats):
+The Go Snowflake Driver supports the following connection syntaxes (or data source name (DSN) formats).
```
sfc-gh-dyoshinaga, line 27: In general, it might be better to use the placeholder `<account_identifier>` with a link to https://docs.snowflake.com/en/user-guide/admin-account-identifier.html

Rather than having copies of this information in several different places, we should rely on https://docs.snowflake.com/en/user-guide/admin-account-identifier.html to cover most of the information and then provide a couple of examples to illustrate what that means in the context of the Go driver.

doc.go
```diff
@@ -24,7 +24,13 @@ Use the Open() function to create a database handle with connection parameters:
...
+The first three lines use the newer format (supported by the Organizations feature of Snowflake).
```
sfc-gh-dyoshinaga, line 28: Rather than putting the burden on the reader to count the number of lines, you should split this up into two separate lists.

```
If your account is under an organization
(link to an explanation of how you find this out),
use the following formats for the connection string:

- ...
- ...
- ...

Otherwise, use the following formats for the connection string:

- ...
- ...
- ...

```

doc.go
```diff
@@ -33,10 +39,10 @@ The Go Snowflake Driver supports the following connection syntaxes (or data sour
 where all parameters must be escaped or use Config and DSN to construct a DSN string.
 
 The following example opens a database handle with the Snowflake account
-myaccount where the username is jsmith, password is mypassword, database is
+my_organization-my_account where the username is jsmith, password is mypassword, database is
```
sfc-gh-dyoshinaga, line 42: I would separate these out

```
with the Snowflake account
my_account under the organization
my_organization
```

doc.go
```diff
@@ -46,11 +52,17 @@ The connection string (DSN) can contain both connection parameters (described be
...
 
 	* account <string>: Specifies the name of your Snowflake account, where string is the name
-		assigned to your account by Snowflake. In the URL you received from
-		Snowflake, your account name is the first segment in the domain (e.g.
-		abc123 in https://abc123.snowflakecomputing.com). This parameter is
-		optional if your account is specified after the @ character. If you
-		are not on us-west-2 region or AWS deployment, then append the region
+		assigned to your account by Snowflake.
+
+		For newer accounts, and for older accounts converted to use the Organizations feature,
```
sfc-gh-dyoshinaga, line 57: "Newer" vs. "older" is not a distinction that's going to be meaningful a year from now.
We should just point to https://docs.snowflake.com/en/user-guide/admin-account-identifier.html and leave it to that page to explain the different formats for specifying the account identifier.

sfc-gh-dyoshinaga (approved):
Just some suggestions

doc.go
sfc-gh-mgilkey: Yes! I need a link to the admin-account-identifier page. Done. Thank you.

doc.go
sfc-gh-mgilkey: Done.

doc.go
sfc-gh-mgilkey: Updated.

PR #458, opened 2021-07-27
Add SnowflakeConnectionAPI interface to allow efficient async queries from lambda functions

rleighton:
### Description
We use lambda functions to implement an async query interface. Roughly the flow is:
 1. make async query, return query id to client
 2. client polls for status of query using query id
 3. when done, client uses query id to get results/errors

Step 2) above is currently  implemented using a sql query which is:
* Slow
* Uses compile/planning resources that can significantly load a cluster when there are many queries in flight

This driver has a method that implements step 2) called  `checkQueryStatus`  but it is private. By making this method publicly accessible  step 2) can be done w/out executing a sql query.

Note that the python api exposes this method.

 https://docs.snowflake.com/en/user-guide/python-connector-api.html#get_query_status

This change was tested in our application.

### DIFF
--- a/async_test.go
+++ b/async_test.go
@@ -4,114 +4,92 @@ package gosnowflake
 
 import (
 	"context"
-	"database/sql"
 	"fmt"
 	"testing"
 )
 
 func TestAsyncMode(t *testing.T) {
-	var db *sql.DB
-	var err error
-
-	if db, err = sql.Open("snowflake", dsn); err != nil {
-		t.Fatalf("failed to open db. %v, err: %v", dsn, err)
-	}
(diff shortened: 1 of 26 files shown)


rleighton:
@sfc-gh-jbahk  can you please take a look?

serbantanasa (approved):

rleighton:
You can see this working in the video in this LinkedIn post: https://www.linkedin.com/feed/update/urn:li:activity:6829415977504043009/

rleighton:
Closing in favor of https://github.com/snowflakedb/gosnowflake/pull/467

PR #518, opened 2021-12-28
WithArrowRecordChan: channel to retrieve arrow.Records

ddl-giuliocapolino:
### Description
For performance and ease-of-use purposes, users may need to access data in Arrow.Record format directly from queries, rather than getting it in sql.Rows format and having to covert it manually. This can be achieved via a channel passthrough via Go's context, which is consistent with other workflows in gosnowflake. This PR covers the impl to get these records via a buffered channel: gosnowflake writes arrays of records to the channel as it gets them from the download workers.

Furthermore, the data retrieved from Snowflake is stored in arrow.Record objects. The data stored in these objects, however, is not in a format that can be sent to a client via the ArrowRecordChan. In order to transform it to its final format, we add a converter along with tests. The conversion documentation can be found [here](https://docs.google.com/document/d/1giKZmbMf5mnxTTnNL3-L4CQXhV2Zl_FwTAq64qeesTo/edit?usp=sharing).

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


sfc-gh-jbahk:
Thank you for the PR! It looks good, but we're actually thinking we can tweak the API such that it looks something like this 
```
query := "select seq4() as n from table(generator(rowcount => 100000))"
ctx := WithDistributedBatches(context.Background)
rows, _ := db.QueryContext(ctx, query)
batches, err := rows.(SnowflakeResult).GetBatches()
for batch := range batches {
  batch.fetch()
}
```
so we can distribute the workload and the fetching would contain your Arrow records. Is that plausible? It should be a couple tweaks to the API.

ddl-giuliocapolino:
as discussed, this is not the approach that will be merged

fsaintjacques:
Are there plan to provide this facility? To make it more friendly with arrow's go library, would it make more sense to return an ArrowRecordReader instead?

sfc-gh-jbahk:
@fsaintjacques we are working to provide a similar functionality that will hopefully allow the Arrow records to be more accessible and portable than it already is. The Domino team is working on it. Thanks for your patience!

PR #531, opened 2022-01-21
feat: add config option to disable file transfers

nathanielc:
With this change its now possible to explicitly disable the file
transfer logic via the Config struct.

Alternatively I could put this option on the SnowflakeFileTransferOptions struct https://pkg.go.dev/github.com/snowflakedb/gosnowflake#SnowflakeFileTransferOptions

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -136,6 +136,14 @@ func (sc *snowflakeConn) exec(
 
 	// handle PUT/GET commands
 	if isFileTransfer(query) {
+		if sc.cfg.DisableFileTransfers {
+			return nil, (&SnowflakeError{
+				Number:   code, //TODO do we need a specific code here?
+				SQLState: data.Data.SQLState,
+				Message:  data.Message,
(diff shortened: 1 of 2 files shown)


sfc-gh-jbahk:
Hi, can you tell us why this feature is needed? You don't need to use the file transfer - it's only triggered by certain commands.

nathanielc:
@sfc-gh-jbahk Hi thanks for the reply. We host a cloud service and allow our clients to make connections to their own snowflake endpoints. We use this `gosnowflake` client code to do so. As a result we do not have control over the commands being issued to the client and therefore want a separate code path to ensure that these PUT/GET commands cannot be issued by the client. Otherwise our users could issue PUT command to read local files from our cloud servers and send the contents to their own snowflake instances.

sfc-gh-jbahk:
As mentioned in the original issue, closing this PR.

PR #544, opened 2022-02-08
Arrow Record Distributed Result Batches

ddl-giuliocapolino:
### Description
For performance and ease-of-use purposes, users may need to access data in `arrow.Record` format directly from queries, rather than getting it in `sql.Rows` format and having to covert it manually. This PR covers the implementation and tests to get these records via a new API endpoint called `GetResultBatches`. We use `ResultBatch` objects that can independently fetch chunks of data in `arrow.Record` format. `GetResultBatches` returns such objects in the correct order as to allow the user to compose, in parallel or not, the full data in Arrow format. Below is a sample client request to fetch the data:

```
rbCtx := sf.WithDistributedResultBatches(ctx)
d := sf.SnowflakeDriver{}
db, err := d.Open(dsn)
if err != nil {
	return nil, err
}
snowflakeDb := db.(sf.SnowflakeConnection)

rows, err := snowflakeDb.QueryContext(rbCtx, query, []driver.NamedValue{})
if err != nil {
	return nil, err
}

batches, err := rows.(sf.SnowflakeRows).GetBatches()
if err != nil {
	return nil, err
}
```

For a client using data in arrow.Record format, this can result in more than a 10x speedup in retrieving usable data, as well as provides the client with the ability to manage memory consumption independently.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


monitoring.go
```diff
@@ -116,6 +116,7 @@ type SnowflakeQueryStatus struct {
 // SnowflakeConnection is a wrapper to snowflakeConn that exposes API functions
 type SnowflakeConnection interface {
 	GetQueryStatus(ctx context.Context, queryID string) (*SnowflakeQueryStatus, error)
+	QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error)
```
sfc-gh-jbahk, line 119: Why is this exposure necessary?

rows.go
```diff
@@ -27,6 +27,11 @@ var (
 	maxChunkDownloaderErrorCounter = 5
 )
 
+// SnowflakeRows provides an API for methods exposed to the clients
+type SnowflakeRows interface {
```
sfc-gh-jbahk, line 31: While this makes sense, I think we can consolidate this into SnowflakeResult since this object can and should be able to return both a query ID and its status as well. Asking the user to convert this into separate interfaces seems cumbersome for both sides in terms of usability & maintenance.

converter.go
```diff
@@ -7,6 +7,7 @@ import (
 	"database/sql/driver"
 	"encoding/hex"
 	"fmt"
+	"github.com/apache/arrow/go/arrow/memory"
```
sfc-gh-jbahk, line 10: nit: can this be separated out by a newline apart from the other dependencies?

chunk_test.go
```diff
@@ -387,3 +388,44 @@ func TestWithStreamDownloader(t *testing.T) {
 		}
 	})
 }
+
+func TestWithDistributedResultBatches(t *testing.T) {
```
sfc-gh-jbahk, line 392: This test succinctly demonstrates the API pathway, but is there another test that can be added to demonstrate that these batches are indeed fetched in a distributed manner? 100 rows is not a lot and would probably be fetched within one chunk. This test doesn't demonstrate to me that these batches can be called independently.

More importantly, with these changes, do you see your memory issues improved? Solved?

chunk_downloader.go
```diff
@@ -630,3 +681,28 @@ func copyChunkStream(body io.Reader, rows chan<- []*string) error {
...
+// ResultBatch object encapsulates a function that retrieves a subset of rows in a result set.
+type ResultBatch struct {
+	Rec                *[]array.Record
+	idx                int
+	scd                *snowflakeChunkDownloader
+	funcDownloadHelper func(context.Context, *snowflakeChunkDownloader, int) error
+}
+
+// Fetch returns an array of records representing a chunk in the query
+func (rb *ResultBatch) Fetch() error {
```
sfc-gh-jbahk, line 694: I'm curious as to the design here: shouldn't this return the actual array.Record itself (along with an error), instead of storing the the record as an accessible object within the ResultBatch struct?

sfc-gh-jbahk (changes requested):
This looks great; thank you for your work. I had a couple questions that I'd love to hear your answers on before we proceed.

monitoring.go
ddl-giuliocapolino: Not necessary any longer - we found the following code used for GetStatus and will plan to have clients do this analogously. We'll also add to go.doc a similar example.

```
err = conn.Raw(func(x interface{}) error {
// Execute asynchronous query
rows, _ := x.(driver.QueryerContext).QueryContext(ctx, "select 1", nil)
defer rows.Close()

// Retrieve and check results of the query after casting the result
status := rows.(SnowflakeResult).GetStatus()
```

rows.go
ddl-giuliocapolino: Sounds good. Felt a bit strange to have SnowflakeResult contain an endpoint related to data since Result is usually referred to by an `exec` query that does not return any data, but I agree it's better to expose only one interface. Related to that, would you prefer that the implementation of `GetBatches() ([]*ResultBatch, error)` by `snowflakeResult` returns an error, or returns an empty array (and nil)?

converter.go
ddl-giuliocapolino: whops that slipped in there! thanks :)

chunk_test.go
ddl-giuliocapolino: sounds good - I will have a first PR iteration with the logic changes, and a second only focusing on this test so that we can iterate on the former first! In terms of memory, it's fully up to the user to handle the workers as they wish - in our case we optimized the client side for parallel downloading, but releasing the records as soon as they've been read and written, and we have not seen any memory issue (queried all the way up to 10GB with very little memory pressure but really good performance).

chunk_downloader.go
ddl-giuliocapolino: My initial rationale was to not have a double memory burden but since the object contains `[]*array.Record`, it won't hurt to have it twice so that the user can Fetch it, but also re-use it in the future if necessary (double benefit). Changing it now.

rows.go
ddl-giuliocapolino: I pushed it out with the former (returns an error), but happy to change it to empty array if you rather have that!

chunk_test.go
ddl-giuliocapolino: ^ changed it to 3000 rows which is approximately 6 ResultBatch objects (ran it a few times and, on average, it's 6 chunks of data)

rows.go
sfc-gh-jbahk: Completely understandable - this SnowflakeResult object was meant to be detached from the snowflakeResult (related to Exec) however, and actually denote a literal result. Perhaps a confusing misnomer on our end.

I think the return value seems fine as is for now. I'll let you know if it needs changing or change it myself.

chunk_test.go
sfc-gh-jbahk: The test seems better but is there a way to explicitly illustrate these batches being fetched on in different worker nodes? Something like multiple goroutines, perhaps? Maybe there are some other ways to demonstrate this but at face value, this test doesn't tell me much.

chunk_downloader.go
sfc-gh-jbahk: Does the object need to contain the record? Can it just be a product of the Fetch() function and nothing else? A close analogy would be returning a row object after a query, but you'd need to query the rows again in order to receive the same rows again; the driver won't do that for you.

ddl-giuliocapolino: In this case we do need the `Rec` on the object because we are leveraging the `downloadChunkHelper` which only returns an `error`. It's analogous to how rows are downloaded in chunks within the `downloadChunkHelper` and then are applied to the `scd.Chunks` object, without explicitly being returned. I don't think we can omit that without having a major refactor on how the download workflow happens unfortunately. However, we can make `Rec` non-public (`rec`) if you rather not have it exposed by the object, but we thought it'd be nice for a user to have a pointer to it within the object itself in case it is needed (it's a cheap extra functionality) - up to you though if you'd rather have it lowercase!

sfc-gh-jbahk: Thanks for the quick turnaround. I think the rec object might best be left private so the users can't control it and have fetch act as a getter (as you've done). 

It looks overall very good to me - I'm going to go over these changes today in a design review and come back with final feedback.

chunk_downloader.go
```diff
@@ -630,3 +681,32 @@ func copyChunkStream(body io.Reader, rows chan<- []*string) error {
...
 	return nil
 }
+
+// ResultBatch object represents a chunk of data, or subset of rows, retrievable in array.Record format
+type ResultBatch struct {
+	Rec                *[]array.Record
+	idx                int
+	scd                *snowflakeChunkDownloader
+	funcDownloadHelper func(context.Context, *snowflakeChunkDownloader, int) error
+}
```
sfc-gh-jbahk, line 691: Can we expose some metadata such as rowcount and size information that we retrieve from the server?

chunk_downloader.go
ddl-giuliocapolino: sounds good. I will change it to private later this evening along with the clearer test. Consider it done for your design review!

chunk_downloader.go
ddl-giuliocapolino: adding metadata for rowcount. For size it does not seem to be achievable given that we read the arrow.Records right away and those do not contain any metadata on size themselves.

chunk_test.go
ddl-giuliocapolino: just updated the test with clear download workers - it is now checking also that the metadata received (rowcount) is correct :)

chunk_downloader.go
```diff
@@ -630,3 +683,43 @@ func copyChunkStream(body io.Reader, rows chan<- []*string) error {
...
+func usesDistributedBatches(ctx context.Context) bool {
+	val := ctx.Value(distributedResultBatches)
+	if val == nil {
+		return false
+	}
+	a, ok := val.(bool)
+	return a && ok
+}
+
+func countResultBatchRows(recs *[]array.Record) int {
```
sfc-gh-jbahk, line 719: Can we move this to the test file?

chunk_downloader.go
```diff
@@ -630,3 +683,43 @@ func copyChunkStream(body io.Reader, rows chan<- []*string) error {
 	}
 	return nil
 }
+
+// ResultBatch object represents a chunk of data, or subset of rows, retrievable in array.Record format
+type ResultBatch struct {
```
sfc-gh-jbahk, line 688: ArrowBatch

chunk_downloader.go
```diff
@@ -245,6 +259,33 @@ func getChunk(
 	return newRetryHTTP(ctx, scd.sc.rest.Client, http.NewRequest, u, headers, timeout).execute()
 }
 
+func (scd *snowflakeChunkDownloader) startDistributedBatches() error {
```
sfc-gh-jbahk, line 262: startArrowBatches()

result.go
```diff
@@ -13,10 +13,11 @@ const (
 	QueryFailed queryStatus = "queryFailed"
 )
 
-// SnowflakeResult provides the associated query ID
+// SnowflakeResult provides an API for methods exposed to the clients
 type SnowflakeResult interface {
 	GetQueryID() string
 	GetStatus() queryStatus
+	GetBatches() ([]*ResultBatch, error)
```
sfc-gh-jbahk, line 20: GetArrowBatches()

sfc-gh-jbahk (changes requested):
We had a design review yesterday and there were a couple points --

- the ultimate point we wanted to reach with distributed fetching was being able serialize these ResultBatches so workers across different machines are able to just pick up them independently

- the fact that we require the chunkdownloader as a member in the struct hinders this and requires some refactoring, which we (Snowflake) will get to

- this has a very good use case in and of itself but we envision having to reserve this API/pathway ​so we'd like to merge it with a couple nomenclature tweak

sfc-gh-jbahk:
Local tests pass successfully. Merging.

ddl-giuliocapolino:
Thanks @sfc-gh-jbahk !

stoffeastrom:
Hi 👋  

I was playing around with the latest release where this is included but I can't seem to access the arrow batches? This seems to be completely internal. The example from @ddl-giuliocapolino seems outdated.

```go
rbCtx := sf.WithDistributedResultBatches(ctx)
d := sf.SnowflakeDriver{}
db, err := d.Open(dsn)
if err != nil {
	return nil, err
}
snowflakeDb := db.(sf.SnowflakeConnection)

rows, err := snowflakeDb.QueryContext(rbCtx, query, []driver.NamedValue{})
if err != nil {
	return nil, err
}

batches, err := rows.(sf.SnowflakeRows).GetBatches()
if err != nil {
	return nil, err
}
```
Seems to suggest it could be access through `SnowflakeConnection` but this interface only has

```go
type SnowflakeConnection interface {
	GetQueryStatus(ctx context.Context, queryID string) (*SnowflakeQueryStatus, error)
}
```

but I might misinterpret things :D

stoffeastrom:
@sfc-gh-jbahk Pls see my comment above. Can we get an example of how to use arrow batches directly?

sfc-gh-jbahk:
Hi @stoffeastrom, this is an internal feature and we have not publicly documented it for this reason.

I also don't actively develop the go driver anymore - please tag @sfc-gh-cshi for future inquiries.

PR #553, opened 2022-02-19
SNOW-542628 Add Default Server Side Error

sfc-gh-jbahk:
### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/connection_test.go
+++ b/connection_test.go
@@ -437,3 +437,38 @@ func TestGetInvalidQueryStatus(t *testing.T) {
 		t.Error("expected an error")
 	}
 }
+
+func TestExecWithServerSideError(t *testing.T) {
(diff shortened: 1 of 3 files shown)


connection_test.go
```diff
@@ -437,3 +437,38 @@ func TestGetInvalidQueryStatus(t *testing.T) {
 		t.Error("expected an error")
 	}
 }
+
+func TestExecWithServerSideError(t *testing.T) {
+	postQueryMock := func(_ context.Context, _ *snowflakeRestful,
```
sfc-gh-cshi, line 442: while mock is ok to test when response is empty, do we know what is the possible reason of this response?

PR #563, opened 2022-03-15
SNOW-559970: Add error cast check for SnowflakeError

satishbhor:
checkQueryStatus API not always returns SnowflakeError, e.g. if context is cancelled by client then it may return *errors.errorString,  Hence we should have check whether its  SnowflakeError or not, otherwise it will panic and will crash app.
### Checklist
- [X] Code compiles correctly
- [x] All tests passing


satishbhor:
Hello Team,
Any update on this?

sfc-gh-dszmolka:
#771

sfc-gh-pfus:
Hello @satishbhor ! Thanks for your input, this is very valuable. Can you rebase and solve conflicts?

sfc-gh-pfus:
I noticed, that the only place in which we use casting to `snowflakeError` was already fixed (as @sfc-gh-dszmolka kindly suggested). So I think that we don't need a dedicated function for casting to `snowflakeError`.
In that case I'm going to close this PR, but personally I'm always happy if community suggests improvements. Thank you very much! I hope, that in some time you will provide another improvement or a bug fix!

Issue #555, opened 2022-02-23
SNOW-551325: Panic on missing Location in call to Time.In


jpmunz:
### Issue description

Not sure if related to https://github.com/snowflakedb/gosnowflake/issues/550, in attempting to upgrade from 1.6.6 to 1.6.7 I started seeing a panic:

```
time: missing Location in call to Time.In
```

I believe as a result of this change:
https://github.com/snowflakedb/gosnowflake/compare/v1.6.6...v1.6.7#diff-ecad48054a05ca5aecf42b278530331772747ed7211a193aacd94214cb53aad0R227

sfc-gh-jbahk:
Hi @jpmunz could you share the code that caused this error? Thanks.

kpruden:
I don't have an easy repro, but the panic I'm getting points to this line of code:

https://github.com/snowflakedb/gosnowflake/blob/master/converter.go#L227

jmrobles:
I think is related with https://github.com/snowflakedb/gosnowflake/issues/565

sfc-gh-dszmolka:
should be resolved by #571

Issue #566, opened 2022-03-30
SNOW-567081: Upgrade of the Terraform SF provider from 0.25.36 --> 0.28.6 causes the plugin to crash


melvinRspectrumio:
### Issue description
When we do a `terraform plan` the plugin should not crash.

### Example code
```go
terraform {
  required_providers {
    snowflake = {
      source  = "chanzuckerberg/snowflake"
      version = "0.28.6"
    }
    aws = {
      source  = "hashicorp/aws"
      version = ">= 3.2"
    }
  }
}

terrform plan

```

### Error log
```
Error: Plugin did not respond
│ 
│   with snowflake_database.client_data,
│   on sf_databases.tf line 2, in resource "snowflake_database" "client_data":
│    2: resource "snowflake_database" "client_data" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more details.
╵
╷
│ Error: Request cancelled
│ 
│   with snowflake_database.solutions_data,
│   on sf_databases.tf line 6, in resource "snowflake_database" "solutions_data":
│    6: resource "snowflake_database" "solutions_data" {
│ 
│ The plugin.(*GRPCProvider).UpgradeResourceState request was cancelled.

goroutine 161 [running]:
time.Time.In(...)
	time/time.go:1106
github.com/snowflakedb/gosnowflake.stringToValue(0xc00002c480, {{0xc000227a10, 0xa}, 0x0, 0x0, {0xc000227a20, 0xd}, 0x0, 0x3, 0x1}, ...)
	github.com/snowflakedb/gosnowflake@v1.6.7/converter.go:227 +0xe06
github.com/snowflakedb/gosnowflake.(*snowflakeRows).Next(0xc0009dc600, {0xc00002c480, 0x9, 0xd})
	github.com/snowflakedb/gosnowflake@v1.6.7/rows.go:173 +0x4b6
github.com/luna-duclos/instrumentedsql.wrappedRows.Next({{{0x139b020, 0xc000452f50}, {0x139b080, 0x1bb5578}, 0x0, 0x0}, {0x13be610, 0xc0001ac020}, {0x13aad58, 0xc0009dc600}}, ...)
	github.com/luna-duclos/instrumentedsql@v1.1.3/rows.go:52 +0x1e2
database/sql.(*Rows).nextLocked(0xc000998500)
	database/sql/sql.go:2967 +0x111
[...]

Issue #565, opened 2022-03-23
SNOW-563221: Crashed when locales files not found


jmrobles:
### Issue description

During the session parameters initialization, there is no checking for error in `time.LoadLocation`
![image](https://user-images.githubusercontent.com/1277154/159680242-da221b89-3a0a-4077-9f9a-436809458928.png)

If `gosnowflake`runs in a container, probably the image can not include the timezone/locales files. 

It's necessary to check for the error and not ignore it.

### Example code

N/A

### Error log

Crashed when the `localLocation` is used (convert.go) during the app execution.

`panic: tiime: missing Location in call to Time.In`

### Configuration
*Driver version (or git SHA):* v1.6.7

*Go version:* 1.18

*Server version:* 6.8.1

*Client OS:* Ubuntu 20.04

sfc-gh-dszmolka:
should be resolved with #571

PR #571, opened 2022-04-13
SNOW-563221 Fix Missing Locale

sfc-gh-jbahk:
### Description
Fix crash when location cannot be parsed for whatever reason. Add another safeguard when parsing timestamp_ltz as well.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/arrow_chunk.go
+++ b/arrow_chunk.go
@@ -17,6 +17,7 @@ type arrowResultChunk struct {
 	rowCount         int
 	uncompressedSize int
 	allocator        memory.Allocator
+	params           map[string]*string
 }
 
 func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType, highPrec bool) ([]chunkRowType, error) {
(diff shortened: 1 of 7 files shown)


connection_util.go
```diff
@@ -153,7 +153,11 @@ func (sc *snowflakeConn) populateSessionParameters(parameters []nameValueParamet
 		sc.cfg.Params[strings.ToLower(param.Name)] = &v
 	}
 	if tz, ok := sc.cfg.Params["timezone"]; ok {
```
sfc-gh-cshi, line 155: Could there be multiple connections using different timezones?  And will they update the same localLocation?

sfc-gh-jbahk: I would imagine there is only one timezone per connection, and thus this attribute is specific to a single connection instance.

sfc-gh-cshi: what about this 'localLocation'? this var is updated by a connection parameter.

arrow_chunk.go
```diff
@@ -17,6 +17,7 @@ type arrowResultChunk struct {
 	rowCount         int
 	uncompressedSize int
 	allocator        memory.Allocator
+	params           map[string]*string
```
sfc-gh-cshi, line 20: We could directly hold a lcoation in this new 'params',  to avoid calling the getCurrentLocation function in loops. e.g. in decodeArrowBatch and  decodeArrowChunk.

sfc-gh-cshi (approved):

sfc-gh-dszmolka:
should fix #555  , #565 , #566 .

PR #574, opened 2022-04-20
Remove describe from PrepareContext due to slowdown

swenson:
### Description
We noticed a big slowdown in our queries that we bisected down to the
1.4.1 release.

Specifically, our `CREATE USER` statements went from around 250ms
duration to 1000ms+ duration after this change.

We noticed that in that release, that `snowflakeConn.PrepareContext`
method started making an additional call to describe the query.

When I deleted this extra call, the duration again dropped from 1000ms+
to 250ms.

I could be wrong, but I don't think this call is strictly necessary.

See https://github.com/hashicorp/vault-plugin-database-snowflake/issues/13
for some additional context.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -212,24 +212,6 @@ func (sc *snowflakeConn) PrepareContext(
 	if sc.rest == nil {
 		return nil, driver.ErrBadConn
 	}
-	noResult := isAsyncMode(ctx)
-	data, err := sc.exec(ctx, query, noResult, false, /* isInternal */
-		true /* describeOnly */, []driver.NamedValue{})
-	if err != nil {
-		if data != nil {
-			code, err := strconv.Atoi(data.Code)
-			if err != nil {
-				return nil, err
-			}
-			return nil, (&SnowflakeError{
-				Number:   code,
-				SQLState: data.Data.SQLState,
-				Message:  err.Error(),
-				QueryID:  data.Data.QueryID,
-			}).exceptionTelemetry(sc)
-		}
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


sfc-gh-turbaszek:
@sfc-gh-kdama @sfc-gh-jbahk can one of you take a look? Also CI seems to be not building.

sfc-gh-jbahk:
The CI doesn't run because @swenson doesn't have the appropriate credentials. I created a PR https://github.com/snowflakedb/gosnowflake/pull/580 that tests this one. It fails because PrepareContext is indeed used.

May I ask why you need to call PrepareContext in the first place when creating a user?

cc: @sfc-gh-cshi

swenson:
@sfc-gh-jbahk It's a standard pattern we use -- we had always `Prepare`d before we `Exec`d. When we updated to the latest version of the driver, though, we had a major performance problem that was causing issues for our customers (who might be doing lots of user creations).

I narrowed it down to the change in this PR, which I believe was introduced in #332.

I say it doesn't appear "strictly necessary" because the `describe` check wasn't done before #332, and it appears that the `describe` is only used to check syntax? It seems to be an expensive way to check if the syntax is valid if it takes >4 times longer than the actual query.

sfc-gh-jbahk:
This PR has been merged here: https://github.com/snowflakedb/gosnowflake/pull/580, that modifies an existing test. Closing.

cc: @sfc-gh-cshi @sfc-gh-sbommakanty @sfc-gh-turbaszek

PR #580, opened 2022-05-16
Remove describe from PrepareContext due to slowdown (Test #574)

sfc-gh-jbahk:
We noticed a big slowdown in our queries that we bisected down to the
1.4.1 release.

Specifically, our `CREATE USER` statements went from around 250ms
duration to 1000ms+ duration after this change.

We noticed that in that release, that `snowflakeConn.PrepareContext`
method started making an additional call to describe the query.

When I deleted this extra call, the duration again dropped from 1000ms+
to 250ms.

I could be wrong, but I don't think this call is strictly necessary.

See https://github.com/hashicorp/vault-plugin-database-snowflake/issues/13
for some additional context.

### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -212,24 +212,6 @@ func (sc *snowflakeConn) PrepareContext(
 	if sc.rest == nil {
 		return nil, driver.ErrBadConn
 	}
-	noResult := isAsyncMode(ctx)
-	data, err := sc.exec(ctx, query, noResult, false, /* isInternal */
-		true /* describeOnly */, []driver.NamedValue{})
-	if err != nil {
-		if data != nil {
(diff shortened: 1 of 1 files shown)


sfc-gh-cshi (approved):
Just have a concern if Prepare() is doing nothing, are we implementing right programming model.... Approve for now.

sfc-gh-jbahk:
Agree with @sfc-gh-cshi. Prepare() and PrepareContext() were implemented as placeholders to abide by the Golang SQL API, but should be investigated on whether it has actual usage such as binding.

PR #586, opened 2022-06-07
bind_uploader: don't allow more than 10M in upload buffer

srabraham:
Prior to this, the buffer maximum could be exceeded if a
column's byte count was >8

I'm unable to run the tests myself, so I'm not sure if this
breaks any of them. It does totally seem to resolve the
issue I've been having in trying to upload large batches
though!

### DIFF
--- a/bind_uploader.go
+++ b/bind_uploader.go
@@ -10,7 +10,6 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
-	"unsafe"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


bind_uploader.go
```diff
@@ -41,7 +38,7 @@ func (bu *bindUploader) upload(bindings []driver.NamedValue) (*execResponse, err
 	bu.fileCount = 0
 	var data *execResponse
 	for rowNum < len(bindingRows) {
-		for numBytes < inputStreamBufferSize && rowNum < len(bindingRows) {
+		for rowNum < len(bindingRows) && numBytes + len(bindingRows[rowNum]) < inputStreamBufferSize {
```
srabraham, line 41: note that the conditionals' ordering had to be swapped, since otherwise you can get a slice index out of bounds on the "bindingRows[rowNum]" part

PR #590, opened 2022-06-22
Add test for new statement type CALL

SimbaGithub:
### Description
Simba ticket 00383533
Add test case for new statement type for procedure call.
The server is adding a new statement type for procedure call in the query response send to client, the test case is to confirm if the driver works correctly with that.
The "alter session" query to enable the new statement type could fail if the server doesn't have the new behavior, so ignore that error to make the test case pass on server with old behavior, while the new behavior could be tested when it's rolled out on server side.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-jbahk (approved):

SimbaGithub:
@sfc-gh-hchaturvedi It's now disabled in PR #622. Please review.

PR #620, opened 2022-07-13
Fixes external browser failed with encrypted SAML assertions

SimbaGithub:
### Description
Simba incident 00388511

When the SAML token encryption is enabled in the identity provider, the driver fails authenticating with an external browser because part of the token was trimmed.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-jbahk (approved):

sfc-gh-hchaturvedi:
@SimbaGithub for the future please write a brief description of the fix and how it solves the problem.

PR #621, opened 2022-07-14
security: update apache/arrow to get security patches on gopkg.in/yaml.v3

dominicbarnes:
### Description

This PR updates `github.com/apache/arrow` with the primary goal being to land https://github.com/apache/arrow/pull/13322 that includes https://github.com/stretchr/testify/pull/1192 which ultimately updates `gopkg.in/yaml.v3` with security patches.

To accomplish this, I needed to update imports of apache/arrow to use go modules (which has been adopted by arrow). The closest major version to what existed before is v7, which has no breaking changes for the usage here as far as I can tell. By using go modules like this, it should be easier to manage this dependency, especially if wanting to upgrade to v8 or v9 (current).

### Checklist
- [x] Code compiles correctly
- [ ] ~~Created tests which fail without the change (if possible)~~
- [ ] All tests passing
- [ ] ~~Extended the README / documentation, if necessary~~


dominicbarnes:
It turns out that this PR will fix #608

dominicbarnes:
Looks like the `security/snyk` test fails, but I cannot see the results. If I need to update some other dependencies, let me know and I'll gladly do that as well. :)

dominicbarnes:
I'm still looking to land this security patch, but could use a bit more guidance in order to get it over the finish line.

dominicbarnes:

dominicbarnes:
I've discovered that this PR doesn't actually work, seems like the changes to this dependency aren't as clean as I hoped. I will close this PR and re-open when I can get around to testing it.

PR #627, opened 2022-07-21
Implementation of AWS KMS to handle digital sign of token claims for JWT.

Greyeye:
### Description
Instead of using a provided private key, this change will allow the user to specify the AWS KMS to help generate JWT.
The user must specify the authenticator as "SNOWFLAKE_JWT_KMS" and provide "AWSKMSKeyARN" for this to work.

also added cmd/selectkms1  for a test client.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


sfc-gh-pfus:
Hello @Greyeye ! I'm wondering, why do we need this change? How is that better than standard JWT auth?

sfc-gh-pfus:
Closing as not received response from 2 months and I don't think this is something that we need. Here in snowflake drivers team we try to align drivers code and this one introduces another approach and causes feature gap.

PR #629, opened 2022-07-28
SNOW-590709 Support for using interface slice []interface{} in array binding

SimbaGithub:
### Description
SNOW-590709 (Simba incident 00383422)

The driver currently supports inserting NULL one by one but not in batch (i.e. array binding). Also, NULL cannot be inserted in arrays like []int and []string, so user will need to use the interface slice []interface{}.

This PR has added support for using interface slice []interface{} in array binding so that NULL can be inserted in batch.
Data types supported: INTEGER, FLOAT, STRING, BOOLEAN, BINARY
Not supported: DATE, TIME, TIMESTAMP_LTZ , TIMESTAMP_NTZ , TIMESTAMP_TZ

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -414,7 +414,7 @@ binds arrays to the parameters in the INSERT statement.
 	// Insert the data from the arrays and wrap in an Array() function into the table.
 	_, err = db.Exec("insert into my_table values (?, ?, ?, ?)", Array(&intArray), Array(&fltArray), Array(&boolArray), Array(&strArray))
 
-If the array contains SQL NULL values, use slice []interface{} which allows Golang nil values.
+If the array contains SQL NULL values, use slice []interface{}, which allows Golang nil values.
 This feature is available in version 1.6.12 (and later) of the driver. For exmaple,
 
  	// Define the arrays containing the data to insert.
@@ -442,7 +442,7 @@ This feature is available in version 1.6.12 (and later) of the driver. For exmap
 		}
 	}
 
-Currently, the driver does not support date, time and timestamps types when slice []interface{} is used in array binding.
+Currently, the driver does not support the DATE, TIME, TIMESTAMP_LTZ, TIMESTAMP_NTZ and TIMESTAMP_TZ data types when slice []interface{} is used in array binding.
 
 Note: For alternative ways to load data into the Snowflake database (including bulk loading using the COPY command), see
 Loading Data into Snowflake (https://docs.snowflake.com/en/user-guide-data-load.html).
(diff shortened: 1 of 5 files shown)


doc.go
```diff
@@ -414,6 +414,36 @@ binds arrays to the parameters in the INSERT statement.
 	// Insert the data from the arrays and wrap in an Array() function into the table.
 	_, err = db.Exec("insert into my_table values (?, ?, ?, ?)", Array(&intArray), Array(&fltArray), Array(&boolArray), Array(&strArray))
 
+If the array contains SQL NULL values, use slice []interface{} which allows Golang nil values.
```
sfc-gh-mheavin, line 417: , which allows

doc.go
```diff
@@ -414,6 +414,36 @@ binds arrays to the parameters in the INSERT statement.
...
+			log.Fatalf("Failed to scan. err: %v", err)
+		}
+		if s.Valid {
+			fmt.Println("Retrieved value:", s.String)
+		} else {
+			fmt.Println("Retrieved value: NULL")
+		}
+	}
+
+Currently, the driver does not support date, time and timestamps types when slice []interface{} is used in array binding.
```
sfc-gh-mheavin, line 445: support the date, time, and timestamps data types when...

sfc-gh-mheavin (approved):

doc.go
SimbaGithub: Updated to "...does not support the DATE, TIME, TIMESTAMP_LTZ, TIMESTAMP_NTZ and TIMESTAMP_TZ data types when..."

doc.go
SimbaGithub: Updated

doc.go
```diff
@@ -414,6 +414,36 @@ binds arrays to the parameters in the INSERT statement.
...
+			log.Fatalf("Failed to scan. err: %v", err)
+		}
+		if s.Valid {
+			fmt.Println("Retrieved value:", s.String)
+		} else {
+			fmt.Println("Retrieved value: NULL")
+		}
+	}
+
+Currently, the driver does not support the DATE, TIME, TIMESTAMP_LTZ, TIMESTAMP_NTZ and TIMESTAMP_TZ data types when slice []interface{} is used in array binding.
```
sfc-gh-hchaturvedi, line 445: Should we explicitly say that nil value won't be supported with timestamp? @SimbaGithub

sfc-gh-hchaturvedi (approved):
LGTM

doc.go
SimbaGithub: []interface{} is the only way to insert nil values so the current description should be enough.

PR #645, opened 2022-08-13
SNOW-559914 Fix multipart binding upload failure due to missing padding

SimbaGithub:
### Description
Fix the "Failed to decrypt. Check file key and master key" error that occurs when uploading a file which the size is a multiple of 65536 bytes. Missing padding bytes at the end of the file is the root cause of the issue.

Removed the temporary workaround in #583.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/encrypt_util.go
+++ b/encrypt_util.go
@@ -74,19 +74,26 @@ func encryptStream(
 	chunk := make([]byte, chunkSize)
 
 	// encrypt file with CBC
+	padded := false
 	for {
 		n, err := src.Read(chunk)
 		if n == 0 || err != nil {
 			break
 		} else if n%aes.BlockSize != 0 || n != chunkSize {
 			chunk = padBytesLength(chunk[:n], aes.BlockSize)
+			padded = true
 		}
 		mode.CryptBlocks(cipherText, chunk)
 		out.Write(cipherText[:len(chunk)])
-
 	}
 	if err != nil {
 		return nil, err
+	}
+
+	if !padded {
+		padding := bytes.Repeat([]byte(string(rune(aes.BlockSize))), aes.BlockSize)
+		mode.CryptBlocks(cipherText, padding)
+		out.Write(cipherText[:len(padding)])
 	}
 
 	// encrypt key with ECB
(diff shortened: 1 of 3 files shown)


encrypt_util.go
```diff
@@ -74,21 +74,28 @@ func encryptStream(
 	chunk := make([]byte, chunkSize)
 
 	// encrypt file with CBC
+	padded := false
```
sfc-gh-hchaturvedi, line 77: Shouldn't this be within the `for` loop? You want to make sure that the last chunk you read was padded or not, but once you set the padding to true it would remain true for the entirety of the for loop.

SimbaGithub: The padding needs to be added to the end of the file if it's never padded.
If it has been padded once, no extra padding is required. This logic is taken from the Python driver.
https://github.com/snowflakedb/snowflake-connector-python/blob/6afaa3025874c8e925caa0cbd6a57f2b62049707/src/snowflake/connector/encryption_util.py#L102

sfc-gh-hchaturvedi: Got it

encrypt_util.go
```diff
@@ -74,21 +74,28 @@ func encryptStream(
...
 	for {
 		n, err := src.Read(chunk)
 		if n == 0 || err != nil {
 			break
 		} else if n%aes.BlockSize != 0 || n != chunkSize {
 			chunk = padBytesLength(chunk[:n], aes.BlockSize)
+			padded = true
```
sfc-gh-hchaturvedi, line 84: The second condition in this `else if` block is a little confusing. Shouldn't the `n%aes.BlockSize` be enough?

SimbaGithub: As chunk is allocated with 65536 (chunkSize) bytes, it will be fully filled with 65536 bytes no matter how many bytes src.Read(chunk) has read. The condition `n != chunkSize` in the `else if` block was needed to avoid getting unnecessary bytes from chunk.

To avoid confusion, I have updated the code and added more comments so it should be clear and consistent with the python driver's implementation. Also added some extra test cases.

sfc-gh-hchaturvedi (approved):
LGTM

PR #647, opened 2022-08-19
test: use `T.TempDir` to create temporary test directory

Juneezee:
### Description
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := os.MkdirTemp("", "")
	if err != nil {
		t.Error(err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/put_get_test.go
+++ b/put_get_test.go
@@ -44,7 +44,7 @@ func TestPutError(t *testing.T) {
 	defer f.Close()
 	f.WriteString("test1")
 	os.Chmod(file1, 0000)
-	defer os.Chmod(file1, 0777)
+	defer os.Chmod(file1, 0644)
 
 	data := &execResponseData{
 		Command:           string(uploadCommand),
--- a/put_get_test.go
+++ b/put_get_test.go
@@ -34,19 +34,17 @@ func TestPutError(t *testing.T) {
 	if isWindows {
 		t.Skip("permission model is different")
 	}
-	tmpDir, err := os.MkdirTemp("", "putfiledir")
-	if err != nil {
-		t.Error(err)
-	}
-	defer os.RemoveAll(tmpDir)
+	tmpDir := t.TempDir()
 	file1 := filepath.Join(tmpDir, "file1")
 	remoteLocation := filepath.Join(tmpDir, "remote_loc")
 	f, err := os.Create(file1)
 	if err != nil {
 		t.Error(err)
 	}
+	defer f.Close()
 	f.WriteString("test1")
 	os.Chmod(file1, 0000)
+	defer os.Chmod(file1, 0777)
 
 	data := &execResponseData{
 		Command:           string(uploadCommand),
@@ -253,11 +251,7 @@ func TestPutWithAutoCompressFalse(t *testing.T) {
 	if runningOnGithubAction() && !runningOnAWS() {
 		t.Skip("skipping non aws environment")
 	}
-	tmpDir, err := os.MkdirTemp("", "put")
-	if err != nil {
-		t.Error(err)
-	}
-	defer os.RemoveAll(tmpDir)
+	tmpDir := t.TempDir()
 	testData := filepath.Join(tmpDir, "data.txt")
 	f, err := os.Create(testData)
 	if err != nil {
@@ -294,11 +288,7 @@ func TestPutWithAutoCompressFalse(t *testing.T) {
 }
 
 func TestPutOverwrite(t *testing.T) {
-	tmpDir, err := os.MkdirTemp("", "data")
-	if err != nil {
-		t.Error(err)
-	}
-	defer os.RemoveAll(tmpDir)
+	tmpDir := t.TempDir()
 	testData := filepath.Join(tmpDir, "data.txt")
 	f, err := os.Create(testData)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


Juneezee:

Juneezee:
recheck

Juneezee:
Hello @sfc-gh-dszmolka @sfc-gh-igarish . I apologize for the explicit mention, but I wanted to follow up on this pull request I created some time ago. 

I would be grateful if you could take some time to review the changes and provide me with your feedback. If there are any specific concerns or questions you have about the changes, I am more than willing to help address them.

Thank you for your time and consideration. I truly appreciate it. :heart:

sfc-gh-igarish:
@Juneezee In original code it was using pattern (the second argument) in each MkdirTemp call. How the replacement call works for the same?

Juneezee:
> @Juneezee In original code it was using pattern (the second argument) in each MkdirTemp call. How the replacement call works for the same?

@sfc-gh-igarish `t.TempDir()` uses the test name as the directory name, You can see an example here: https://go.dev/play/p/OtvRbfHywtk

Juneezee:
friendly ping @sfc-gh-igarish

sfc-gh-igarish (approved):
LGTM

Juneezee:
Hi @sfc-gh-igarish, can you re-run those failing tests? I think those failures are not related to this PR. Thanks!

sfc-gh-igarish:
@Juneezee Please update the branch. Even if it's not related to this PR, it's a merge gate. I will try again once you update the branch.

sfc-gh-igarish:
@Juneezee These are the failures, which I believe related to this PR:
 RUN   TestEncryptDecryptFilePadding
    testing.go:1097: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestEncryptDecryptFilePadding3894548605\006\file65544: The process cannot access the file because it is being used by another process.
--- FAIL: TestEncryptDecryptFilePadding (3.80s)
=== RUN   TestEncryptDecryptLargeFile
    testing.go:1097: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestEncryptDecryptLargeFile100552289\001\file0: The process cannot access the file because it is being used by another process.
--- FAIL: TestEncryptDecryptLargeFile (2.09s)

Juneezee:
@sfc-gh-igarish 

Windows handle open files differently so we need to close all open file descriptors explicitly. I've found a few places where we missed adding `.Close()`, and I have added them in the latest commit. Let's try running test again. Thanks!

put_get_test.go
```diff
@@ -34,19 +34,17 @@ func TestPutError(t *testing.T) {
...
 	file1 := filepath.Join(tmpDir, "file1")
 	remoteLocation := filepath.Join(tmpDir, "remote_loc")
 	f, err := os.OpenFile(file1, os.O_CREATE|os.O_WRONLY, os.ModePerm)
 	if err != nil {
 		t.Error(err)
 	}
+	defer f.Close()
 	f.WriteString("test1")
 	os.Chmod(file1, 0000)
+	defer os.Chmod(file1, 0777)
```
sfc-gh-pfus, line 47: Why do we need 777? 644 is not enough?

put_get_test.go
```diff
@@ -386,23 +372,19 @@ func testPutGet(t *testing.T, isStream bool) {
...
 	gzw.Close()
-	if err = os.WriteFile(fname, b.Bytes(), os.ModePerm); err != nil {
+	if err := os.WriteFile(fname, b.Bytes(), os.ModePerm); err != nil {
 		t.Fatal("could not write to gzip file")
 	}
 
 	runTests(t, dsn, func(dbt *DBTest) {
 		dbt.mustExec("create or replace table " + tableName +
 			" (a int, b string)")
+		defer dbt.mustExec("drop table " + tableName)
```
sfc-gh-pfus, line 382: What do you think about replacing defer with this? https://ieftimov.com/posts/testing-in-go-clean-tests-using-t-cleanup/
Of course it'd better to use `drop table if exists` in that manner.

Juneezee: I would prefer `t.Cleanup` in most cases, but here I think `defer` is more appropriate because we want these defer statements to execute as soon as `func(dbt *DBTest)` finishes. If we use `t.Cleanup` here, the table we created will remain until the whole test finishes. 

Although it doesn't matter if we use `defer` or `t.Cleanup` here,  the behaviour is still different

https://github.com/snowflakedb/gosnowflake/blob/ed37dbc38c55554609f596670add06c7d04468de/driver_test.go#L314-L330

sfc-gh-pfus (approved):
LGTM

sfc-gh-pfus:
Thank you very much @Juneezee for your input! I merged your PR :)

PR #453, opened 2021-07-19
SNOW-370765 Arrow High Precision Flag

sfc-gh-jbahk:
### Description
Enable client side flag that enables higher precision for Arrow.

### Checklist
- [x] Code compiles correctly
- [x] Run ``make fmt`` to fix inconsistent formats
- [x] Run ``make lint`` to get lint errors and fix all of them
- [x] Created tests which fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

PR #655, opened 2022-09-09
Fixes floats from being truncated to 7 digit precision

tampajohn:
### Description
#453 introduced a bug with how float64 values are handled.  Since float64 is a native type, there is no reason to utilise the higherPrecision check.  The string formatting that was being utilised was truncating float values past the 7th decimal place.

This PR reverts the change that caused the issue.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -195,7 +195,8 @@ func stringToValue(
 	dest *driver.Value,
 	srcColumnMeta execResponseRowType,
 	srcValue *string,
-	loc *time.Location) error {
+	loc *time.Location,
+) error {
 	if srcValue == nil {
 		logger.Debugf("snowflake data type: %v, raw value: nil", srcColumnMeta.Type)
 		*dest = nil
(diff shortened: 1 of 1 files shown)


converter.go
```diff
@@ -427,11 +427,7 @@ func arrowToValue(
 	case realType:
 		for i, flt64 := range array.NewFloat64Data(data).Float64Values() {
 			if !srcValue.IsNull(i) {
-				if higherPrecision {
```
tampajohn, line 430: This check shouldn't be used for native Go types, only high precision arrow types that require big.* structs.

sfc-gh-jbahk (approved):

tampajohn:
@sfc-gh-jbahk It seems like the tests failed for some unrelated reason?

srabraham:
> @sfc-gh-jbahk It seems like the tests failed for some unrelated reason?

@sfc-gh-jbahk @sfc-gh-igarish could you please have another look and get this merged?

sfc-gh-igarish:
Here what I tried:
1. Get the changes of this PR and created a new PR. This doesn't have CLA Assist issue. But still it has test failures. Look in to test failures and it's related to this change. I discuss with Harsh and Nate. This is going to be a breaking change as currently without high precision context, it expect float in a String. So we suggest to use WithHigherPrecision context during query execution. e.g. "rows := dbt.mustQueryContext(
                        WithHigherPrecision(context.Background()), selectVariousTypes)". This will get you float64 always and it will have high precision value.

tampajohn:
I guess what I don't get is that $

PR #664, opened 2022-09-26
Cache tz -> *time.Location results

tysonmote:
### Description

snowflakeRows.Next calls getCurrentLocation for every row, which means that the driver calls time.LoadLocation (which loads the TZ database) for every single row in the results of a query. Instead, we should only fetch it once, which we can do on the Config object.

In a simple application that runs queries and fetches a large number of rows, getCurrentLocation accounts for 65% of our overall CPU usage:

image

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


tysonmote:

PR #665, opened 2022-09-27
Change GetArrowBatches() in snowflakeRows to be public and blocking call

SimbaGithub:
### Description
Salesforce ticket 00396666

The GetArrowBatches() API was implemented to access data in arrow.Record format directly from queries (#544).
However, the API is not public so it's not usable by customers. It is also not a blocking call which results in panic in async mode.

This PR adds the public interface SnowflakeRows and exposes the functions GetQueryID(), GetStatus() and GetArrowBatches(). This matches the functions that are exposed in SnowflakeResult. GetArrowBatches() is also changed to a blocking call to ensure the arrow batches are ready before fetching it. Customer has verified the changes in this branch.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/chunk_downloader.go
+++ b/chunk_downloader.go
@@ -720,6 +720,11 @@ func (rb *ArrowBatch) Fetch() (*[]array.Record, error) {
 	return rb.rec, nil
 }
 
+// GetRowCount returns the number of rows in an arrow batch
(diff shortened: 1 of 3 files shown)


sfc-gh-igarish (approved):
LGTM

PR #680, opened 2022-10-31
Ignore 405 error when S3 Acceleration is disabled on the S3 bucket

sfc-gh-ext-simba-lb:
### Description
Simba Salesforce ticket number 00406729.

Description provided by Snowflake Support:
When we need to upload a file to an internal stage using the PUT command it fails when GetBucketAccelerateConfiguration returns 405.

It happens when the driver tries to use the S3 acceleration service on the accounts where the S3 Acceleration is disabled on the buckets. 

Observed behavior:
The driver fails with: ERRO[0000]connection.go:318 gosnowflake.(*snowflakeConn).queryContextInternal error: operation error S3: GetBucketAccelerateConfiguration, https response error StatusCode:
The file is not uploaded
In Snowflake job history the PUT query appears successful

Expected behavior:
The 405 error appears in the log
The file is uploaded
In Snowflake job history the PUT query appears successful
(This is how it is implemented in the python connector.)

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -615,6 +615,8 @@ func (sfa *snowflakeFileTransferAgent) transferAccelerateConfig() error {
 			if errors.As(err, &ae) {
 				if ae.ErrorCode() == "AccessDenied" {
 					return nil
+				} else if ae.ErrorCode() == "MethodNotAllowed" {
+					return nil
 				}
 			}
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (approved):
LGTM

Issue #632, opened 2022-07-28
SNOW-636725: panic: crypto/cipher: input not full blocks


Thomasdezeeuw:
### Issue description

See backtrace below.

### Example code

Not available.

### Error log

Trace:
```
panic: crypto/cipher: input not full blocks

goroutine 99 [running]:
crypto/cipher.(*cbcDecrypter).CryptBlocks(0x140003b61a0?, {0x14004800000?, 0x0?, 0x26?}, {0x14004800000?, 0x21?, 0x1400060c978?})
        /opt/homebrew/Cellar/go/1.18.4/libexec/src/crypto/cipher/cbc.go:119 +0x3cc
github.com/snowflakedb/gosnowflake.decryptFile(0x140005a49f0, 0x14000496300?, {0x1400004c6e0, 0x4c}, 0x0, {0x140003b4e00, 0x3b})
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/encrypt_util.go:235 +0x408
github.com/snowflakedb/gosnowflake.(*remoteStorageUtil).downloadOneFile(0x0?, 0x14000496300)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/storage_client.go:213 +0x484
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadOneFile(0x140002a08c0, 0x14000496300)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:986 +0x17c
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadFilesParallel.func1(0x2, 0x140002ddd00?)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:886 +0x7c
created by github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadFilesParallel
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:884 +0x264
```

### Configuration

*Driver version (or git SHA):* gosnowflake@v1.6.11

*Go version:* go version go1.18.4 darwin/arm64

Thomasdezeeuw:
Any update on this? This is blocking our support for Snowflake.

[...]

Issue #631, opened 2022-07-28
SNOW-636720: Panic due to slice bounds out of range in paddingTrim


Thomasdezeeuw:
### Issue description

Panic due to slice bounds out of range, see backtrace below.

### Example code

I can't share at the moment.

### Error log

Trace:

```
panic: runtime error: slice bounds out of range [:-150]

goroutine 114 [running]:
github.com/snowflakedb/gosnowflake.paddingTrim(...)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/encrypt_util.go:277
github.com/snowflakedb/gosnowflake.decryptFile(0x14000311f50, 0x14000314c00?, {0x140005da240, 0x2b}, 0x0, {0x1400031a000, 0x3a})
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/encrypt_util.go:206 +0x518
github.com/snowflakedb/gosnowflake.(*remoteStorageUtil).downloadOneFile(0x0?, 0x14000314c00)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/storage_client.go:213 +0x484
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadOneFile(0x1400051cf00, 0x14000314c00)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:986 +0x17c
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadFilesParallel.func1(0x1, 0x0?)
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:886 +0x7c
created by github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).downloadFilesParallel
        /Users/thomas/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.11/file_transfer_agent.go:884 +0x264
```

### Configuration

*Driver version (or git SHA):* gosnowflake@v1.6.11

*Go version:* go version go1.18.4 darwin/arm64

Thomasdezeeuw:
Any update on this? This is blocking our support for Snowflake.

Thomasdezeeuw:
[...]

Issue #630, opened 2022-07-28
Panic inside `snowflakeFileTransferAgent.parseCommand` due to nil pointer dereference


Thomasdezeeuw:
### Issue description

Panic inside `snowflakeFileTransferAgent.parseCommand` due to nil pointer dereference.

### Example code

The relevant code is here: https://github.com/snowflakedb/gosnowflake/blob/d62ff8279fcb2f72c7bd1b0c85354eb64e4e5f88/file_transfer_agent.go#L257

It uses `fi` and only afterwards check if `err` is nil, this should be flipped: check the error first, then use `fi`.

### Error log

Can't share.

### Configuration

N/A, gosnowflake commit d62ff8279fcb2f72c7bd1b0c85354eb64e4e5f88.

Thomasdezeeuw:
Any update on this? This is blocking our support for Snowflake...

Thomasdezeeuw:
It's been two months now... we're trying to support Snowflake, but without a response this makes it quite difficult.

Thomasdezeeuw:
@sfc-gh-dszmolka I can confirm that this that tree fixes this issue. Thanks @sfc-gh-ext-simba-lb!

sfc-gh-dszmolka:
fixed by #687 earlier

PR #687, opened 2022-11-09
Fix panic issues in file downloading

sfc-gh-ext-simba-lb:
### Description
1. SNOW-636725: panic: crypto/cipher: input not full blocks (Issue #632): Add padding when the source being decrypted is not divisible by block size (16).
3. SNOW-636720: Panic due to slice bounds out of range in paddingTrim (Issue #631): Add check on unpadding to make sure the length is not 0 or too big. Return an error if the padding is invalid.
4. SNOW-636717: Panic inside `snowflakeFileTransferAgent.parseCommand` due to nil pointer dereference (Issue #630): check for error first to avoid nil pointer reference.

The customer has confirmed the fixes on their end.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/encrypt_util.go
+++ b/encrypt_util.go
@@ -215,7 +215,10 @@ func decryptFile(
 	if err = decryptECB(decryptedKey, keyBytes, decodedKey); err != nil {
 		return "", err
 	}
-	decryptedKey = paddingTrim(decryptedKey)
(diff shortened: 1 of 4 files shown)


sfc-gh-igarish (approved):
LGTM

PR #697, opened 2023-01-03
Improve performance for db.ExecContext() with multistatement queries

sfc-gh-ext-simba-lb:
### Description
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/89

db.ExecContext() is used to execute statements that don't return a result set. The current behavior calls getQueryResultResp() for each child query in order to get the update count, even when a query does not return update count. This causes unnecessary requests to the /queries/%s/result endpoint.

Alter behavior to check for DML before calling getQueryResultResp() for each child query, [similar to single statement queries](https://github.com/snowflakedb/gosnowflake/blob/master/connection.go#L261).

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-igarish (commented):
Do we have a test? If not, please add a test.

sfc-gh-ext-simba-lb:
> Do we have a test? If not, please add a test.

There are tests in multistatement_test.go:
- https://github.com/snowflakedb/gosnowflake/blob/master/multistatement_test.go#L10 (queries with no result set but have update count)
- https://github.com/snowflakedb/gosnowflake/blob/master/multistatement_test.go#L148 (mix of result set and update count queries)

I will add an additional test for the case when there is no result or update count.

sfc-gh-igarish:
> 

I am thinking to add a test case which display the timing to make sure the code changes really improve the performance.

sfc-gh-ext-simba-lb:
> > 
> 
> I am suggesting to add a test case which display the timing to make sure the code changes really improve the performance.

Oh I see! Do you have a suggestion how we can prove this in integration test? If we capture the new timing, it won't be compared to the previous version.

sfc-gh-igarish:
> 

Manually you can run test with old driver and then run same test using new driver. This way we need to verify once to prove that improvement happens and note down the percentage of improvement here. In future we need to think about to make it automated.

sfc-gh-ext-simba-lb:
@sfc-gh-igarish Here was the results:
On the old driver the total completion time for the query was 26.5 seconds with the driver, when the query completed within 16.7 seconds on the server side. So there was a driver side delay of ~ 10 seconds.
With the new driver, the query completed in 17.8 seconds with the driver and 17.7 seconds on the server side.

This works out **~32% improvement**:
26.5 - 17.8 = 8.7 
8.7 / 26.5 * 100 = 32%

sfc-gh-igarish (approved):
LGTM

PR #661, opened 2022-09-21
Add connection caching for mfa and id token

SimbaGithub:
### Description
Add a new authenticator USERNAME_PASSWORD_MFA. 
CLIENT_REQUEST_MFA_TOKEN parameter is added.
Support both ID_TOKEN and MFATOKEN in Local Secure Storage.

### Checklist
- [x] Code compiles correctly
- [x] Run ``make fmt`` to fix inconsistent formats
- [x] Run ``make lint`` to get lint errors and fix all of them
- [ ] Created tests which fail without the change (if possible)
- [x] All tests passing
- [ ] Extended the README / documentation, if necessary

PR #705, opened 2023-01-23
Fix connection caching for MFA and external browser authenticators

sfc-gh-ext-simba-lb:
### Description
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/119

Connection caching was added in version 1.6.15 as part of #661 but is not working as expected. 

The PR added CLIENT_REQUEST_MFA_TOKEN/CLIENT_STORE_TEMPORARY_CREDENTIAL properties in the authentication request so the server would send back MFA/ID token. The implementation was sending the session parameter value as string "true" instead of boolean true.

This PR is to change the type of CLIENT_REQUEST_MFA_TOKEN/CLIENT_STORE_TEMPORARY_CREDENTIAL from string to boolean so that it is recognized by the server, and returns the token so it can be cached.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-igarish (approved):
LGTM. As a side notes, if we don't support mfa caching for Linux , as JDBC/ODBC support then how do we track it? Can you create a ticket to track it? we need to support mfa caching on Linux as other drivers although it's not most secure way of doing it.

sfc-gh-ext-simba-lb:
@sfc-gh-igarish the go driver supports writing to local cache file the same as ODBC and JDBC: https://github.com/snowflakedb/gosnowflake/blob/master/secure_storage_manager.go#L239

For Linux it's disabled by default the same as JDBC and ODBC. We use the CLIENT_STORE_TEMPORARY_CREDENTIAL and CLIENT_REQUEST_MFA_TOKEN session parameters to decide whether to cache the token or not. For Windows and Mac OS these properties will be set to true by default if the authenticator is external browser or MFA, but for Linux they are disabled by default. 

i.e JDBC: https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/core/SessionUtil.java#L261

sfc-gh-igarish:
Can we make sure this feature work on all supported platforms? i.e. Windows, Linux, and Mac.

sfc-gh-ext-simba-lb:
@sfc-gh-igarish Added the Linux option to the test and confirmed the feature on Windows, Linux and OSX.

PR #710, opened 2023-01-25
Add connection option to disable connection caching for MFA and external browser

sfc-gh-ext-simba-lb:
### Description
This is to add the ability to disable connection caching for MFA and external browser which are enabled by default. 
If ClientStoreTemporaryCredential = ConfigBoolFalse or ClientRequestMfaToken = ConfigBoolFalse the connection token will not be cached in the credential manager.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing


auth.go
```diff
@@ -478,7 +478,7 @@ func authenticateWithConfig(sc *snowflakeConn) error {
 	//var consentCacheIdToken = true
 
 	if sc.cfg.Authenticator == AuthTypeExternalBrowser {
-		if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		if (runtime.GOOS == "windows" || runtime.GOOS == "darwin") && sc.cfg.ClientStoreTemporaryCredential != ConfigBoolFalse {
```
sfc-gh-igarish, line 481: This logic looks little weird. when flag is already true (i.e. !false) then why to make it true. Please make it clean to set the values.

sfc-gh-igarish (approved):
LGTM

PR #740, opened 2023-03-03
Migrate from azure-storage-blob-go to azure-sdk-for-go

sfc-gh-ext-simba-lb:
### Description
Snyk has reported vulnerabilities that are nested within older transitive dependencies.

The azure sdk used by the driver is managed on a new repository, meaning the one used is outdated. 
`https://github.com/Azure/azure-storage-blob-go` has moved to `https://github.com/Azure/azure-sdk-for-go`

By migrating and upgrading to the new sdk we can remove some older transitive libraries that may contain the nested vulnerabilities:

```
github.com/Azure/azure-pipeline-go v0.2.3
github.com/google/uuid v1.3.0
github.com/mattn/go-ieproxy v0.0.1
gopkg.in/yaml.v2 v2.4.0

```

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


sfc-gh-igarish:
Are we using GCS SDK only for downscope token or even for pre-signed URL? We need to keep using earlier code for pre-signed URL.

sfc-gh-igarish (approved):
LGTM

Issue #701, opened 2023-01-11
SNOW-724764: Silent failure for PUT operation


mihaitodor:
### Issue description

This driver fails to upload files silently when compiled with `CGO_ENABLED=0` and the `USER` environment variable isn't set. This can happen, for example, when running the app in a simple Go container with `--user 1000:1000` (try `docker run --rm -it --user 1000:1000 golang:1.18-alpine sh`). I tried adding `gosnowflake.GetLogger().SetLogLevel("trace")`, but there's no hint in the output that a failure has occurred.

One cause for the issue is [here](https://github.com/snowflakedb/gosnowflake/blob/eaa864cad905d5376097c490a03cb125d38d098b/storage_client.go#L101), where the error is swallowed for some reason... I modified the code and logged the error returned by it and it contains "user: Current requires cgo or $USER set in environment", which helped me find the root cause. Even so, returning non-nil errors there doesn't seem to be enough to trigger a failure log, so more changes might be needed.

Not sure if you considered adding extra linting to this repo, but I believe [`golangci-lint`](https://golangci-lint.run/) should be able to detect missing error checks, among other potential issues and it can be run with ease using GitHub Actions: https://github.com/marketplace/actions/run-golangci-lint

### Example code

The sample code from the docs [here](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Support_For_PUT_and_GET) should suffice.

### Error log

N/A

### Configuration
*Driver version (or git SHA):*
v1.6.16

*Go version:* run `go version` in your console
go version go1.19.4 darwin/amd64

*Server version:* E.g. 1.90.1
You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```
7.0.0

*Client OS:* E.g. Debian 8.1 (Jessie), Windows 10

`golang:1.18-alpine` Docker container.

sfc-gh-dszmolka:
hi and thank you for submitting this issue with us! special thanks for providing a detailed error description 👍 
tried to reproduce the same issue but obviously doing something wrong, because could not get to it.

1. using `golang:1.19-alpine` or `golang:1.18-alpine` base image, did not matter
2. using the [sample1.go](https://github.com/snowflakedb/gosnowflake/blob/master/cmd/select1/select1.go) example script to connect to Snowflake + execute the query; which has been modified to be
```
query := "PUT file:///go/gosnowflake/cmd/select1/select1.go @~ OVERWRITE=true"
```

[...]

PR #770, opened 2023-04-11
SNOW-724764 Log error to avoid silent failure for PUT operation

sfc-gh-ext-simba-lb:
### Description
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/299
#701 

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/storage_client.go
+++ b/storage_client.go
@@ -98,7 +98,10 @@ func (rsu *remoteStorageUtil) uploadOneFile(meta *fileMetadata) error {
 			}
 		}
 		if meta.overwrite || meta.resStatus == notFoundFile {
-			utilClass.uploadFile(dataFile, meta, encryptMeta, maxConcurrency, meta.options.MultiPartThreshold)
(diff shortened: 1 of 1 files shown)


sfc-gh-igarish (approved):
LGTM. Are we taking action in other part of code for handling the error?

sfc-gh-ext-simba-lb:
> LGTM. Are we taking action in other part of code for handling the error?

The error is handled. For parallel uploads the driver doesn't throw the error for a single upload failure. The result and error details are stored for each file: https://github.com/snowflakedb/gosnowflake/blob/master/file_transfer_agent.go#L745

When getting the result for each file there's a check on whether to throw the error or not: https://github.com/snowflakedb/gosnowflake/blob/master/file_transfer_agent.go#L1049

This change is just to log the error so users can know if a file upload failed, but keeping original error handling behavior.

PR #781, opened 2023-04-20
feat: Add QueryArrowStream to snowflakeConn

zeroshade:
### Description
Added a method to `snowflakeConn` called `QueryArrowStream` which returns a loader object which can be used to retrieve `io.ReadCloser`s which are Arrow IPC streams of record batches of results. This allows a consumer to fetch Arrow data directly without having to rely on using the "exact same" version of the arrow library as this driver does.

This was instrumental in building out the Snowflake ADBC Driver as per https://github.com/apache/arrow-adbc/pull/586

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/chunk_downloader.go
+++ b/chunk_downloader.go
@@ -269,14 +269,14 @@ func (scd *snowflakeChunkDownloader) startArrowBatches() error {
 		loc = getCurrentLocation(scd.sc.cfg.Params)
 	}
 	firstArrowChunk := buildFirstArrowChunk(scd.RowSet.RowSetBase64, loc, scd.pool)
-	scd.FirstBatch = &ArrowBatch{
+	firstBatch := &ArrowBatch{
 		idx:                0,
 		scd:                scd,
 		funcDownloadHelper: scd.FuncDownloadHelper,
 	}
 	// decode first chunk if possible
 	if firstArrowChunk.allocator != nil {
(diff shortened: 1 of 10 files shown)


chunk_downloader.go
```diff
@@ -65,7 +65,7 @@ type snowflakeChunkDownloader struct {
 	RowSet             rowSetType
 	FuncDownload       func(context.Context, *snowflakeChunkDownloader, int)
 	FuncDownloadHelper func(context.Context, *snowflakeChunkDownloader, int) error
-	FuncGet            func(context.Context, *snowflakeChunkDownloader, string, map[string]string, time.Duration) (*http.Response, error)
+	FuncGet            func(context.Context, *snowflakeConn, string, map[string]string, time.Duration) (*http.Response, error)
```
sfc-gh-igarish, line 68: Can we add another method? I believe once interface publish not a good idea to change the method signature.

zeroshade: I'm not sure what you mean, this interface isn't a public one, so it shouldn't be an issue to change the signature, right?

sfc-gh-igarish (approved):
LGTM

zeroshade:
@sfc-gh-igarish would you please re-run the tests, I was unable to replicate the test failure myself but I did see one situation where there's a difference in some logic so I think the last change should fix the issue. If not I'll likely need some help debugging this from you guys since I can't replicate this at all on my end hitting my own snowflake test warehouse.

PR #784, opened 2023-04-26
Add tests to increase code coverage

sfc-gh-ext-simba-lb:
### Description
1. Add tests to increase code coverage to above 80%
2. Add code coverage command to build files to run on Github so we can manually check the coverage in the log file. We need to find a way to automate this later.
3. There is an existing mock api for AWS upload. Add the same for Azure and GCS upload/download and AWS download for testing errors.

Current percentages with this PR:
```
aws on ubuntu	    81.4% 
azure on ubuntu	    81.2%
gcp on ubuntu	    81.8%
aws on mac	         80.8%
azure on mac    	 80.5%
gcp on mac     	    81.1%
aws on windows 	    80.3%
azure on windows	    80.1%
gcp on windows 	    80.7%
```

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


storage_client.go
```diff
@@ -226,23 +226,6 @@ func (rsu *remoteStorageUtil) downloadOneFile(meta *fileMetadata) error {
 				meta.dstFileSize = fi.Size()
 			}
 			return nil
-		} else if meta.resStatus == renewPresignedURL {
```
sfc-gh-ext-simba-lb, line 229: This part of the code is unreachable. Since either an error will be returned on line 206 or it will be nil error and status downloaded.

monitoring.go
```diff
@@ -237,20 +237,20 @@ func (sc *snowflakeConn) rowsForRunningQuery(
 	resp, err := sc.getQueryResultResp(ctx, resultPath)
 	if err != nil {
 		logger.WithContext(ctx).Errorf("error: %v", err)
-		if resp != nil {
```
sfc-gh-ext-simba-lb, line 240: sc.getQueryResultResp() will only ever return a resp or an error, never both. Changing the flow to either return the error or check the response.

ci/scripts/test_component.sh
```diff
@@ -12,4 +12,4 @@ if [[ -n "$GITHUB_WORKFLOW" ]]; then
 fi
 env | grep SNOWFLAKE | grep -v PASS | sort
 cd $TOPDIR
-go test -timeout 30m -race $COVFLAGS -v . 
+go test -timeout 30m -race -covermode=atomic $COVFLAGS -v .
```
sfc-gh-igarish, line 15: will this increase time of build? Can we add new atomic flag in the COVFLAGS?

gcs_storage_client.go
```diff
@@ -66,12 +66,18 @@ func (util *snowflakeGcsClient) getFileHeader(meta *fileMetadata, filename strin
 		for k, v := range gcsHeaders {
 			req.Header.Add(k, v)
 		}
-		client := &http.Client{}
+		var client gcsAPI
+		client = &http.Client{}
+		if meta.mockGcsClient != nil {
```
sfc-gh-igarish, line 71: Please put comment like "test only"

gcs_storage_client.go
```diff
@@ -83,8 +89,6 @@ func (util *snowflakeGcsClient) getFileHeader(meta *fileMetadata, filename strin
 				meta.lastError = fmt.Errorf(resp.Status)
 				meta.resStatus = renewToken
 			}
-			meta.lastError = fmt.Errorf(resp.Status)
-			meta.resStatus = errStatus
```
sfc-gh-igarish, line 87: changing position of this lines, could cause some behavior change as values were overwritten before and now it will see retry errors. Is that intentional?

gcs_storage_client.go
```diff
@@ -197,13 +208,16 @@ func (util *snowflakeGcsClient) uploadFile(
 	for k, v := range gcsHeaders {
 		req.Header.Add(k, v)
 	}
-	client := &http.Client{}
+	var client gcsAPI
+	client = &http.Client{}
+	if meta.mockGcsClient != nil {
```
sfc-gh-igarish, line 213: Please add comments for "test only"

gcs_storage_client.go
```diff
@@ -266,24 +282,31 @@ func (util *snowflakeGcsClient) nativeDownloadFile(
...
```
sfc-gh-igarish, line 287: Please add comments

sfc-gh-igarish (approved):
LGTM. This was a bit larger PR. In future please break into smaller PRs.

gcs_storage_client.go
sfc-gh-ext-simba-lb: It's intentional. Previously the error status was always overwritten to errStatus, so for example, if the status code was 404, the status notFoundFile would be overwritten to errStatus at the end of the block. If we set the errStatus first, it won't override other statuses and will remain as the default.

PR #804, opened 2023-05-25
SNOW-823072: Use the context object supplied by BeginTx instead of from the parent connection

### LINKED ISSUE #796: snowflakeTx needs to use the context object supplied by BeginTx(), not the context on the parent connection object
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1.6.17
   
mac OS Ventura 13.3.1 (a) x86_64
   
go version go1.20.4 darwin/amd64

4. Server version: 7.17.0

5. What did you do?

* Establish a connection to snowflake using PingContext(ctx) where the context is set with a short Deadline (say 3 seconds)
* Begin a transaction using BeginTx(ctx, opts) using a separate context object without a deadline
* Run a transaction that could potentially take more than 3 seconds
* Wait for transaction to commit or rollback.

   The transaction executing successfully with a Commit() or failing with a Rollback()
   
   What should have happened and what happened instead?
   
   The transaction never commits or rolls back. Instead, the transaction gets stuck in holding state, and the server returns an error: `000605: Identified SQL statement is not currently executing

7. Can you set logging to DEBUG and collect the logs?

   https://community.snowflake.com/s/article/How-to-generate-log-file-on-Snowflake-connectors
   
8. What is your Snowflake account identifier, if any? (Optional)

---
This issue really stems from the fact that `Commit()` and `Rollback()` are not honoring the context supplied with `BeginTx(ctx)` - they are using a context supplied during PingContext(ctx) that gets stored in the parent `snowflakeConn` object.

So while BeginTx() uses one user-supplied context: https://github.com/snowflakedb/gosnowflake/blob/61ad7b87c8f94489df2ef848a91d8c37f43be00e/connection.go#L167-L194

The returned Transaction object doesn't obtain this context. Instead it uses the context from the parent connection:
https://github.com/snowflakedb/gosnowflake/blob/61ad7b87c8f94489df2ef848a91d8c37f43be00e/transaction.go#L9-L35

This means that `BEGIN` is executed using one context, whereas `COMMIT` and `ROLLBACK` are executed using a different one. If the context supplied in `PingContext(ctx)` has a short deadline, a transaction can get orphaned because it starts, but `COMMIT` and/or `ROLLBACK` fail because the context they're using has past its deadline.

This should be a straightforward change - pass the same context from BeginTx(ctx, opts) through to `snowflakeTx{}` and have all methods there use that context instead.

sfc-gh-ext-simba-lb:
### Description
Fixes #796 
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/378

Commit() and Rollback() are currently not using the same context set in BeginTx(), they are using the context stored in the parent snowflakeConn object. If a connection context has a short deadline, it can result in commit() or rollback() not being called after a transaction due to the deadline exceeded. This is causing locks in customers environment when subsequent sessions hang and fail.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -5,7 +5,10 @@ package gosnowflake
 import (
 	"context"
 	"database/sql"
+	"errors"
+	"fmt"
 	"testing"
+	"time"
 )
 
 func TestTransactionOptions(t *testing.T) {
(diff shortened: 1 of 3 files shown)


sfc-gh-igarish (approved):
LGTM. Just curious why application believes more to txn context and not the connection context? As such txn is associated with a connection.

transaction_test.go
```diff
@@ -35,3 +38,65 @@ func TestTransactionOptions(t *testing.T) {
...
+	defer db.Close()
+
+	ctx1, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	ctx2 := context.Background()
+
+	pingWithRetry := withRetry(PingFunc, 5, 3*time.Second)
+
+	err = pingWithRetry(ctx1, db)
```
rohantandon, line 57: This should use `ctx2`, not `ctx1` since the `withRetry` helper function sets the timeout to 3 seconds automatically.  You can actually get rid of `ctx1`, but given the retry logic, you may end up trying to create a context from a base context that has already timed out.

sfc-gh-ext-simba-lb: I will update this in a separate PR, I missed the edit on the repro. Thanks @rohantandon

PR #807, opened 2023-05-26
SNOW-828186 Remove code duplication in transaction.go

sfc-gh-dheyman-1:
### Description
Common code was extracted to separate function. In addition to that, an enum representation of transaction command was added to disallow any string being passed as a transaction command.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -4,29 +4,47 @@ package gosnowflake
 
 import (
 	"database/sql/driver"
+	"errors"
 )
 
 type snowflakeTx struct {
 	sc *snowflakeConn
 }
 
+type txCommand int
+
+const (
+	commit txCommand = iota
+	rollback
+)
+
+func (cmd txCommand) string() (txStr string, err error) {
+	switch cmd {
+	case commit:
+		return "COMMIT", nil
+	case rollback:
+		return "ROLLBACK", nil
+	}
+	return "", errors.New("unsupported transaction command")
+}
+
 func (tx *snowflakeTx) Commit() (err error) {
-	if tx.sc == nil || tx.sc.rest == nil {
-		return driver.ErrBadConn
-	}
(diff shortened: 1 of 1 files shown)


transaction.go
```diff
@@ -12,23 +12,30 @@ type snowflakeTx struct {
...
 }
 
+type txCommand int
+
+const (
+	commit txCommand = iota
+	rollback
+)
+
+func (cmd txCommand) string() string {
```
sfc-gh-pbulawa, line 22: I think that it would be safer to explicitly bind the enum to the string representation like:
```go
func (cmd txCommand) string() (txStr string, err error) {
	switch cmd {
	case commit:
		return "COMMIT", nil
	case rollback:
		return "ROLLBACK", nil
	}
	return "", errors.New("incorrect tx command")
}
```

sfc-gh-dheyman-1: I double checked and current codebase convention is to use string arrays. Let's discuss this later and create new issue to refactor enums if needed.

sfc-gh-dheyman-1: Added suggestion.

sfc-gh-igarish (approved):
LGTM

transaction.go
```diff
@@ -5,30 +5,48 @@ package gosnowflake
...
+		return "COMMIT", nil
+	case rollback:
+		return "ROLLBACK", nil
 	}
-	tx.sc = nil
-	return
+	return "", errors.New("unsupported transaction command")
+}
+
+func (tx *snowflakeTx) Commit() (err error) {
```
sfc-gh-pfus, line 33: Do we need the return value name? I think we can drop `err` and parenthesis?

sfc-gh-pbulawa (approved):

transaction.go
sfc-gh-dheyman-1: True, applied suggestion.

sfc-gh-pfus (approved):

Issue #795, opened 2023-05-16
Queries that run for over 45 seconds fail with no error 


madisonchamberlain:
I found that running queries which take longer than 45 seconds return with no result and no error. The endpoint I am using when I experience this is `rowsForRunningQuery`. Based on the name of this API, I would expect to be able to submit a query in async mode, and then call this API to wait for it to return. The function, `getQueryResultResp` just waits 45 seconds, but since I am using it through `QueryContext` I would expect this to be blocking. When this happens nothing returns. The response from FuncGet indicates the status of the query, and I am wondering, why not pull until the query is no longer running? (see `ErrQueryIsRunning = 279301 from errors.go:240)

If you are interested in reproducing this is what I am doing:
```
ctx = gosnowflake.WithAsyncMode(gosnowflake.WithAsyncModeNoFetch(ctx))
ctx = gosnowflake.WithFetchResultByID(ctx, snowflakeQueryIdFromOriginalSubmit)

// schedule the execution of the query in async mode
asyncResult, err := conn.QueryContext(ctx, "")
if err == nil && asyncResult != nil {
	if asyncResult.Inner() != nil {
		if snowflakeResult, ok := asyncResult.Inner().(gosnowflake.SnowflakeResult); ok {
			if snowflakeResult == nil {
				do something
				}
			}
		}

		// calling Close() here will block until either the query completes or the ctx expires. If the context expires
		// we rely on the snowflake driver async impl to also cancel the running query
		err = asyncResult.Close()
	}


```

Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of GO driver are you using?
Latest (1.6.20)
   
2. What operating system and processor architecture are you using?
Mac M1
   
3. What version of GO are you using?
run `go version` in your console
1.20

4.Server version:* E.g. 1.90.1
You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```
5. What did you do?

   If possible, provide a recipe for reproducing the error.
   A complete runnable program is good.

6. What did you expect to see?

   What should have happened and what happened instead?

7. Can you set logging to DEBUG and collect the logs?

   https://community.snowflake.com/s/article/How-to-generate-log-file-on-Snowflake-connectors
   
8. What is your Snowflake account identifier, if any? (Optional)

sfc-gh-dszmolka:
[...]

Issue #768, opened 2023-04-07
SNOW-782588: Long running async queries are not handled correctly


pravindra:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of GO driver are you using?

1.6.16
  
2. What operating system and processor architecture are you using?

Mac, arm64
   
3. What version of GO are you using?
run `go version` in your console

go1.19.4 darwin/arm64

4.Server version:* E.g. 1.90.1
You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```

7.10.1

5. What did you do?

   If possible, provide a recipe for reproducing the error.
   A complete runnable program is good.

Ran an async query that takes longer than a minute.

```
ctx, _ := sf.WithMultiStatement(context.Background(), 0)
rows, err := db.QueryContext(sf.WithAsyncMode(ctx), "CALL SYSTEM$WAIT(1, 'MINUTES');use snowflake_sample_data")
if err != nil {
	log.Fatalf("failed to run a query. %v, err: %v", query, err)
}
defer rows.Close()
var v interface{}
for {
	for rows.Next() {
		err := rows.Scan(&v)
		if err != nil {
			log.Fatalf("failed to get result. err: %v", err)
		}
		log.Printf("result: %v", v)
	}
	if !rows.NextResultSet() {
		break
	}
}
```
6. What did you expect to see?

```
result: waited 1 minutes
result: Statement executed successfully.
```

   What should have happened and what happened instead?

did not get any output.

I got the correct behaviour with this patch

```
diff --git a/async.go b/async.go
index 5c477dc..80b704f 100644
--- a/async.go
+++ b/async.go
@@ -63,6 +63,7 @@ func (sr *snowflakeRestful) getAsync(
        defer close(errChannel)
        token, _, _ := sr.TokenAccessor.GetTokens()
        headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, token)
+RETRY:
        resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
        if err != nil {
                logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
@@ -86,6 +87,9 @@ func (sr *snowflakeRestful) getAsync(
 
        sc := &snowflakeConn{rest: sr, cfg: cfg}
        if respd.Success {
+               if respd.Code == queryInProgressCode || respd.Code == queryInProgressAsyncCode {
[...]

PR #824, opened 2023-06-15
SNOW-782588: Retry result request for async query if still in progress

sfc-gh-ext-simba-lb:
### Description
For #768 and #795 
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/323

If a statement is executed and the execution is completed in 45 seconds, the server returns the result. If the statement execution takes longer to complete, the statement handle is returned. 

In the current driver implementation, when making a get request to the `/queries/<query-id>/result` endpoint, the response will be processed even if the server returns the queryInProgressAsyncCode (333334). Since it doesn't check the response code for in progress queries, it will be marked as complete even though the query hasn't finished and the result was not returned. 

This change is to retry the request to the /result endpoint if the query is still in progress so that results are retrieved for queries longer than 45 seconds. Adding a backoff up to 5 seconds similar to the pattern from the jdbc driver: https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/jdbc/SFAsyncResultSet.java#L125

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/async.go
+++ b/async.go
@@ -63,25 +63,46 @@ func (sr *snowflakeRestful) getAsync(
 	defer close(errChannel)
 	token, _, _ := sr.TokenAccessor.GetTokens()
 	headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, token)
-	resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
-	if err != nil {
-		logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
-		sfError.Message = err.Error()
-		errChannel <- sfError
-		return err
-	}
-	if resp.Body != nil {
-		defer resp.Body.Close()
-	}
 
-	respd := execResponse{}
-	err = json.NewDecoder(resp.Body).Decode(&respd)
-	resp.Body.Close()
-	if err != nil {
-		logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
-		sfError.Message = err.Error()
-		errChannel <- sfError
-		return err
+	var err error
+	var respd execResponse
+	retry := 0
+	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
+
+	for {
+		resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
+		if err != nil {
+			logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		defer resp.Body.Close()
+
+		respd = execResponse{} // reset the response
+		err = json.NewDecoder(resp.Body).Decode(&respd)
+		if err != nil {
+			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		if respd.Code != queryInProgressAsyncCode {
+			// If the query takes longer than 45 seconds to complete the results are not returned.
+			// If the query is still in progress after 45 seconds, retry the request to the /results endpoint.
+			// For all other scenarios continue processing results response
+			break
+		} else {
+			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
+			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
+			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
+			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
+			time.Sleep(sleepTime)
+		}
+		if retry < len(retryPattern)-1 {
+			retry++
+		}
+
 	}
 
 	sc := &snowflakeConn{rest: sr, cfg: cfg}
--- a/async_test.go
+++ b/async_test.go
@@ -136,3 +136,38 @@ func retrieveRows(rows *RowsExtended, ch chan string) {
 	ch <- s
 	close(ch)
 }
+
+func TestLongRunningAsyncQuery(t *testing.T) {
+	db := openDB(t)
+	defer db.Close()
+
+	ctx, _ := WithMultiStatement(context.Background(), 0)
+	query := "CALL SYSTEM$WAIT(50, 'SECONDS');use snowflake_sample_data"
+
+	rows, err := db.QueryContext(WithAsyncMode(ctx), query)
+	if err != nil {
+		t.Fatalf("failed to run a query. %v, err: %v", query, err)
+	}
+	defer rows.Close()
+	var v string
+	i := 0
+	for {
+		for rows.Next() {
+			err := rows.Scan(&v)
+			if err != nil {
+				t.Fatalf("failed to get result. err: %v", err)
+			}
+			if v == "" {
+				t.Fatal("should have returned a result")
+			}
+			results := []string{"waited 50 seconds", "Statement executed successfully."}
+			if v != results[i] {
+				t.Fatalf("unexpected result returned. expected: %v, but got: %v", results[i], v)
+			}
+			i++
+		}
+		if !rows.NextResultSet() {
+			break
+		}
+	}
+}


async.go
```diff
@@ -63,25 +63,36 @@ func (sr *snowflakeRestful) getAsync(
...
+		resp.Body.Close()
+		if err != nil {
+			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		if respd.Code != queryInProgressAsyncCode {
+			// If the query takes longer than 45 seconds to complete the results are not returned.
+			// Retry the request when the query is still in progress.
```
sfc-gh-igarish, line 92: Do you like to retry with same query or using query id? Where is the code first query execution and then checking status using the query id?

sfc-gh-igarish, line 92: Please update the comment what you mean by retry?

async.go
```diff
@@ -63,25 +63,36 @@ func (sr *snowflakeRestful) getAsync(
...
-	resp.Body.Close()
-	if err != nil {
-		logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
-		sfError.Message = err.Error()
-		errChannel <- sfError
-		return err
+	var err error
+	var respd execResponse
+
+	for {
```
sfc-gh-pfus, line 70: Are we sure, that we don't need another condition (like based on time?) here? If something goes wrong we can jump into infinite loop and block the thread forever?

async.go
```diff
@@ -63,25 +63,36 @@ func (sr *snowflakeRestful) getAsync(
...
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		if resp.Body != nil {
+			defer resp.Body.Close()
+		}
+		respd = execResponse{} // reset the response
+		err = json.NewDecoder(resp.Body).Decode(&respd)
+		resp.Body.Close()
```
sfc-gh-pfus, line 83: Why do we close it here? Three lines above you defer closing body.

async.go
```diff
@@ -63,25 +63,36 @@ func (sr *snowflakeRestful) getAsync(
...
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		if respd.Code != queryInProgressAsyncCode {
+			// If the query takes longer than 45 seconds to complete the results are not returned.
+			// Retry the request when the query is still in progress.
+			// For all other scenarios continue processing response.
+			break
+		}
```
sfc-gh-pfus, line 95: Don't we need some sleep/wait here? Otherwise we can DoS Snowflake a bit :)

async.go
```diff
@@ -63,25 +63,36 @@ func (sr *snowflakeRestful) getAsync(
...
+		err = json.NewDecoder(resp.Body).Decode(&respd)
+		resp.Body.Close()
+		if err != nil {
+			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
+			sfError.Message = err.Error()
+			errChannel <- sfError
+			return err
+		}
+		if respd.Code != queryInProgressAsyncCode {
+			// If the query takes longer than 45 seconds to complete the results are not returned.
```
sfc-gh-pfus, line 91: Is it the matter of 45 seconds or we can also send a query parameter ensuring async mode?

async_test.go
```diff
@@ -136,3 +136,39 @@ func retrieveRows(rows *RowsExtended, ch chan string) {
 	ch <- s
 	close(ch)
 }
+
+func TestLongRunningAsyncQuery(t *testing.T) {
+	db := openDB(t)
+	defer db.Close()
+
+	ctx, _ := WithMultiStatement(context.Background(), 0)
+	query := "CALL SYSTEM$WAIT(50, 'SECONDS');use snowflake_sample_data"
```
sfc-gh-pfus, line 145: According to my question above - if we can ensure that query is invoked in async manner, maybe we can make this test shorter? Even now tests take like 30 min to run.

async_test.go
```diff
@@ -136,3 +136,39 @@ func retrieveRows(rows *RowsExtended, ch chan string) {
...
+
+	rows, err := db.QueryContext(WithAsyncMode(ctx), query)
+	if err != nil {
+		t.Fatalf("failed to run a query. %v, err: %v", query, err)
+	}
+	defer rows.Close()
+	var v interface{}
+	i := 0
+	for {
```
sfc-gh-pfus, line 154: Can we write it as `for i := 0; rows.NextResultSet(); i++`, what do you think?

async_test.go
```diff
@@ -136,3 +136,39 @@ func retrieveRows(rows *RowsExtended, ch chan string) {
...
+
+	ctx, _ := WithMultiStatement(context.Background(), 0)
+	query := "CALL SYSTEM$WAIT(50, 'SECONDS');use snowflake_sample_data"
+
+	rows, err := db.QueryContext(WithAsyncMode(ctx), query)
+	if err != nil {
+		t.Fatalf("failed to run a query. %v, err: %v", query, err)
+	}
+	defer rows.Close()
+	var v interface{}
```
sfc-gh-pfus, line 152: I suppose this var can be a string?

async.go
sfc-gh-ext-simba-lb: As discussed on the call, the retry is to the `/queries/<query-id>/result `endpoint and not re-executing the original query request. I updated the comment.

async.go
sfc-gh-ext-simba-lb: The only documentation I could find on the async query parameter is for the `/api/v2/statements` endpoint. But we are using `/queries/v1/query-request`. https://docs.snowflake.com/en/developer-guide/sql-api/reference#query-parameters

I'm not sure on the purpose of this parameter either. I understood this as if it's set to true, the statement handle is returned right away vs when false or not specified, a result will be returned if execution completed in 45 seconds.

async.go
sfc-gh-ext-simba-lb: I think this is sufficient. It's similar to how [non async queries in progress](https://github.com/snowflakedb/gosnowflake/blob/master/restful.go#L241) are handled. Can you think of any scenario where it might get stuck? Error codes would be handled or else another response code would be returned and it would break out.

async.go
sfc-gh-ext-simba-lb: Updated.

async_test.go
sfc-gh-ext-simba-lb: This skips the first result set.

async.go
sfc-gh-pfus: Ok, if you are sure that it's ok - I wouldn't argue :)

async.go
sfc-gh-pfus: Ok, I checked that in v1 there is no such parameter as async.

async.go
sfc-gh-pfus: Ok, I didn't know that. In that way, we do not overflow Snowflake, so sleep is not required. Sorry for that!

async_test.go
sfc-gh-pfus: Good catch! I fixed my code as well, if you find it more readable it would be great: https://github.com/snowflakedb/gosnowflake/pull/827/files#diff-9e7181917b4fbf7b4e6ca304941bafec645e1cd26eaef9a6c191f43c7166008aR70

sfc-gh-pfus (approved):

async_test.go
sfc-gh-ext-simba-lb: @sfc-gh-pfus sorry I missed this but I see your PR is already merged.

async.go
```diff
@@ -63,25 +63,45 @@ func (sr *snowflakeRestful) getAsync(
...
-		sfError.Message = err.Error()
-		errChannel <- sfError
-		return err
-	}
-	if resp.Body != nil {
+
+	var err error
+	var respd execResponse
+	retry := 0
+	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
```
sfc-gh-igarish, line 70: why 10 and not 16?

sfc-gh-ext-simba-lb: It's up 5 seconds. Why would it be 16?

`sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])`

async.go
```diff
@@ -63,25 +63,45 @@ func (sr *snowflakeRestful) getAsync(
...
+			// If the query is still in progress after 45 seconds, retry the request to the /results endpoint.
+			// For all other scenarios continue processing results response
+			break
+		} else {
+			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
+			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
+			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
+			time.Sleep(sleepTime)
+		}
+		if retry < len(retryPattern)-1 {
```
sfc-gh-igarish, line 101: There is no else here, does it mean we will keep retrying with last backoff sleeptime? if yes, please put comment in else part like else { // keep retyring with last sleeptime}

async.go
sfc-gh-igarish: 16 is in 2^4. So it's exponential back off.

sfc-gh-ext-simba-lb: Oh but this pattern is 500ms * 10 so 5 seconds

sfc-gh-igarish (approved):
LGTM

PR #834, opened 2023-06-26
SNOW-848019: Add mutex lock to prevent race condition in populateErrorFields()

sfc-gh-ext-simba-lb:
### Description
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/491

An intermittent issue was reported where if a query failed while executing queries simultaneously, the wrong error message could be shown in the log.

The root cause is that during processing of the query response from the server inside the exec() function, there is a function populateErrorFields(). If two or more query responses are returned simultaneously and enter this function around the same time, it can cause a race condition where the error data is updated incorrectly. I am able to reproduce this about ~60% of the time without this change.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -73,6 +73,7 @@ type snowflakeConn struct {
 var (
 	queryIDPattern = `[\w\-_]+`
 	queryIDRegexp  = regexp.MustCompile(queryIDPattern)
+	errMutex = &sync.Mutex{}
 )
 
 func (sc *snowflakeConn) exec(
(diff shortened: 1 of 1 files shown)


connection.go
```diff
@@ -138,7 +139,10 @@ func (sc *snowflakeConn) exec(
 	}
 	logger.WithContext(ctx).Infof("Success: %v, Code: %v", data.Success, code)
 	if !data.Success {
-		return nil, (populateErrorFields(code, data)).exceptionTelemetry(sc)
+		errMutex.Lock()
+		err = (populateErrorFields(code, data)).exceptionTelemetry(sc)
+		errMutex.Unlock()
```
sfc-gh-igarish, line 144: Is it guaranteed even above line throw an error/exception it will come to this line? Basically in any situation, this mutex stay in locked stage without unlock() executed for any reason?

sfc-gh-igarish (approved):
LGTM

connection.go
sfc-gh-pfus: I would rewrite this code as:
```
		errMutex.Lock()
		defer errMutex.Unlock()
		err = (populateErrorFields(code, data)).exceptionTelemetry(sc)
		return nil, err
```
What do you think?

PR #839, opened 2023-06-29
SNOW-832825: Fix async documentation

sfc-gh-pbulawa:
### Description
Fix the async example in documentation.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -825,13 +825,22 @@ and before retrieving the results. For a more elaborative example please see cmd
 			ctx := sf.WithAsyncMode(context.Background())
 
 			// Run the query with asynchronous context
-			rows, _ := db.QueryContext(ctx, "select 1")
+			rows, err := db.QueryContext(ctx, "select 1")
+			if err != nil {
+				// handle error
+			}
 
 			// do something as the workflow continues whereas the query is computing in the background
 			...
 
 			// Get the data when you are ready to handle it
-			cols, err := rows.Columns()
+			var val int
+			err = rows.Scan(&val)
+			if err != nil {
+				// handle error
+			}
+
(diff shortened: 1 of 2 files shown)


doc.go
```diff
@@ -797,28 +797,17 @@ and before retrieving the results.
...
-				} else if status == QueryFailed {
-					// do something
-				}
-				return nil
-			})
+			// Enable asynchronous mode
+			ctx := sf.WithAsyncMode(context.Background())
+
+			// Run the query with asynchronous context
+			rows, _ := db.QueryContext(ctx, "select 1")
```
sfc-gh-dprzybysz, line 804: is it a good practice to ignore the second parameter? is it an error? I understand that if the passed query is wrong then we receive quick answer that sth is wrong?

doc.go
```diff
@@ -797,28 +797,17 @@ and before retrieving the results.
...
+
+			// do something as the workflow continues whereas the query is computing in the background
+			...
+
+			// Get the data when you are ready to handle it
+			cols, err := rows.Columns()
```
sfc-gh-pfus, line 810: Why fetching columns?

sfc-gh-pbulawa: I took it from the `cmd/async.go` example. Replaced it with `.Scan()`.

doc.go
sfc-gh-pbulawa: It is a common practice to skip the errors - in this case I wanted to keep the example, however you are right. This has been addressed.

sfc-gh-igarish (approved):
LGTM

sfc-gh-pfus (approved):

PR #840, opened 2023-06-30
SNOW-853914 Add distributed fetch example

sfc-gh-pfus:
### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/cmd/distributedfetch/distributedfetch.go
+++ b/cmd/distributedfetch/distributedfetch.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"database/sql"
+	"flag"
+	"fmt"
+	"log"
+
+	sf "github.com/snowflakedb/gosnowflake"
+)
+
+func main() {
+	defer printExampleDescription()
+	if !flag.Parsed() {
+		flag.Parse()
+	}
+
+	cfg, err := sf.GetConfigFromEnv([]*sf.ConfigParam{
+		{Name: "Account", EnvName: "SNOWFLAKE_TEST_ACCOUNT", FailOnMissing: true},
+		{Name: "User", EnvName: "SNOWFLAKE_TEST_USER", FailOnMissing: true},
+		{Name: "Password", EnvName: "SNOWFLAKE_TEST_PASSWORD", FailOnMissing: true},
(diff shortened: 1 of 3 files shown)


sfc-gh-igarish (commented):
LGTM. But do we have multi thread like in JDBC to fetch chunks in parallel? Otherwise this will looks like a large resultset fetch.

sfc-gh-pfus:
> LGTM. But do we have multi thread like in JDBC to fetch chunks in parallel? Otherwise this will looks like a large resultset fetch.

@sfc-gh-igarish yes, we have. This is what I meant [here](https://github.com/snowflakedb/gosnowflake/pull/840/files#diff-c521a6b8675547f03ee40c82b33a3db642436d86ac219ab2e863510340b8585dR70). If you observe logs, you can notice that each batch is decoded in separated goroutine.

sfc-gh-dprzybysz (approved):

sfc-gh-pbulawa (approved):

sfc-gh-pmotacki (approved):

cmd/distributedfetch/distributedfetch.go
```diff
@@ -0,0 +1,72 @@
...
+		fmt.Printf("ERROR: %v\n", rows.Err())
+		return
+	}
+
+	sf.GetLogger().SetLogLevel("error")
+}
+
+func printSummary() {
+	fmt.Printf(`
+		Please observe as follows.
```
sfc-gh-dheyman-1, line 68: This sentence is obsolete. Additionally I would inform a reader about what are the configuration possibilities here (batches size etc.). Also i think this is more of a exampleDescription and not a summary.

sfc-gh-dheyman-1 (approved):

sfc-gh-igarish (approved):
LGTM

cmd/distributedfetch/distributedfetch.go
sfc-gh-pfus: 1. Removed first line.
2. Renamed function.
3. I don't think that we should add configuration possibilities. We haven't done that in other examples. Also, this is summary, not configuration instructions :)

sfc-gh-dheyman-1: Agreed, LGTM

PR #845, opened 2023-07-11
SNOW-833537 Each time retry keypair auth with new token

sfc-gh-pfus:
### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -215,7 +215,7 @@ func postAuth(
 	client *http.Client,
 	params *url.Values,
 	headers map[string]string,
-	body []byte,
+	bodyCreator bodyCreatorType,
 	timeout time.Duration) (
 	data *authResponse, err error) {
 	params.Add(requestIDKey, getOrGenerateRequestIDFromContext(ctx).String())
@@ -223,7 +223,7 @@ func postAuth(
 
 	fullURL := sr.getFullURL(loginRequestPath, params)
 	logger.Infof("full URL: %v", fullURL)
-	resp, err := sr.FuncAuthPost(ctx, client, fullURL, headers, body, timeout, true)
+	resp, err := sr.FuncAuthPost(ctx, client, fullURL, headers, bodyCreator, timeout, true)
 	if err != nil {
 		return nil, err
 	}
@@ -287,6 +287,23 @@ func authenticate(
 	samlResponse []byte,
 	proofKey []byte,
 ) (resp *authResponseMain, err error) {
+	if sc.cfg.Authenticator == AuthTypeTokenAccessor {
+		logger.Info("Bypass authentication using existing token from token accessor")
+		sessionInfo := authResponseSessionInfo{
+			DatabaseName:  sc.cfg.Database,
+			SchemaName:    sc.cfg.Schema,
+			WarehouseName: sc.cfg.Warehouse,
+			RoleName:      sc.cfg.Role,
+		}
+		token, masterToken, sessionID := sc.cfg.TokenAccessor.GetTokens()
+		return &authResponseMain{
+			Token:       token,
+			MasterToken: masterToken,
+			SessionID:   sessionID,
+			SessionInfo: sessionInfo,
+		}, nil
+	}
+
 	headers := getHeaders()
 	clientEnvironment := authRequestClientEnvironment{
 		Application: sc.cfg.Application,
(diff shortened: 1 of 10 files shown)


restful.go
```diff
@@ -179,12 +179,12 @@ func postAuthRestful(
...
 	fullURL *url.URL,
 	headers map[string]string,
-	body []byte,
+	bodyCreator func() ([]byte, error),
 	timeout time.Duration,
 	raise4XX bool) (
 	*http.Response, error) {
 	return newRetryHTTP(
-		ctx, client, http.NewRequest, fullURL, headers, timeout).doPost().setBody(body).doRaise4XX(raise4XX).execute()
+		ctx, client, http.NewRequest, fullURL, headers, timeout).doPost().setBodyCreator(bodyCreator).doRaise4XX(raise4XX).execute()
```
sfc-gh-dheyman-1, line 187: you might chain these method calls for better readability.

auth.go
```diff
@@ -310,6 +327,67 @@ func authenticate(
...
+	if sessionParameters[clientStoreTemporaryCredential] == true {
+		token := respd.Data.IDToken
+		setCredential(sc, idToken, token)
+	}
+	return &respd.Data, nil
+}
+
+func createRequestBody(sc *snowflakeConn, sessionParameters map[string]interface{},
+	clientEnvironment authRequestClientEnvironment, proofKey []byte, samlResponse []byte,
+) ([]byte, error) {
```
sfc-gh-dheyman-1, line 390: perhaps we can leverage creating a type for this function like `type bodyCreator func() ([]byte, error)`?

retry.go
```diff
@@ -161,15 +161,23 @@ type clientInterface interface {
...
+	bodyCreator func() ([]byte, error)
+	timeout     time.Duration
+	raise4XX    bool
+}
+
+func (r *retryHTTP) createBody() ([]byte, error) {
+	if r.bodyCreator != nil {
+		return r.bodyCreator()
+	}
+	return r.body, nil
```
sfc-gh-dheyman-1, line 180: can bodyCreator be actually nil? Because from what I can see, either you set it explicitly or set it in `setBody`

retry.go
```diff
@@ -161,15 +161,23 @@ type clientInterface interface {
...
+	method      string
+	fullURL     *url.URL
+	headers     map[string]string
+	body        []byte
+	bodyCreator func() ([]byte, error)
+	timeout     time.Duration
+	raise4XX    bool
+}
+
+func (r *retryHTTP) createBody() ([]byte, error) {
```
sfc-gh-dheyman-1, line 176: I think this shouldn't be responsibility of this struct. I would rather have it as a part of request (maybe instead of `requestFunc` there should be another struct that has this `requestFunc` and body). Although this might be a part of a bigger refactoring.

ci/scripts/hang_webserver.py
```diff
@@ -4,18 +4,35 @@
...
 import threading
 import time
+import json
 
 class HTTPRequestHandler(BaseHTTPRequestHandler):
+    invocations = 0
+
     def do_POST(self):
-        if self.path.startswith('/403'):
+        if self.path.startswith('/reset'):
```
sfc-gh-dheyman-1, line 13: we should consider if we want to keep our `hang_webserver` consistent with other drivers, because these seem to branch out rather hard lately

retry.go
```diff
@@ -202,7 +210,14 @@ func (r *retryHTTP) doPost() *retryHTTP {
 }
 
 func (r *retryHTTP) setBody(body []byte) *retryHTTP {
```
sfc-gh-dheyman-1, line 212: I am not a fan of having both `setBody` and `setBodyCreator` in as a part of creating `newRetryHTTP` API. I think it creates a noise and ambiguity to what actually the body will be.

auth.go
sfc-gh-pfus: Sure thing!

ci/scripts/hang_webserver.py
sfc-gh-pfus: I think for now this is the simplest option to mock errors/timeouts from SF, without any new tools.

retry.go
sfc-gh-pfus: Considering the next comment - this is obsolete and removed.

retry.go
sfc-gh-pfus: You are right. I removed that function at all.

retry.go
sfc-gh-pfus: And still this API is for "client" convenience (now "client" is someone who creates retries). Sometimes client gives as a body creator (auth) and sometimes just body (remaining invocations). I would leave both methods, but if you don't like it we can remove `setBody`.

sfc-gh-dheyman-1: Ok, let's leave them. Perhaps the convenience of it exceeds potential problems.

sfc-gh-dheyman-1 (approved):

sfc-gh-igarish:
@sfc-gh-pfus Please add description.

sfc-gh-igarish (approved):
LGTM

Issue #644, opened 2022-08-12
SNOW-645253: Named bind parameters in exec fails


tayknight:
_go version_
go version go1.19 windows/amd64

_go mod_
go 1.19
github.com/snowflakedb/gosnowflake v1.6.12

_Server version_
6.27.0

### Issue description
When trying to use a named bind variable in an db.Exec statement an error is retuned.

`sf_result, err := snowflake_db.Exec("INSERT INTO DA_DW_DEV.STG.CV_TLMSD_API_TERMS (ID) VALUES (:id)", sql.Named("id", "1"))`

```
002049 (42601): SQL compilation error: error line 1 at position 58
Bind variable :id not set.
```

A "normal" question-mark bind works fine...
`sf_result, err := snowflake_db.Exec("INSERT INTO DA_DW_DEV.STG.CV_TLMSD_API_TERMS (ID) VALUES (?)", "1")`

RowsAffected() returns 1 row.

Thanks for any advice or help.

anekdoti:
I have the same problem with gosnowflake v1.6.13. Executing the statement `db.Exec("CREATE SCHEMA IDENTIFIER(:FOO)", sql.Named("FOO", "MYSCHEMA"))` yields the error

sfc-gh-dszmolka:
thank you for submitting this issue, it is indeed reproducible even with the latest version of the `gosnowflake` driver, `v1.6.19`. we'll take a look to see how this issue can be addressed. In the meantime, the `?` or even the positional `:1` (`:2`, `:3` etc) binding can be used as a workaround.

sfc-gh-dszmolka:
#850

sfc-gh-dszmolka:
PR is merged and will be part of the next upcoming release, expected towards end of July

sfc-gh-dszmolka:
version 1.6.23 of the driver has been released with the fix

PR #850, opened 2023-07-17
SNOW-645253 Handle binding named parameters

sfc-gh-pfus:
### Description
Fixes https://snowflakecomputing.atlassian.net/browse/SNOW-645253 / https://github.com/snowflakedb/gosnowflake/issues/644
Binding does not work for two cases (date/time) - but it doesn't work for any times of binding, not just for named parameters. Need to investigate it further, will provide solution in another PR.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/bindings_test.go
+++ b/bindings_test.go
@@ -301,7 +301,7 @@ func TestBindingInterface(t *testing.T) {
 		if !rows.Next() {
 			dbt.Error("failed to query")
 		}
-		var v1, v2, v3, v4, v5, v6 interface{}
+		var v1, v2, v3, v4, v5, v6 any
 		if err := rows.Scan(&v1, &v2, &v3, &v4, &v5, &v6); err != nil {
 			dbt.Errorf("failed to scan: %#v", err)
 		}
@@ -327,7 +327,7 @@ func TestBindingInterfaceString(t *testing.T) {
 		if !rows.Next() {
 			dbt.Error("failed to query")
 		}
-		var v1, v2, v3, v4, v5, v6 interface{}
+		var v1, v2, v3, v4, v5, v6 any
 		if err := rows.Scan(&v1, &v2, &v3, &v4, &v5, &v6); err != nil {
 			dbt.Errorf("failed to scan: %#v", err)
 		}
@@ -348,7 +348,7 @@ func TestBindingInterfaceString(t *testing.T) {
 }
 
 func TestBulkArrayBindingInterfaceNil(t *testing.T) {
-	nilArray := make([]interface{}, 1)
+	nilArray := make([]any, 1)
 
 	runTests(t, dsn, func(dbt *DBTest) {
 		dbt.mustExec(createTableSQL)
@@ -413,22 +413,22 @@ func TestBulkArrayBindingInterfaceNil(t *testing.T) {
 }
 
 func TestBulkArrayBindingInterface(t *testing.T) {
-	intArray := make([]interface{}, 3)
+	intArray := make([]any, 3)
 	intArray[0] = int32(100)
 	intArray[1] = int32(200)
 
-	fltArray := make([]interface{}, 3)
+	fltArray := make([]any, 3)
 	fltArray[0] = float64(0.1)
 	fltArray[2] = float64(5.678)
 
-	boolArray := make([]interface{}, 3)
+	boolArray := make([]any, 3)
 	boolArray[1] = false
 	boolArray[2] = true
 
-	strArray := make([]interface{}, 3)
+	strArray := make([]any, 3)
 	strArray[2] = "test3"
 
-	byteArray := make([]interface{}, 3)
+	byteArray := make([]any, 3)
 	byteArray[0] = []byte{0x01, 0x02, 0x03}
 	byteArray[2] = []byte{0x07, 0x08, 0x09}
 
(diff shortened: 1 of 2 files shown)


bindings_test.go
```diff
@@ -930,3 +930,137 @@ func TestFunctionParameters(t *testing.T) {
...
+			transform: func(v interface{}) interface{} { return v },
+		},
+		{
+			desc:      ":1",
+			insert:    "INSERT INTO BINDING_MODES VALUES(:1)",
+			query:     "SELECT * FROM BINDING_MODES WHERE param1 = :1",
+			transform: func(v interface{}) interface{} { return v },
+		},
+		{
+			desc:      ":id",
```
sfc-gh-dprzybysz, line 972: should be `:param`?

bindings_test.go
```diff
@@ -930,3 +930,137 @@ func TestFunctionParameters(t *testing.T) {
...
+			query:     "SELECT * FROM BINDING_MODES WHERE param1 = :param",
+			transform: func(v interface{}) interface{} { return sql.Named("param", v) },
+		},
+	}
+
+	runTests(t, dsn, func(dbt *DBTest) {
+		for _, tc := range testcases {
+			for _, bindingMode := range bindingModes {
+				t.Run(tc.testDesc+" "+bindingMode.desc, func(t *testing.T) {
+					query := fmt.Sprintf(`CREATE OR REPLACE TABLE BINDING_MODES(param1 %v)`, tc.paramType)
```
sfc-gh-dprzybysz, line 983: who drops that table after tests?

bindings_test.go
```diff
@@ -930,3 +930,137 @@ func TestFunctionParameters(t *testing.T) {
...
+					if !rows.Next() {
+						t.Fatal("Expected to return a row")
+					}
+				})
+			}
+		}
+	})
+}
+
+func TestVariousBindingModesForNullValues(t *testing.T) {
```
sfc-gh-dprzybysz, line 1001: what's the difference between these suite and the one above?

bindings_test.go
```diff
@@ -930,3 +930,137 @@ func TestFunctionParameters(t *testing.T) {
...
+			query:     "SELECT * FROM BINDING_MODES WHERE param1 IS NULL",
+			transform: func(v interface{}) interface{} { return sql.Named("param", v) },
+		},
+	}
+
+	runTests(t, dsn, func(dbt *DBTest) {
+		for _, tc := range testcases {
+			for _, bindingMode := range bindingModes {
+				t.Run(tc.testDesc+" "+bindingMode.desc, func(t *testing.T) {
+					query := fmt.Sprintf(`CREATE OR REPLACE TABLE BINDING_MODES(param1 %v)`, tc.paramType)
```
sfc-gh-dprzybysz, line 1050: who drops that table?

bindings_test.go
```diff
@@ -930,3 +930,137 @@ func TestFunctionParameters(t *testing.T) {
...
+			transform: func(v interface{}) interface{} { return v },
+		},
+		{
+			desc:      ":1",
+			insert:    "INSERT INTO BINDING_MODES VALUES(:1)",
+			query:     "SELECT * FROM BINDING_MODES WHERE param1 IS NULL",
+			transform: func(v interface{}) interface{} { return v },
+		},
+		{
+			desc:      ":id",
```
sfc-gh-dprzybysz, line 1039: should be `:param`?

bindings_test.go
sfc-gh-pfus: Yes it should.

bindings_test.go
sfc-gh-pfus: The whole schema is dropped when tests are finished.

bindings_test.go
sfc-gh-pfus: There are nulls tested. Nulls are handled in a different manner in Go which supports nulls only for pointers. Also, the query is also different.

bindings_test.go
sfc-gh-pfus: Yes, it should.

bindings_test.go
sfc-gh-pfus: Same as above.

bindings_test.go
sfc-gh-dprzybysz: In the new test there is a lot of duplication - maybe you can extract one parameter more and then put those tests in one suite?

bindings_test.go
sfc-gh-pfus: No, in developer machine it's the same. See [teardown](https://github.com/snowflakedb/gosnowflake/blob/master/driver_test.go#L134) invoked after [all tests](https://github.com/snowflakedb/gosnowflake/blob/master/driver_test.go#L158).

bindings_test.go
sfc-gh-pfus: I tried - but it's less readable. You have to change query between `= ?`/`= :1`/`= :1` and `is null`. IMHO it wasn't worth it.

sfc-gh-dheyman-1: I think I would consolidate these. Seems like theres no point in maintaining the same test running logic twice. Also It seems like we would only add some scenarios' config so readability wouldn't suffer that much

sfc-gh-dprzybysz (approved):

sfc-gh-pbulawa (approved):

sfc-gh-dheyman-1 (approved):

sfc-gh-igarish (approved):
LGTM

PR #851, opened 2023-07-17
fix(arrow): handle non-arrow result sets

zeroshade:
### Description
Some queries will not return Arrow data even when requested with format == arrow, such as "SHOW WAREHOUSES". This ended up causing an [issue](https://github.com/apache/arrow-adbc/issues/863) in ADBC when it happened. In order to address this scenario, we need to expose the raw json data through the ArrowStreamLoader in addition to the batches. 

### Checklist
- [x] Code compiles correctly


sfc-gh-igarish (approved):
LGTM

PR #856, opened 2023-07-19
SNOW-859548 Reuse connection in tests

sfc-gh-pfus:
### Description
All tests opened new connections, even if they could reuse a connection from the pool. Migrating to the pool decreased tests time around 15%.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/arrow_test.go
+++ b/arrow_test.go
@@ -14,8 +14,9 @@ import (
 )
 
 func TestArrowBigInt(t *testing.T) {
-	db := openDB(t)
-	dbt := &DBTest{t, db}
+	conn := openConn(t)
+	defer conn.Close()
+	dbt := &DBTest{t, conn}
 
 	testcases := []struct {
 		num  string
@@ -24,7 +25,7 @@ func TestArrowBigInt(t *testing.T) {
 	}{
 		{"10000000000000000000000000000000000000", 38, 0},
 		{"-10000000000000000000000000000000000000", 38, 0},
-		{"12345678901234567890123456789012345678", 38, 0},
+		{"12345678901234567890123456789012345678", 38, 0}, // #pragma: allowlist secret
 		{"-12345678901234567890123456789012345678", 38, 0},
 		{"99999999999999999999999999999999999999", 38, 0},
 		{"-99999999999999999999999999999999999999", 38, 0},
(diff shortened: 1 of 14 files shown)


async_test.go
```diff
@@ -124,7 +133,7 @@ func TestMultipleAsyncQueries(t *testing.T) {
 	})
 }
 
-func retrieveRows(rows *RowsExtended, ch chan string) {
```
sfc-gh-pbulawa, line 127: What is the reason to replace `RowsExtended` with `sql.Rows`?

arrow_test.go
```diff
@@ -95,7 +97,12 @@ func TestArrowBigFloat(t *testing.T) {
 
 func TestArrowIntPrecision(t *testing.T) {
 	db := openDB(t)
-	dbt := &DBTest{t, db}
```
sfc-gh-pbulawa, line 98: What is the reason we are no longer using `DBTest` in some tests? I think that it would be nice to unify the approach and select only one of the options - `DBTest` or `sql.DB`.

sfc-gh-pfus: I wanted to remove `db` from `DBTest` and provide `conn` only. I'd say that using db is only for specific use cases, when default connection cannot be used.

async_test.go
sfc-gh-pfus: Because I changed function providing rows - from `dbt.mustQueryContext` to `db.QueryContext`. It returns `Rows` instead of `RowsExtended`. But we don't use features of `RowsExtended` so this is fine IMHO.

sfc-gh-pbulawa (approved):

arrow_test.go
```diff
@@ -53,8 +54,9 @@ func TestArrowBigInt(t *testing.T) {
 }
 
 func TestArrowBigFloat(t *testing.T) {
-	db := openDB(t)
-	dbt := &DBTest{t, db}
+	conn := openConn(t)
```
sfc-gh-igarish, line 57: Do we guarantee that this test run always first to pool the connection?

sfc-gh-igarish:
make sure one test failure won't affect run of other tests. e.g. first test getting connection failed, other tests should try to get the connection and run as independent test.

sfc-gh-igarish (approved):
LGTM

arrow_test.go
sfc-gh-pfus: No, but it doesn't matter. Pool is managed by Go SQL and it works very well. I tried to run single test, multiple, all, and even tests in which connections were hang (using burp suite), and it worked as expected.

PR #858, opened 2023-07-20
SNOW-859548 Replace test table in each test

sfc-gh-pfus:
### Description
Each test did a round trip or even two to drop tables, which were not created in most tests. Additionally, all test tables are dropped on the end of the suite, when the whole schema is dropped.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/bindings_test.go
+++ b/bindings_test.go
@@ -43,7 +43,7 @@ func TestBindingFloat64(t *testing.T) {
 		var out float64
 		var rows *RowsExtended
 		for _, v := range types {
-			dbt.mustExec(fmt.Sprintf("CREATE TABLE test (id int, value %v)", v))
+			dbt.mustExec(fmt.Sprintf("CREATE OR REPLACE TABLE test (id int, value %v)", v))
 			dbt.mustExec("INSERT INTO test VALUES (1, ?)", expected)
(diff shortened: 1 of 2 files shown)


sfc-gh-pbulawa (approved):

bindings_test.go
```diff
@@ -43,7 +43,7 @@ func TestBindingFloat64(t *testing.T) {
 		var out float64
 		var rows *RowsExtended
 		for _, v := range types {
-			dbt.mustExec(fmt.Sprintf("CREATE TABLE test (id int, value %v)", v))
+			dbt.mustExec(fmt.Sprintf("CREATE OR REPLACE TABLE test (id int, value %v)", v))
```
sfc-gh-igarish, line 46: If table exist will some rows inserted already, will it break the test, which expect only certain rows in the table?

sfc-gh-pfus: No :) `CREATE OR REPLACE` is very similar to `DROP` followed by `CREATE`. The only difference is the time travel feature, which is irrelevant in our tests. See red box [in docs](https://docs.snowflake.com/en/sql-reference/sql/create-table#usage-notes)

sfc-gh-pmotacki (approved):

sfc-gh-igarish (approved):
LGTM

PR #871, opened 2023-08-01
SNOW-880442: Add pass context to arrow fetch function

sfc-gh-pfus:
### Description
Pass context to Arrow `Fetch` function to enable specific contexts like cancellations etc.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/chunk_downloader.go
+++ b/chunk_downloader.go
@@ -707,6 +707,13 @@ type ArrowBatch struct {
 	rowCount           int
 	scd                *snowflakeChunkDownloader
 	funcDownloadHelper func(context.Context, *snowflakeChunkDownloader, int) error
+	ctx                context.Context
+}
+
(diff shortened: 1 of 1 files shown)


chunk_downloader.go
```diff
@@ -707,6 +707,12 @@ type ArrowBatch struct {
 	rowCount           int
 	scd                *snowflakeChunkDownloader
 	funcDownloadHelper func(context.Context, *snowflakeChunkDownloader, int) error
+	ctx                context.Context
```
sfc-gh-jtang, line 710: if it's an internal interface, have you updated all the call sites? If it's a public interface and we won't be able to enumerate all the call sites, we will need to make the extra parameter optional with a default value, right?

sfc-gh-dprzybysz (approved):

Issue #832, opened 2023-06-23
Add ability to specify / disable temp directory


mihaitodor:
## What is the current behavior?

Currently, users can specify the location where temp files are created via the Go-specific env var `TMPDIR`. The driver needs to write temp files as part of its workflow.

## What is the desired behavior?

We’d like the ability to specify the temp directory used by this driver via a dedicated mechanism, such as the `SNOWFLAKE_TMPDIR` env var instead of relying on Go’s general purpose `TMPDIR` env var. Additionally, it would be great to be able to disable it completely, even if that means disabling some functionality.

## How would this improve `gosnowflake`?

This would allow people to use the driver when the local file system is readonly.

## References, Other Background

This is a follow-up to #700.

sfc-gh-dszmolka:
thank you so much @mihaitodor for creating this follow-up issue; now I can update this properly once there's any new progress on the improvement part

sfc-gh-dszmolka:
#874 aims to implement this functionality.

sfc-gh-dszmolka:
PR has been merged and will be part of the next release, which is expected towards the end of August

mihaitodor:
@sfc-gh-dszmolka Awesome, thank you! Looking at the changes, I'm not sure if this also adds the ability to disable the usage of temp files. Is this something that you're planning to implement in the future?

Also, I think it might be worth documenting this tempdir functionality in https://github.com/snowflakedb/gosnowflake/blob/master/doc.go so people can learn about it via the package docs: https://pkg.go.dev/github.com/snowflakedb/gosnowflake

sfc-gh-dszmolka:
got some update on this one:
1. docs: #884 for the newly implemented feature (ability to specify the temp dir)
2. the driver team has considered the other request too (disabling temp dir usage completely, even at the cost of loss of driver functionality) and for now we decided not to move forward with this request, as it seems to be a non-generic use-case of the gosnowflake connector. Of course this can be re-evaluated any time if more users and customers are asking for it, we will reconsider it then.

Speaking of which, as with any feature request with any Snowflake connector: if you (or the future readers, who would need the ability of disabling the usage of temp directory completely) are already a Snowflake customer, please do speak to your Sales representative and let them know how important this feature would be to you. They have access to certain other aspects of your account, which often can help the product teams to prioritize the requests differently.

[...]

Issue #700, opened 2023-01-10
SNOW-723750: Panic with readonly file system


mihaitodor:
### Issue description

The `gosnowflake` driver panics for PUT operations when run in a container with a readonly file system. The panic happens [here](https://github.com/snowflakedb/gosnowflake/blob/0450f0b16a4679b216baecd3fd6cdce739dbb683/file_transfer_agent.go#L750) and the issue comes from [here](https://github.com/snowflakedb/gosnowflake/blob/0450f0b16a4679b216baecd3fd6cdce739dbb683/file_transfer_agent.go#L833). I see a similar problem for download operations [here](https://github.com/snowflakedb/gosnowflake/blob/0450f0b16a4679b216baecd3fd6cdce739dbb683/file_transfer_agent.go#L980).

Why does the driver require a disk buffer? I don't see any mention of this in the docs and there isn't any Snowflake-specific configuration variable for the disk path.

I can work around it by setting the `TMPDIR` environment variable as described in the Go docs [here](https://pkg.go.dev/os#TempDir), but not being able to run this with a fully read-only file system can cause security challenges.

### Example code
The sample code from the docs [here](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Support_For_PUT_and_GET) should suffice.

### Error log
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe0 pc=0x292c38c]

goroutine 74 [running]:
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).uploadFilesParallel(0xc001276b40, {0xc000c845d0, 0x1, 0x1})
	/Users/ntodor/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.16/file_transfer_agent.go:750 +0x40c
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).upload(0xc001276b40, {0xc001a2e2d0?, 0x0, 0x0}, {0xc000c845d0?, 0x1, 0x1})
	/Users/ntodor/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.6.16/file_transfer_agent.go:686 +0x191
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).execute(0xc001276b40)
[...]

PR #874, opened 2023-08-03
SNOW-845282: Allow configuring tmpdir in DSN

sfc-gh-pfus:
### Description
Introduces configuring temp dir used for encryption and compression. The change was needed as in some container environments temporary directory is readonly.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -91,10 +91,23 @@ type Config struct {
 
 	Tracing string // sets logging level
 
+	TmpDirPath string // sets temporary directory used by a driver for operations like encrypting, compressing etc
+
 	MfaToken                       string     // Internally used to cache the MFA token
 	IDToken                        string     // Internally used to cache the Id Token for external browser
 	ClientRequestMfaToken          ConfigBool // When true the MFA token is cached in the credential manager. True by default in Windows/OSX. False for Linux.
 	ClientStoreTemporaryCredential ConfigBool // When true the ID token is cached in the credential manager. True by default in Windows/OSX. False for Linux.
+}
+
+// Validate enables testing if config is correct.
+// A driver client may call it manually, but it is also called during opening first connection.
+func (c *Config) Validate() error {
+	if c.TmpDirPath != "" {
+		if _, err := os.Stat(c.TmpDirPath); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // ocspMode returns the OCSP mode in string INSECURE, FAIL_OPEN, FAIL_CLOSED
@@ -213,6 +226,9 @@ func DSN(cfg *Config) (dsn string, err error) {
 	}
 	if cfg.Tracing != "" {
 		params.Add("tracing", cfg.Tracing)
+	}
+	if cfg.TmpDirPath != "" {
+		params.Add("tmpDirPath", cfg.TmpDirPath)
 	}
 
 	params.Add("ocspFailOpen", strconv.FormatBool(cfg.OCSPFailOpen != OCSPFailOpenFalse))
@@ -684,6 +700,8 @@ func parseDSNParams(cfg *Config, params string) (err error) {
 			}
 		case "tracing":
 			cfg.Tracing = value
+		case "tmpDirPath":
+			cfg.TmpDirPath = value
 		default:
 			if cfg.Params == nil {
 				cfg.Params = make(map[string]*string)
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -580,6 +580,20 @@ func TestParseDSN(t *testing.T) {
 			ocspMode: ocspModeFailOpen,
 			err:      nil,
 		},
+		{
+			dsn: "u:p@a.r.c.snowflakecomputing.com/db/s?account=a.r.c&tmpDirPath=%2Ftmp",
+			config: &Config{
+				Account: "a", User: "u", Password: "p",
+				Protocol: "https", Host: "a.r.c.snowflakecomputing.com", Port: 443,
+				Database: "db", Schema: "s", ValidateDefaultParameters: ConfigBoolTrue, OCSPFailOpen: OCSPFailOpenTrue,
+				ClientTimeout:          defaultClientTimeout,
+				JWTClientTimeout:       defaultJWTClientTimeout,
+				ExternalBrowserTimeout: defaultExternalBrowserTimeout,
+				TmpDirPath:             "/tmp",
+			},
+			ocspMode: ocspModeFailOpen,
+			err:      nil,
+		},
 	}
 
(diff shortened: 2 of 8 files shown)


file_transfer_agent_test.go
```diff
@@ -597,3 +597,126 @@ func TestUploadWhenFilesystemReadOnlyError(t *testing.T) {
...
+	}
+	_, err = sfa.uploadOneFile(uploadMeta)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = sfa.downloadOneFile(downloadMeta)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove("download.txt")
```
sfc-gh-dprzybysz, line 668: shouldn't we cleanup after creating this file?

file_transfer_agent_test.go
```diff
@@ -597,3 +597,126 @@ func TestUploadWhenFilesystemReadOnlyError(t *testing.T) {
...
+				TmpDirPath: tmpDir,
+			},
+		},
+		stageLocationType: local,
+	}
+	_, err = sfa.uploadOneFile(uploadMeta)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = sfa.downloadOneFile(downloadMeta)
```
sfc-gh-dprzybysz, line 664: shouldn't we be interested in results? not only in errors?

file_transfer_agent_test.go
```diff
@@ -597,3 +597,126 @@ func TestUploadWhenFilesystemReadOnlyError(t *testing.T) {
...
+		t.Error(err)
+	}
+	f.WriteString("test1,test2\ntest3,test4\n")
+	f.Close()
+
+	err = os.Chmod(tmpDir, 0400)
+	if err != nil {
+		t.Fatalf("cannot mark directory as readonly: %v", err)
+	}
+	defer os.Chmod(tmpDir, 0600)
```
sfc-gh-dprzybysz, line 690: maybe it is not important here but shuldn't you before `chmod 0400` get initial rights and here set exactly the same?

dsn_test.go
```diff
@@ -580,6 +580,20 @@ func TestParseDSN(t *testing.T) {
...
+		{
+			dsn: "u:p@a.r.c.snowflakecomputing.com/db/s?account=a.r.c&tmpDirPath=%2Ftmp",
+			config: &Config{
+				Account: "a", User: "u", Password: "p",
+				Protocol: "https", Host: "a.r.c.snowflakecomputing.com", Port: 443,
+				Database: "db", Schema: "s", ValidateDefaultParameters: ConfigBoolTrue, OCSPFailOpen: OCSPFailOpenTrue,
+				ClientTimeout:          defaultClientTimeout,
+				JWTClientTimeout:       defaultJWTClientTimeout,
+				ExternalBrowserTimeout: defaultExternalBrowserTimeout,
+				TmpDirPath:             "/tmp",
```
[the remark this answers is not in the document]
sfc-gh-pfus: No, it doesn't matter, this is only a config entry, any arbitrary string here suffices.

file_transfer_agent_test.go
sfc-gh-pfus: This is exactly what this line does :)

file_transfer_agent_test.go
sfc-gh-pfus: No, it doesn't matter. The purpose of this test is to test if custom path does not cause any error.

file_transfer_agent_test.go
sfc-gh-dprzybysz: you mean that `download.txt` was created just before the if statement above? if yes then close the comment

file_transfer_agent_test.go
sfc-gh-pfus: No, this is the directory I created for tests only and I have to give it 600 to be able to remove it afterwards.

file_transfer_agent_test.go
sfc-gh-pfus: Yes, the `downloadOneFile` creates this file, this is the actual cleanup.

file_transfer_agent_test.go
```diff
@@ -597,3 +597,126 @@ func TestUploadWhenFilesystemReadOnlyError(t *testing.T) {
 		t.Fatalf("should error when creating the temporary directory. Instead errored with: %v", err)
 	}
 }
+
+func TestCustomTmpDirPath(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("/tmp", "")
```
sfc-gh-dprzybysz, line 602: should it be `/tmp` also for Windows?

sfc-gh-pfus: No, of course not.

sfc-gh-dprzybysz (approved):

dsn.go
```diff
@@ -684,6 +689,8 @@ func parseDSNParams(cfg *Config, params string) (err error) {
 			}
 		case "tracing":
 			cfg.Tracing = value
+		case "tmpDirPath":
+			cfg.TmpDirPath = value
```
sfc-gh-igarish, line 693: do we need to validate the value and throw an error if it's not valid such as empty, null, or doesn't exist? Also why we need this fetaure and not use the OS specific environment variable exist such as TMP or TMPDIR etc?

mihaitodor: Hey @sfc-gh-igarish, there are some details in #832 and #700, but having an OS-specific path causes a lot of confusion for end users who leverage this driver as part of more generic tooling. Where I work, Kubernetes clusters tend to be quite locked down and the default is to have a read-only file system, which caused some friction for people who tried to run this driver as part of a [Benthos](https://benthos.dev) pipeline. The first issue is that the package docs don't mention anything about this driver needing access to a tempdir to write data, so users found out the hard way by getting a cryptic panic (now fixed by emitting a more informative error). Then it's difficult to document in a clear way and it's not possible to expose it as something like `SNOWFLAKE_TMPDIR`, so users know exactly which temp files are produced by this driver, assuming the larger codebase which includes it might also need to write some temp files for other purposes. Finally, while I haven't studied the code in detail, I think it would be very useful for this driver to document explicitly in which situations it needs to write temp files and to provide a mechanism to supress them for simple workflows where the user just needs to stream some in-memory data to / from a stage file or to execute some simple SQL query. Hope that helps!

sfc-gh-igarish: @sfc-gh-pfus ok. please merge it. But still I don't understand why someone make TMP_DIR as read only. Temp directory meant for outputting any temporary files. Also advantage of default TEMP_DIR, for any reason file left over in it then on restart it can be clean easily and normally no one hesitate to clean the TMP_DIR as they know default usage of it when disk become full.

PR #875, opened 2023-08-03
SNOW-540086 Code coverage: Refactor unused function downloadFilesSequentially()

sfc-gh-ext-simba-lb:
### Description
When executing a command the fileTransferAgent checks the srcFileSize and assigns it to either large or small file metadata. Large file metadata is uploaded/downloaded sequentially while small file metadata is uploaded/downloaded in parallel.

However, the srcFileSize for downloading is always 0, since we don't yet know the file size in the stage. So when it hits [this block](https://github.com/snowflakedb/gosnowflake/blob/master/file_transfer_agent.go#L169), it will always be added to small file metadata and then downloaded in parallel.

This PR is to remove the unused function.

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -176,7 +176,7 @@ func (sfa *snowflakeFileTransferAgent) execute() error {
 		if sfa.stageLocationType != local {
 			sizeThreshold := sfa.options.MultiPartThreshold
 			meta.options.MultiPartThreshold = sizeThreshold
-			if meta.srcFileSize > sizeThreshold {
+			if meta.srcFileSize > sizeThreshold && sfa.commandType == uploadCommand {
 				meta.parallel = sfa.parallel
 				largeFileMetas = append(largeFileMetas, meta)
 			} else {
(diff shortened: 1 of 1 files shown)


file_transfer_agent.go
```diff
@@ -697,31 +697,19 @@ func (sfa *snowflakeFileTransferAgent) upload(
 }
 
 func (sfa *snowflakeFileTransferAgent) download(
-	largeFileMetadata []*fileMetadata,
-	smallFileMetadata []*fileMetadata) error {
+	fileMetadata []*fileMetadata) error {
```
sfc-gh-pfus, line 700: Can we now join this two lines?

sfc-gh-pfus (approved):

file_transfer_agent.go
sfc-gh-ext-simba-lb: yeah I changed the L197 to only take in the small metadata since large metadata is unused now.

PR #876, opened 2023-08-03
SNOW-870356: Integrate code coverage with CodeCov

sfc-gh-ext-simba-lb:
### Description
Upload code coverage files from Github action runs to code cov which calculates an average.

18 files are uploaded; for the combination of each OS, version and cloud provider. Code cov will merge the files.
https://app.codecov.io/gh/snowflakedb/gosnowflake/commit/43f021f4593d1ada3290a56aa2361e4ebc804372

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -49,6 +49,8 @@ jobs:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
               run: ./ci/test.sh
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
     build-test-mac:
         runs-on: macos-latest
         strategy:
@@ -72,6 +74,8 @@ jobs:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
               run: ./ci/test.sh
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
     build-test-windows:
         runs-on: windows-latest
         strategy:
@@ -99,3 +103,5 @@ jobs:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
               run: ci\\test.bat
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -51,6 +51,8 @@ jobs:
               run: ./ci/test.sh
             - name: Upload coverage to Codecov
               uses: codecov/codecov-action@v3
+              with:
+                token: ${{ secrets.CODE_COV_UPLOAD_TOKEN }}
     build-test-mac:
         runs-on: macos-latest
         strategy:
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (approved):
LGTM. Make sure you have CodeCov badge in the github

.github/workflows/build-test.yml
```diff
@@ -49,6 +49,8 @@ jobs:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
               run: ./ci/test.sh
+            - name: Upload coverage to Codecov
```
sfc-gh-pfus, line 52: I have a couple of questions :) 
1. How it calculates coverage? Is it based on our `coverage.txt`?
2. I'm wondering how it should work. We have n go versions * m cloud providers * k operating systems. So what goes to codecov? All n * m * k values? What should be shown in the badge?
3. How does the authentication to codecov works? Is it some global snowflake integration?
4. There are a lot of comments of codecov below the PR. Can we limit it only to files changed in this PR?

.github/workflows/build-test.yml
```diff
@@ -49,6 +49,8 @@ jobs:
...
+              uses: codecov/codecov-action@v3
```
sfc-gh-dprzybysz, line 53: Yesterday at we talked with @sfc-gh-igarish that codecov has some problems with rate limitting on GH and we should use code cov token

.github/workflows/build-test.yml
sfc-gh-dprzybysz: ad 2. in node js we gather coverage only from one job (MacOS+AWS) and publish it to CodeCov
If you want to publish more than you should add some additional tags to codecov to distiniguish between coverages?

sfc-gh-ext-simba-lb: 1. It uses the formula `hits / (sum of hit + partial + miss)` https://docs.codecov.com/docs/about-code-coverage
2. Codecov merges all the files uploaded. So when the tests are run n * m * k files (18 uploads) are uploaded and merged to get the total coverage.
3. I'm not sure of the auth configuration, it was already set up. Maybe @sfc-gh-igarish can answer this one.
4. The codecov run here is comparing to a base commit from 3 years ago when the gosnowflake project was last uploaded there. There was only 23 files then and now there are 48 plus a lot more code in the files. We need to merge this to get a new baseline.

.github/workflows/build-test.yml
sfc-gh-ext-simba-lb: @sfc-gh-dprzybysz where does it read `${{ secrets.CODE_COV_UPLOAD_TOKEN }}` from? I added it to the build file but in the log it says no token was found although I can see `token=****` in the upload url.

sfc-gh-igarish: @sfc-gh-ext-simba-lb I just put the token for the GO driver. try it now. Yesterday I did it for Node.js.

.github/workflows/build-test.yml
sfc-gh-pfus: 1. Ok, I was wondering how it integrates with coverage.txt, but I see that it is automatic.
2. I wonder what is the policy. But looking at codecov I believe that it combines all lines from all builds to check if a line is tested in any build.
3. Ok.
4. Do we need to do it manually? How to keep codecov updated automatically?

.github/workflows/build-test.yml
sfc-gh-ext-simba-lb: It works now thanks.

.github/workflows/build-test.yml
sfc-gh-ext-simba-lb: After the change is merged it will run against master, same as the [PHP PR](https://github.com/snowflakedb/pdo_snowflake/pull/330).

sfc-gh-pfus (approved):

sfc-gh-dprzybysz (approved):

PR #880, opened 2023-08-07
SNOW-857660: Improve JSON performance

sfc-gh-pfus:
### Description
Before my change, for each row we tried to find out the location for string localisation (like formatting dates etc, even if it was not relevant). Checking location was easy if someone specified the location in the connection parameter, but if it was not (and I believe it is usual), we needed to guess this location. This involved very inefficient system call for each row and each column. I changed it so the location is cached for connection. The code handling JSON responses is like 30-50 times faster (!!!). Quick stats: before my change 1M rows (single column) was retrieved in 1,5min. After my changes it takes 3 seconds. In 1,5min we are able to handle nearly 50M of rows.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/location.go
+++ b/location.go
@@ -91,16 +91,20 @@ func init() {
 }
 
 // retrieve current location based on connection
-func getCurrentLocation(params map[string]*string) *time.Location {
+func (sc *snowflakeConn) getCurrentLocation() *time.Location {
+	if sc.location != nil {
+		return sc.location
+	}
 	loc := time.Now().Location()
 	var err error
 	paramsMutex.Lock()
-	if tz, ok := params["timezone"]; ok {
+	defer paramsMutex.Unlock()
+	if tz, ok := sc.cfg.Params["timezone"]; ok {
 		loc, err = time.LoadLocation(*tz)
 		if err != nil {
 			loc = time.Now().Location()
 		}
 	}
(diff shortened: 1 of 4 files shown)


sfc-gh-pbulawa (approved):

location.go
```diff
@@ -91,16 +91,20 @@ func init() {
...
 	paramsMutex.Lock()
-	if tz, ok := params["timezone"]; ok {
+	defer paramsMutex.Unlock()
+	if tz, ok := sc.cfg.Params["timezone"]; ok {
 		loc, err = time.LoadLocation(*tz)
 		if err != nil {
 			loc = time.Now().Location()
 		}
 	}
-	paramsMutex.Unlock()
```
sfc-gh-igarish, line 104: How to take care of ALTER SESSION command changes some parameter like TIMEZONE? will it be change in the cache? Could you please add test case for it?

sfc-gh-pfus: You are right. With your question in my mind, I started from scratch: https://github.com/snowflakedb/gosnowflake/pull/882

sfc-gh-igarish:
@sfc-gh-pfus Could you please double check why our perf dashboard is not showing 30-50 times improvement? cc: @sfc-gh-jtang

PR #883, opened 2023-08-14
SNOW-892592: Fix panic when SF_OCSP_RESPONSE_CACHE_SERVER_ENABLED=false

sfc-gh-ext-simba-lb:
### Description
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/608
When testing for https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/593 I found a panic when SF_OCSP_RESPONSE_CACHE_SERVER_ENABLED is false. The initOcspCache() function initializes a mutex lock on the ocsp response cache. If SF_OCSP_RESPONSE_CACHE_SERVER_ENABLED=false it exits the init function before the lock is initialized. The driver tries to access it later and it causes a panic during connection.

This change is to initialize the lock first and then check for SF_OCSP_RESPONSE_CACHE_SERVER_ENABLED.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/ocsp_test.go
+++ b/ocsp_test.go
@@ -533,3 +533,21 @@ func TestInitOCSPCacheFileCreation(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestCacheServerDisabled(t *testing.T) {
+	_ = os.Setenv(cacheServerEnabledEnv, "false")
+
+	if _, err := buildSnowflakeConn(context.Background(), Config{
+		Account:  "testaccount",
+		User:     "testuser",
+		Password: "testpassword",
+		Host:     "testaccount.us-east-1.snowflakecomputing.com",
+	}); err != nil {
+		t.Error(err)
+	}
+
+	ocspURL := os.Getenv(cacheServerURLEnv)
+	if ocspURL != "" {
+		t.Fatalf("cache server is disabled. Expected empty url but got %v ", cacheServerURLEnv)
+	}
+}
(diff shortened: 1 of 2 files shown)


ocsp_test.go
```diff
@@ -533,3 +533,21 @@ func TestInitOCSPCacheFileCreation(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestCacheServerDisabled(t *testing.T) {
+	_ = os.Setenv(cacheServerEnabledEnv, "false")
+
+	if _, err := buildSnowflakeConn(context.Background(), Config{
```
sfc-gh-pfus, line 540: Can extract building sf conn outside of the `if`? Or at least `Config`? Now it's not very readable.

ocsp_test.go
```diff
@@ -533,3 +533,21 @@ func TestInitOCSPCacheFileCreation(t *testing.T) {
...
+		Account:  "testaccount",
+		User:     "testuser",
+		Password: "testpassword",
+		Host:     "testaccount.us-east-1.snowflakecomputing.com",
+	}); err != nil {
+		t.Error(err)
+	}
+
+	ocspURL := os.Getenv(cacheServerURLEnv)
+	if ocspURL != "" {
```
sfc-gh-pfus, line 550: I'm not sure if I understand it. According to the first line in the test, you just set an env variable and now you read another one. I believe this just reads env variable, there is no driver logic. The only reason why this could work is that `buildSnowflakeConn` removes this env var, which seems unlike. But if this should work, maybe we should enrich the test with setting `cacheServerUrlEnv` at the beginning of the test?

sfc-gh-ext-simba-jl:
Closing this PR since I can't reproduce the original issue.

PR #889, opened 2023-08-21
Snow 895534 add entries to cache

sfc-gh-pfus:
### Description
When HTAP query context entries are returned from the server, we save it in the local cache.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/htap.go
+++ b/htap.go
@@ -1,8 +1,26 @@
 package gosnowflake
 
+import "sync"
+
 type queryContextEntry struct {
 	ID        int   `json:"id"`
 	Timestamp int64 `json:"timestamp"`
 	Priority  int   `json:"priority"`
 	Context   any   `json:"context,omitempty"`
+}
+
+type queryContextCache struct {
+	mutex   *sync.Mutex
+	entries []queryContextEntry
+}
+
+func (qcc *queryContextCache) init() *queryContextCache {
+	qcc.mutex = &sync.Mutex{}
+	return qcc
+}
+
+func (qcc *queryContextCache) add(qces ...queryContextEntry) {
+	qcc.mutex.Lock()
+	defer qcc.mutex.Unlock()
+	qcc.entries = append(qcc.entries, qces...)
 }
--- a/htap_test.go
+++ b/htap_test.go
@@ -100,3 +100,50 @@ func trimWhitespaces(s string) string {
 		"\n", "",
 	)
 }
+
+func TestAddingQueryContextCacheEntry(t *testing.T) {
+	runSnowflakeConnTest(t, func(sc *snowflakeConn) {
+		t.Run("First query (may be on empty cache)", func(t *testing.T) {
+			entriesBefore := sc.queryContextCache.entries
+			if _, err := sc.Query("SELECT 1", nil); err != nil {
+				t.Fatalf("cannot query. %v", err)
+			}
+			entriesAfter := sc.queryContextCache.entries
+
+			if !containsNewEntries(entriesAfter, entriesBefore) {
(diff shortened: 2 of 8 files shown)


htap.go
```diff
@@ -1,8 +1,33 @@
...
+func (qcc *queryContextCache) add(qce queryContextEntry) {
+	qcc.mutex.Lock()
+	defer qcc.mutex.Unlock()
+
+	qcc.entries = append(qcc.entries, qce)
+}
+
+func (qcc *queryContextCache) addAll(qces []queryContextEntry) {
+	for _, qce := range qces {
+		qcc.add(qce)
```
sfc-gh-dprzybysz, line 31: should we lock append for each item in the list or we could do it once for the whole list?

htap_test.go
```diff
@@ -100,3 +100,52 @@ func trimWhitespaces(s string) string {
...
+			entriesBefore := sc.queryContextCache.entries
+			if len(entriesBefore) == 0 {
+				t.Fatalf("cache should not be empty after first query")
+			}
+			if _, err := sc.Query("SELECT 1", nil); err != nil {
+				t.Fatalf("cannot query. %v", err)
+			}
+			entriesAfter := sc.queryContextCache.entries
+
+			if len(entriesAfter) > len(entriesBefore) {
```
sfc-gh-dprzybysz, line 137: let's extract a function e.g. `cacheEntriesChanged(entriesBefore, entriesAfter)` returning bool and base on that fail test or not

htap.go
sfc-gh-pfus: This operates on connection objects, which shouldn't be used among threads.

htap_test.go
sfc-gh-pfus: Done.

htap.go
sfc-gh-dprzybysz: my concerns were more about calling multiple times Lock and Unlock (which have its cost?) where Lock and Unlock can be used around for statement

sfc-gh-pfus: I would say, that the overhead here is completely negligible. We shouldn't get here more than 5 elements, and from my testing it's rather one (for non HTAP queries) to two (for HTAP queries).

sfc-gh-dprzybysz: We should think about future cases - now there are only one or two entries but do we know how it will change in the future?

sfc-gh-pfus: No, this works like this. The won't be more than 5 entries. Or to be precise - more than `QUERY_CONTEXT_CACHE_SIZE`, which is 5.

sfc-gh-pfus: Ok, I inlined this method, so I'm resolving this thread as outdated.

sfc-gh-pmotacki (approved):
LGTM

sfc-gh-igarish:
I will review it but please check why tests are failing.

htap.go
```diff
@@ -1,8 +1,28 @@
...
 type queryContextEntry struct {
 	ID        int   `json:"id"`
 	Timestamp int64 `json:"timestamp"`
 	Priority  int   `json:"priority"`
 	Context   any   `json:"context,omitempty"`
 }
+
+type queryContextCache struct {
+	mutex   *sync.Mutex
+	entries []queryContextEntry
```
sfc-gh-igarish, line 14: How to restrict the size of the cache? This value comes from server, if you don't see value then initialize with 5 Max entries in the cache.

sfc-gh-pfus: Yes, I have a task for it: https://snowflakecomputing.atlassian.net/browse/SNOW-895536
But I wanted to do it after the implementing prioritizing.

PR #890, opened 2023-08-21
SNOW-889572 Refactored snowflake type implementation to map

sfc-gh-dheyman-1:
### Description

### Checklist
- [] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/datatype.go
+++ b/datatype.go
@@ -32,27 +32,45 @@ const (
 	unSupportedType
 )
 
-var snowflakeTypes = [...]string{"FIXED", "REAL", "TEXT", "DATE", "VARIANT",
-	"TIMESTAMP_LTZ", "TIMESTAMP_NTZ", "TIMESTAMP_TZ", "OBJECT", "ARRAY",
-	"BINARY", "TIME", "BOOLEAN", "NULL", "SLICE", "CHANGE_TYPE", "NOT_SUPPORTED"}
+var stringToSnowflakeType = map[string]snowflakeType{
+	"FIXED":         fixedType,
+	"REAL":          realType,
+	"TEXT":          textType,
+	"DATE":          dateType,
+	"VARIANT":       variantType,
+	"TIMESTAMP_LTZ": timestampLtzType,
+	"TIMESTAMP_NTZ": timestampNtzType,
+	"TIMESTAMP_TZ":  timestampTzType,
+	"OBJECT":        objectType,
+	"ARRAY":         arrayType,
+	"BINARY":        binaryType,
+	"TIME":          timeType,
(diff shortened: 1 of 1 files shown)


datatype.go
```diff
@@ -32,27 +32,45 @@ const (
 	unSupportedType
 )
 
-var snowflakeTypes = [...]string{"FIXED", "REAL", "TEXT", "DATE", "VARIANT",
-	"TIMESTAMP_LTZ", "TIMESTAMP_NTZ", "TIMESTAMP_TZ", "OBJECT", "ARRAY",
-	"BINARY", "TIME", "BOOLEAN", "NULL", "SLICE", "CHANGE_TYPE", "NOT_SUPPORTED"}
+var stringToSnowflakeType = map[string]snowflakeType{
```
sfc-gh-pfus, line 35: I wonder if we can rename it to be more precise. Because the first one is actual snowflake type and the second is the driver (go?) type. WDYT?

datatype.go
```diff
@@ -32,27 +32,45 @@ const (
...
+	"NULL":          nullType,
+	"SLICE":         sliceType,
+	"CHANGE_TYPE":   changeType,
+	"NOT_SUPPORTED": unSupportedType}
 
-func (st snowflakeType) String() string {
-	return snowflakeTypes[st]
+var snowflakeTypeToString = invertMap(stringToSnowflakeType)
+
+func invertMap(m map[string]snowflakeType) map[snowflakeType]string {
```
sfc-gh-pfus, line 56: I wonder what would happen if some duplication occurred here... Is it only used for `String` method? Is it safe?

sfc-gh-dheyman-1: Added check for duplicated values

datatype.go
sfc-gh-dheyman-1: Changed names to be more descriptive

datatype.go
```diff
@@ -32,27 +32,48 @@ const (
...
+	"NOT_SUPPORTED": unSupportedType}
 
-func (st snowflakeType) String() string {
-	return snowflakeTypes[st]
+var driverTypeToSnowflake = invertMap(snowflakeToDriverType)
+
+func invertMap(m map[string]snowflakeType) map[snowflakeType]string {
+	inv := make(map[snowflakeType]string)
+	for k, v := range m {
+		if _, exists := inv[v]; exists {
```
sfc-gh-pfus, line 59: Naming this variable `ok` will give us more go-like naming :)

sfc-gh-pfus (approved):

datatype.go
sfc-gh-dheyman-1: Done!

PR #899, opened 2023-08-28
SNOW-726742: Implement GetQueryId for statements

sfc-gh-pfus:
### Description
Implement GetQueryId on statements

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -7,9 +7,15 @@ import (
 	"database/sql/driver"
 )
 
+// SnowflakeStmt represents the prepared statement in driver.
+type SnowflakeStmt interface {
+	GetQueryID() string
+}
+
 type snowflakeStmt struct {
-	sc    *snowflakeConn
-	query string
+	sc          *snowflakeConn
+	query       string
+	lastQueryID string
 }
 
 func (stmt *snowflakeStmt) Close() error {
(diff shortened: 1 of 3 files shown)


statement.go
```diff
@@ -7,9 +7,15 @@ import (
 	"database/sql/driver"
 )
 
+// SnowflakeStmt represents the prepared statement in driver.
```
sfc-gh-dprzybysz, line 10: don't we want to name it `PreparedSnowflakeStmt` then? it looks really similar to `snowflakeStmt` struct defined below

sfc-gh-pfus: No, and there are two reasons:
1. This is a common pattern in go - you have a private struct (name starting with small letter) and public interface (name starting with capital letter), which differentiate only by this first letter. This is something like public interface and private implementation.
2. Go SQL uses name `stmt` in code and `prepared statement` only in docs. We have this convention to enrich with our interfaces by prepending with `Snowflake` (see `driver.Rows` -> `SnowflakeRows` or `driver.Result` -> `SnowflakeResult`).

sfc-gh-dheyman-1 (approved):

statement_test.go
```diff
@@ -287,3 +288,129 @@ func TestUnitCheckQueryStatus(t *testing.T) {
...
+					t.Error("queryId should be empty before executing any query")
+				}
+				firstQuery, err := tc.f(stmt)
+				if err != nil {
+					t.Fatal(err)
+				}
+				if stmt.(SnowflakeStmt).GetQueryID() == "" {
+					t.Error("queryId should not be empty after executing query")
+				}
+				if stmt.(SnowflakeStmt).GetQueryID() != firstQuery.(SnowflakeRows).GetQueryID() {
```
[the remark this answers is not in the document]
sfc-gh-pfus: I don't think this is a good idea. This would be more java-way, where having private methods in tests is common. Here, in go, this is not the way.

sfc-gh-igarish (approved):
LGTM

PR #900, opened 2023-08-28
SNOW-895535: Prioritise entries in query context cache

sfc-gh-pfus:
### Description
Introduced algorithm of prioritising entries in query context cache.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/htap.go
+++ b/htap.go
@@ -1,6 +1,7 @@
 package gosnowflake
 
 import (
+	"sort"
 	"strconv"
 	"sync"
 )
@@ -33,7 +34,36 @@ func (qcc *queryContextCache) add(sc *snowflakeConn, qces ...queryContextEntry) 
 	if len(qces) == 0 {
 		qcc.prune(0)
 	} else {
-		qcc.entries = append(qcc.entries, qces...)
+		for _, newQce := range qces {
+			logger.Debugf("adding query context: %v", newQce)
+			newQceProcessed := false
+			for existingQceIdx, existingQce := range qcc.entries {
+				if newQce.ID == existingQce.ID {
+					newQceProcessed = true
+					if newQce.Timestamp > existingQce.Timestamp {
+						qcc.entries[existingQceIdx] = newQce
+					} else if newQce.Timestamp == existingQce.Timestamp {
+						if newQce.Priority != existingQce.Priority {
+							qcc.entries[existingQceIdx] = newQce
+						}
+					}
+				}
+			}
+			if !newQceProcessed {
+				for existingQceIdx, existingQce := range qcc.entries {
+					if newQce.Priority == existingQce.Priority {
+						qcc.entries[existingQceIdx] = newQce
+						newQceProcessed = true
+					}
+				}
+			}
+			if !newQceProcessed {
+				qcc.entries = append(qcc.entries, newQce)
+			}
+		}
+		sort.Slice(qcc.entries, func(idx1, idx2 int) bool {
+			return qcc.entries[idx1].Priority < qcc.entries[idx2].Priority
+		})
 		qcc.prune(qcc.getQueryContextCacheSize(sc))
 	}
 }
--- a/htap_test.go
+++ b/htap_test.go
@@ -101,10 +101,166 @@ func trimWhitespaces(s string) string {
 	)
 }
 
+func TestSortingByPriority(t *testing.T) {
+	qcc := (&queryContextCache{}).init()
+	sc := htapTestSnowflakeConn()
+
+	qceA := queryContextEntry{ID: 12, Timestamp: 123, Priority: 7, Context: "a"}
+	qceB := queryContextEntry{ID: 13, Timestamp: 124, Priority: 9, Context: "b"}
+	qceC := queryContextEntry{ID: 14, Timestamp: 125, Priority: 6, Context: "c"}
+	qceD := queryContextEntry{ID: 15, Timestamp: 126, Priority: 8, Context: "d"}
+
+	t.Run("Add to empty cache", func(t *testing.T) {
+		qcc.add(sc, qceA)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA}) {
+			t.Fatalf("no entries added to cache. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different id, timestamp and priority - greater priority", func(t *testing.T) {
+		qcc.add(sc, qceB)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("unexpected qcc entries. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different id, timestamp and priority - lesser priority", func(t *testing.T) {
+		qcc.add(sc, qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceA, qceB}) {
+			t.Fatalf("unexpected qcc entries. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different id, timestamp and priority - priority in the middle", func(t *testing.T) {
+		qcc.add(sc, qceD)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceA, qceD, qceB}) {
+			t.Fatalf("unexpected qcc entries. %v", qcc.entries)
+		}
+	})
+}
+
+func TestAddingQcesWithTheSameIdAndLaterTimestamp(t *testing.T) {
+	qcc := (&queryContextCache{}).init()
+	sc := htapTestSnowflakeConn()
+
+	qceA := queryContextEntry{ID: 12, Timestamp: 123, Priority: 7, Context: "a"}
+	qceB := queryContextEntry{ID: 13, Timestamp: 124, Priority: 9, Context: "b"}
+	qceC := queryContextEntry{ID: 12, Timestamp: 125, Priority: 6, Context: "c"}
+	qceD := queryContextEntry{ID: 12, Timestamp: 126, Priority: 6, Context: "d"}
+
+	t.Run("Add to empty cache", func(t *testing.T) {
+		qcc.add(sc, qceA)
+		qcc.add(sc, qceB)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("no entries added to cache. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(sc, qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceB}) {
+			t.Fatalf("unexpected qcc entries. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) {
+		qcc.add(sc, qceD)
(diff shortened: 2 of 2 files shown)


sfc-gh-igarish (approved):
LGTM

htap.go
```diff
@@ -22,5 +25,41 @@ func (qcc *queryContextCache) init() *queryContextCache {
...
-	qcc.entries = append(qcc.entries, qces...)
+	var newQcesBasedOnPriorities []queryContextEntry
+	for _, newQce := range qces {
+		logger.Debugf("adding query context: %v", newQce)
+		newQceProcessed := false
+		for existingQceIdx, existingQce := range qcc.entries {
+			if newQce.ID == existingQce.ID {
+				newQceProcessed = true
+				if newQce.Timestamp > existingQce.Timestamp {
+					if newQce.Priority == existingQce.Priority {
```
sfc-gh-dprzybysz, line 36: cannot we here just replace item in the list as we do in else branch?

htap.go
```diff
@@ -22,5 +25,41 @@ func (qcc *queryContextCache) init() *queryContextCache {
...
+						qcc.entries[existingQceIdx].Timestamp = newQce.Timestamp
+						qcc.entries[existingQceIdx].Context = newQce.Context
+					} else {
+						qcc.entries[existingQceIdx] = newQce
+					}
+				} else if newQce.Timestamp == existingQce.Timestamp {
+					if newQce.Priority != existingQce.Priority {
```
sfc-gh-dprzybysz, line 43: could we join if conditions and don't have nested one?

htap.go
```diff
@@ -22,5 +25,41 @@ func (qcc *queryContextCache) init() *queryContextCache {
...
+						qcc.entries[existingQceIdx].Priority = newQce.Priority
+						qcc.entries[existingQceIdx].Context = newQce.Context
+					}
+				}
+			} else {
+				if newQce.Priority == existingQce.Priority {
+					qcc.entries[existingQceIdx].ID = newQce.ID
```
sfc-gh-dprzybysz, line 50: cannot we just override the whole item in the list?

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+			t.Fatalf("no entries added to cache. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) { // TODO should anything happen here?
```
sfc-gh-dprzybysz, line 188: when are we going to resolve this issue? let's add issue id

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+			t.Fatalf("no entries added to cache. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) { // TODO should anything happen here?
```
sfc-gh-dprzybysz, line 217: let's address this todo comment

htap.go
sfc-gh-pfus: Actually we can.

htap.go
sfc-gh-pfus: For me it is more readable, as this upper `else if` is connected to `if` also checking timestamp. So this level of indentation is about timestamps and the lower one is another is for priority.

htap.go
sfc-gh-pfus: Let's try this out :)

htap_test.go
sfc-gh-pfus: Removed.

htap_test.go
sfc-gh-pfus: Removed.

htap.go
```diff
@@ -22,5 +25,33 @@ func (qcc *queryContextCache) init() *queryContextCache {
...
 	qcc.mutex.Lock()
 	defer qcc.mutex.Unlock()
-	qcc.entries = append(qcc.entries, qces...)
+	var newQcesBasedOnPriorities []queryContextEntry
+	for _, newQce := range qces {
+		logger.Debugf("adding query context: %v", newQce)
+		newQceProcessed := false
+		for existingQceIdx, existingQce := range qcc.entries {
+			if newQce.ID == existingQce.ID {
+				newQceProcessed = true
```
sfc-gh-dprzybysz, line 34: we want to keep it true when newQce.Timestamp is lower than the existing one?

sfc-gh-dprzybysz (approved):

htap.go
sfc-gh-pfus: Yes. This is because here we considered it and decided that we don't use that qce. If we don't it would be added either way at the end of the loop.

htap.go
```diff
@@ -22,5 +25,33 @@ func (qcc *queryContextCache) init() *queryContextCache {
 func (qcc *queryContextCache) add(qces ...queryContextEntry) {
 	qcc.mutex.Lock()
 	defer qcc.mutex.Unlock()
-	qcc.entries = append(qcc.entries, qces...)
+	var newQcesBasedOnPriorities []queryContextEntry
```
sfc-gh-dstempniak, line 28: What is newQcesBasedOnPriorities for? I don't see it filled.

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+		qcc.add(qceA)
+		qcc.add(qceB)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("no entries added to cache. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
```
sfc-gh-dstempniak, line 156: so why it is added at the beggining? and maybe it is worth adding "and should replace entry with same ID"?

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) {
+		qcc.add(qceD)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceD, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
```
sfc-gh-dstempniak, line 162: so why it is added at the beggining?

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) {
+		qcc.add(qceD)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceC, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
```
sfc-gh-dstempniak, line 191: but qceD wasn't added....

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+	t.Run("Add another entry with different priority", func(t *testing.T) {
+		qcc.add(qceC)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
+		}
+	})
+	t.Run("Add another entry with same priority", func(t *testing.T) {
+		qcc.add(qceD)
+		if !reflect.DeepEqual(qcc.entries, []queryContextEntry{qceA, qceB}) {
+			t.Fatalf("entry with greater priority should be added at the end. %v", qcc.entries)
```
sfc-gh-dstempniak, line 220: but C and D wasn't added

htap.go
```diff
@@ -22,5 +25,33 @@ func (qcc *queryContextCache) init() *queryContextCache {
...
+					qcc.entries[existingQceIdx] = newQce
+				} else if newQce.Timestamp == existingQce.Timestamp {
+					if newQce.Priority != existingQce.Priority {
+						qcc.entries[existingQceIdx] = newQce
+					}
+				}
+			} else {
+				if newQce.Priority == existingQce.Priority {
+					qcc.entries[existingQceIdx] = newQce
+					newQceProcessed = true
```
sfc-gh-dstempniak, line 45: Don't we need any break/exit after setting newQceProcessed to true?
Lets assume scenario: newQce has same ID as one qcc element and same Priority as other qcc element. If I'm not wrong it will replace two qcc elements.

htap_test.go
```diff
@@ -101,10 +101,161 @@ func trimWhitespaces(s string) string {
...
+		}
+	})
+}
+
+func TestAddingQcesWithTheSameIdAndSameTimestamp(t *testing.T) {
+	qcc := (&queryContextCache{}).init()
+
+	qceA := queryContextEntry{ID: 12, Timestamp: 123, Priority: 7, Context: "a"}
+	qceB := queryContextEntry{ID: 13, Timestamp: 124, Priority: 9, Context: "b"}
+	qceC := queryContextEntry{ID: 12, Timestamp: 123, Priority: 6, Context: "c"}
```
sfc-gh-dstempniak, line 172: What if qceC has priority = 9? I suppose it will also replace qceB.

sfc-gh-pfus: I believe this is not possible, as priorities and id returned from server are unique.

htap_test.go
sfc-gh-pfus: Copied wrong comment, fixed!

htap_test.go
sfc-gh-pfus: Copied wrong comment, fixed!

htap.go
sfc-gh-pfus: Confirmed with the server team (@sfc-gh-lqin ) that this is not possible scenario.

htap_test.go
sfc-gh-lqin: I think you can check the implementation in JDBC/Python connector side as a reference, and also their unit testcases.

sfc-gh-pfus: They don't have a test for that case :) And I checked the impl as well.

sfc-gh-lqin: Another suggestion I mentioned in slack is to run multiple complex queries which are multi-db queries and check if the query context contents are updated in a reasonable way.

sfc-gh-pfus: Yeah, but what I asked you on Slack was about some E2E tests. This is only unit test. I'll provide E2E test when I have whole implementation finished.

sfc-gh-dstempniak (approved):

PR #904, opened 2023-08-30
SNOW-895537: Send query context with request

sfc-gh-pfus:
### Description
Query contexts are now sent to snowflake backend.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -85,6 +85,10 @@ func (sc *snowflakeConn) exec(
 	var err error
 	counter := atomic.AddUint64(&sc.SequenceCounter, 1) // query sequence counter
 
+	queryContext, err := buildQueryContext(sc.queryContextCache)
+	if err != nil {
+		logger.Errorf("error while building query context: %v", err)
+	}
 	req := execRequest{
 		SQLText:      query,
 		AsyncExec:    noResult,
@@ -92,6 +96,7 @@ func (sc *snowflakeConn) exec(
 		IsInternal:   isInternal,
 		DescribeOnly: describeOnly,
 		SequenceID:   counter,
+		QueryContext: queryContext,
 	}
 	if key := ctx.Value(multiStatementCount); key != nil {
 		req.Parameters[string(multiStatementCount)] = key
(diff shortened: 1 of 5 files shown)


connection.go
```diff
@@ -85,13 +86,18 @@ func (sc *snowflakeConn) exec(
 	var err error
 	counter := atomic.AddUint64(&sc.SequenceCounter, 1) // query sequence counter
 
+	queryContext, err := buildQueryContext(sc.queryContextCache)
+	if err != nil {
+		logger.Warnf("error while building query context: %v", err)
```
sfc-gh-dheyman-1, line 91: I think this should be error.

connection.go
```diff
@@ -160,6 +166,45 @@ func (sc *snowflakeConn) exec(
 	return data, err
 }
 
+func buildQueryContext(qcc *queryContextCache) (requestQueryContext, error) {
```
sfc-gh-dheyman-1, line 169: don't we want to extract these two methods to another file? connection.go is getting rather large.

connection.go
```diff
@@ -160,6 +166,45 @@ func (sc *snowflakeConn) exec(
...
+			Timestamp: qce.Timestamp,
+			Context: contextData{
+				Base64Data: contextBase64,
+			},
+		})
+	}
+	return rqc, nil
+}
+
+func marshallContext(context any) (string, error) {
```
sfc-gh-dheyman-1, line 192: `marshalContext` EDIT: actually a matter of US spelling, so no need to change.

sfc-gh-dheyman-1 (approved):

sfc-gh-dstempniak (approved):

connection.go
sfc-gh-pfus: Right, done.

connection.go
sfc-gh-pfus: This one is pretty internal. But I agree, that we should reorganise files a bit, but in different commit.

htap_test.go
```diff
@@ -416,3 +419,79 @@ func htapTestSnowflakeConn() *snowflakeConn {
...
+		})
+		t.Run("Run tests on first database again", func(t *testing.T) {
+			sct.mustExec(fmt.Sprintf("USE DATABASE %v", testDb1), nil)
+
+			sct.mustExec("INSERT INTO test_hybrid_table VALUES (4, 'd')", nil)
+
+			rows := sct.mustQuery("SELECT * FROM test_hybrid_table", nil)
+			defer rows.Close()
+			if len(sct.sc.queryContextCache.entries) != 3 {
+				t.Errorf("expected two entries in query context cache, got: %v", sct.sc.queryContextCache.entries)
```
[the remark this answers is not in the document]
sfc-gh-pfus: Really appreciate, that someone reads so carefully and finds my easter eggs!

connection.go
```diff
@@ -85,13 +85,18 @@ func (sc *snowflakeConn) exec(
 	var err error
 	counter := atomic.AddUint64(&sc.SequenceCounter, 1) // query sequence counter
 
+	queryContext, err := buildQueryContext(sc.queryContextCache)
+	if err != nil {
+		logger.Errorf("error while building query context: %v", err)
```
sfc-gh-igarish, line 90: Do we continue even if error occurred?

sfc-gh-pfus: Yes. This is confirmed with Lianke, that we should proceed and just not send query context.

PR #915, opened 2023-09-15
Snow 893925 ocsp responders in privatelink

sfc-gh-dheyman-1:
### Description

### Checklist
- [x] Code compiles correctly
- [x] All tests passing


ocsp.go
```diff
@@ -428,15 +428,65 @@ func retryOCSP(
...
-	logger.Debug("reading contents")
 	ocspResBytes, err = io.ReadAll(res.Body)
 	if err != nil {
 		return ocspRes, ocspResBytes, &ocspStatus{
 			code: ocspFailedExtractResponse,
 			err:  err,
 		}
 	}
-	logger.Debug("parsing OCSP response")
+	ocspRes, err = ocsp.ParseResponse(ocspResBytes, issuer)
```
sfc-gh-dprzybysz, line 438: are we able to test added code? e.g. with hang webserver?

connection_util.go
```diff
@@ -281,10 +281,12 @@ func populateChunkDownloader(
 
 func (sc *snowflakeConn) setupOCSPPrivatelink(app string, host string) error {
 	ocspCacheServer := fmt.Sprintf("http://ocsp.%v/ocsp_response_cache.json", host)
+	logger.Debugf("OCSP Cache Server for Privatelink: %v\n", ocspCacheServer)
 	if err := os.Setenv(cacheServerURLEnv, ocspCacheServer); err != nil {
 		return err
 	}
 	ocspRetryHost := fmt.Sprintf("http://ocsp.%v/retry/", host) + "%v/%v"
```
sfc-gh-pfus, line 288: If this is a template maybe we should include it in variable name?

ocsp.go
```diff
@@ -428,15 +428,65 @@ func retryOCSP(
...
+	if err != nil {
+		logger.Warnf("error when parsing ocsp response: %v\n", err)
```
sfc-gh-pfus, line 440: Maybe we don't need this `\n`?

ocsp.go
```diff
@@ -445,11 +495,36 @@ func retryOCSP(
...
+	}
+}
+
+func fullOCSPURL(url *url.URL) string {
+	fullURL := url.Hostname()
+	if url.Path != "" {
+		if !strings.HasPrefix(url.Path, "/") {
+			fullURL += "/"
+		}
+		fullURL += url.Path
```
sfc-gh-pfus, line 523: Maybe we can move inside if?

ocsp.go
```diff
@@ -484,9 +559,14 @@ func getRevocationStatus(ctx context.Context, subject, issuer *x509.Certificate)
 	hostnameStr := os.Getenv(ocspTestResponderURLEnv)
 	var hostname string
 	if retryURL := os.Getenv(ocspRetryURLEnv); retryURL != "" {
-		hostname = fmt.Sprintf(retryURL, u.Hostname(), base64.StdEncoding.EncodeToString(ocspReq))
+		hostname = fmt.Sprintf(retryURL, fullOCSPURL(u), base64.StdEncoding.EncodeToString(ocspReq))
```
sfc-gh-pfus, line 562: We should encode OCSP url here, right?

sfc-gh-dheyman-1: We used existing ODBC solution which was confirmed to work.

ocsp.go
sfc-gh-dheyman-1: We cannot because we the upper if checks for leading slash.

ocsp.go
sfc-gh-dheyman-1: removed

connection_util.go
sfc-gh-dheyman-1: added suffix `Template`

ocsp.go
sfc-gh-dheyman-1: IMHO it is not worth it to write an automatic test, since this is such a specific case and it would require a lot effort. We already have a confluence page that describes how to manually test Privatelink related issues.

sfc-gh-dprzybysz (approved):

PR #916, opened 2023-09-15
SNOW-911238: Add WithOriginalTimestamp context switch

sfc-gh-pfus:
### Description
Add WithOriginalTimestamp context switch.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/arrow_chunk.go
+++ b/arrow_chunk.go
@@ -50,14 +50,14 @@ func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType, hig
 	return chunkRows, arc.reader.Err()
 }
 
-func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader) (*[]arrow.Record, error) {
+func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader, originalTimestamp bool) (*[]arrow.Record, error) {
 	var records []arrow.Record
 	defer arc.reader.Release()
 
 	for arc.reader.Next() {
 		rawRecord := arc.reader.Record()
 
-		record, err := arrowToRecord(rawRecord, arc.allocator, scd.RowSet.RowType, arc.loc)
+		record, err := arrowToRecord(rawRecord, arc.allocator, scd.RowSet.RowType, arc.loc, originalTimestamp)
 		if err != nil {
 			return nil, err
 		}
--- a/arrow_chunk.go
+++ b/arrow_chunk.go
@@ -50,14 +50,14 @@ func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType, hig
 	return chunkRows, arc.reader.Err()
 }
 
-func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader, originalTimestamp bool) (*[]arrow.Record, error) {
+func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader, useOriginalTimestamp bool) (*[]arrow.Record, error) {
 	var records []arrow.Record
 	defer arc.reader.Release()
 
 	for arc.reader.Next() {
 		rawRecord := arc.reader.Record()
 
-		record, err := arrowToRecord(rawRecord, arc.allocator, scd.RowSet.RowType, arc.loc, originalTimestamp)
+		record, err := arrowToRecord(rawRecord, arc.allocator, scd.RowSet.RowType, arc.loc, useOriginalTimestamp)
 		if err != nil {
 			return nil, err
 		}
--- a/cmd/arrow/batches/arrow_batches.go
+++ b/cmd/arrow/batches/arrow_batches.go
@@ -39,8 +39,8 @@ func main() {
 		{Name: "Host", EnvName: "SNOWFLAKE_TEST_HOST", FailOnMissing: false},
 		{Name: "Port", EnvName: "SNOWFLAKE_TEST_PORT", FailOnMissing: false},
 		{Name: "Protocol", EnvName: "SNOWFLAKE_TEST_PROTOCOL", FailOnMissing: false},
-		{Name: "Database", EnvName: "SNOWFLAKE_TEST_DATABASE", FailOnMissing: true},
-		{Name: "Schema", EnvName: "SNOWFLAKE_TEST_SCHEMA", FailOnMissing: true},
+		{Name: "Database", EnvName: "SNOWFLAKE_TEST_DATABASE", FailOnMissing: false},
+		{Name: "Schema", EnvName: "SNOWFLAKE_TEST_SCHEMA", FailOnMissing: false},
 		{Name: "Role", EnvName: "SNOWFLAKE_TEST_ROLE", FailOnMissing: false},
 	})
 	if err != nil {
--- a/cmd/arrow/batches/arrow_batches.go
+++ b/cmd/arrow/batches/arrow_batches.go
@@ -9,10 +9,10 @@ import (
 	"github.com/apache/arrow/go/v12/arrow"
 	"github.com/apache/arrow/go/v12/arrow/array"
 	"github.com/apache/arrow/go/v12/arrow/memory"
+	sf "github.com/snowflakedb/gosnowflake"
 	"log"
 	"sync"
-
-	sf "github.com/snowflakedb/gosnowflake"
+	"time"
 )
 
 type sampleRecord struct {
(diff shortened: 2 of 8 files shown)


arrow_chunk.go
```diff
@@ -50,14 +50,14 @@ func (arc *arrowResultChunk) decodeArrowChunk(rowType []execResponseRowType, hig
 	return chunkRows, arc.reader.Err()
 }
 
-func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader) (*[]arrow.Record, error) {
+func (arc *arrowResultChunk) decodeArrowBatch(scd *snowflakeChunkDownloader, originalTimestamp bool) (*[]arrow.Record, error) {
```
sfc-gh-dprzybysz, line 53: should it be more "useOriginamTimestamp" as it is a flag?

cmd/arrow/batches/arrow_batches.go
```diff
@@ -38,6 +39,9 @@ func main() {
 		{Name: "Host", EnvName: "SNOWFLAKE_TEST_HOST", FailOnMissing: false},
 		{Name: "Port", EnvName: "SNOWFLAKE_TEST_PORT", FailOnMissing: false},
 		{Name: "Protocol", EnvName: "SNOWFLAKE_TEST_PROTOCOL", FailOnMissing: false},
+		{Name: "Database", EnvName: "SNOWFLAKE_TEST_DATABASE", FailOnMissing: true},
+		{Name: "Schema", EnvName: "SNOWFLAKE_TEST_SCHEMA", FailOnMissing: true},
```
sfc-gh-dprzybysz, line 43: should it really fail on missing database or schema paramter?

converter.go
```diff
@@ -354,6 +354,74 @@ func decimalToBigFloat(num decimal128.Num, scale int64) *big.Float {
 	return new(big.Float).Quo(f, s)
 }
 
+// ArrowSnowflakeTimestampToTime converts original timestamp returned by Snowflake to time.TIme
```
sfc-gh-dprzybysz, line 357: typo `TIme`?

converter_test.go
```diff
@@ -1299,30 +1599,135 @@ func TestSmallTimestampBinding(t *testing.T) {
...
+				}
+			}
 		}
+	})
+}
 
-		rows := sct.mustQueryContext(ctx, "SELECT ?", parameters)
-		defer rows.Close()
+func TestTimestampConversionWithArrowBatchesAndWithOriginalTimestamp(t *testing.T) {
+	timestamps := [3]string{
```
sfc-gh-dprzybysz, line 1679: we have this list defined multiple timest in the tests?

converter_test.go
```diff
@@ -1299,30 +1599,135 @@ func TestSmallTimestampBinding(t *testing.T) {
...
+		"2000-10-10 10:10:10.123456789", // neutral
+		"9999-12-12 23:59:59.123456789", // max
```
sfc-gh-dprzybysz, line 1681: max should have only nines in the last part?

converter_test.go
```diff
@@ -1299,30 +1599,135 @@ func TestSmallTimestampBinding(t *testing.T) {
...
+		"0001-01-01 00:00:00.123456789"} // min
```
sfc-gh-dprzybysz, line 1682: min should have only zeros in the last part?

util.go
```diff
@@ -18,16 +18,17 @@ import (
...
+	asyncMode               contextKey = "ASYNC_MODE_QUERY"
+	queryIDChannel          contextKey = "QUERY_ID_CHANNEL"
+	snowflakeRequestIDKey   contextKey = "SNOWFLAKE_REQUEST_ID"
+	fetchResultByID         contextKey = "SF_FETCH_RESULT_BY_ID"
+	fileStreamFile          contextKey = "STREAMING_PUT_FILE"
+	fileTransferOptions     contextKey = "FILE_TRANSFER_OPTIONS"
+	enableHigherPrecision   contextKey = "ENABLE_HIGHER_PRECISION"
+	arrowBatches            contextKey = "ARROW_BATCHES"
+	arrowAlloc              contextKey = "ARROW_ALLOC"
+	enableOriginalTimestamp contextKey = "ENABLE_ORIGINAL_TIMESTAMP"
```
sfc-gh-dprzybysz, line 31: PLease extend readme with description of the new paramter

converter.go
```diff
@@ -354,6 +354,74 @@ func decimalToBigFloat(num decimal128.Num, scale int64) *big.Float {
...
+func ArrowSnowflakeTimestampToTime(rec arrow.Record, rb *ArrowBatch, colIdx int, recIdx int) *time.Time {
+	scale := int(rb.scd.RowSet.RowType[colIdx].Scale)
+	dbType := strings.ToUpper(rb.scd.RowSet.RowType[colIdx].Type)
```
sfc-gh-dheyman-1, line 360: perhaps we can encapsulate scale, dbtype and loc in a new structure? that way we wouldn't have to pass the whole ArrowBatch to the method? Just an idea

converter.go
```diff
@@ -354,6 +354,74 @@ func decimalToBigFloat(num decimal128.Num, scale int64) *big.Float {
...
+	loc *time.Location) *time.Time {
+
+	if column.IsNull(recIdx) {
+		return nil
+	}
+	var ret time.Time
+	switch sfType {
+	case timestampNtzType:
+		if column.DataType().ID() == arrow.STRUCT {
+			structData := column.(*array.Struct)
```
sfc-gh-dheyman-1, line 378: don't we need to check if the struct has all the fields to avoid errors when extracting the actual values?

converter.go
```diff
@@ -354,6 +354,74 @@ func decimalToBigFloat(num decimal128.Num, scale int64) *big.Float {
...
+	dbType := strings.ToUpper(rb.scd.RowSet.RowType[colIdx].Type)
+	return arrowSnowflakeTimestampToTime(rec.Column(colIdx), getSnowflakeType(dbType), scale, recIdx, rb.loc)
+}
+
+func arrowSnowflakeTimestampToTime(
+	column arrow.Array,
+	sfType snowflakeType,
+	scale int,
+	recIdx int,
+	loc *time.Location) *time.Time {
```
sfc-gh-dheyman-1, line 369: do we need to return a pointer here?

converter.go
```diff
@@ -354,6 +354,74 @@ func decimalToBigFloat(num decimal128.Num, scale int64) *big.Float {
...
+		}
+		return &ret
+	case timestampTzType:
+		structData := column.(*array.Struct)
+		if structData.NumField() == 2 {
+			value := structData.Field(0).(*array.Int64).Int64Values()
+			timezone := structData.Field(1).(*array.Int32).Int32Values()
+			loc := Location(int(timezone[recIdx]) - 1440)
+			epoch := value[recIdx] / int64(math.Pow10(scale))
+			fraction := value[recIdx] % int64(math.Pow10(scale)) * int64(math.Pow10(9-scale))
```
sfc-gh-dheyman-1, line 411: perhaps we can extract this to a method `extractFraction(value, scale)` because there are some repetitions of `int64(math.Pow10(scale)) * int64(math.Pow10(9-scale))`. Same with epoch

converter_test.go
sfc-gh-dstempniak: yes, but since it's not a large block of code, I think defining it in each test is more readable

util.go
sfc-gh-dstempniak: There is no such section in the readme file. I'll later create the ticket for the doc team to update https://pkg.go.dev/github.com/snowflakedb/gosnowflake#WithArrowBatches

converter.go
sfc-gh-dstempniak: I don't think it is a good option. In such case a driver user will have to extract data from ArrowBatch by himself and create such object to pass. Moreover scd and loc aren't public.

util.go
sfc-gh-dszmolka: @sfc-gh-dstempniak for Go, we don't use (yet) docs.snowflake.com so please feel free to directly update `doc.go` so it can be generated automatically at pkg.go.dev

converter.go
sfc-gh-dstempniak: I'm not sure if this is needed. We know how the struct looks like for each timestamp type. Additional checking will be done for each row (can be time consuming) and will result in an error being returned. This would be SF issue that will come to us anyway.

converter.go
sfc-gh-dstempniak: Timestamp can be null. Is there other way to return null values?

util.go
sfc-gh-dstempniak: The parameter (or rather the function enabling the functionality) is described in `util.go`. So as I understand it won't go to [README](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#section-readme) section, but it will go to [Documentation](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#pkg-functions) section like other similar functions. @sfc-gh-dszmolka  Am I right? 
If so, I think it is enough.

PR #919, opened 2023-10-02
SNOW-878073 Refactor retry policy to support HTTP 503 & 429

sfc-gh-dheyman-1:
### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing


retry.go
```diff
@@ -298,27 +299,19 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
...
-				// or
-				// abort connection if raise4XX flag is enabled and the range of HTTP status code are 4XX.
-				// This is currently used for Snowflake login. The caller must generate an error object based on HTTP status.
-				break
-			}
 			logger.WithContext(r.ctx).Warningf(
 				"failed http connection. HTTP Status: %v. retrying...\n", res.StatusCode)
-			res.Body.Close()
 		}
+		res.Body.Close()
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: Because otherwise this will cause resource leak (tcp connection won't be reused)

retry.go
```diff
@@ -366,36 +359,13 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
...
-		}
-		errString := urlError.Err.Error()
-		if runtime.GOOS == "darwin" && strings.HasPrefix(errString, "x509:") && strings.HasSuffix(errString, "certificate is expired") {
-			// Certificate is expired
-			return true, err
-		}
-
+func (r *retryHTTP) isRetryableError(req *http.Request, res *http.Response, err error) (bool, error) {
+	if res == nil {
+		return false, err
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: this happens when there is a timeout on client side (can't get any response from a server). IMHO no, since this is caused most likely by the clients connectivity issues. I would rather fail fast here.

retry.go
```diff
@@ -5,13 +5,12 @@ package gosnowflake
 import (
 	"bytes"
 	"context"
-	"crypto/x509"
 	"fmt"
+	"golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: Removed the library and added contains() method

retry.go
```diff
@@ -20,6 +19,15 @@ import (
 
 var random *rand.Rand
 
+var endpointsEligibleForRetry = []string{
+	loginRequestPath,
+	queryRequestPath,
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: You are right, removed query endpoint.

retry.go
```diff
@@ -374,36 +381,19 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
...
+	if res == nil || req == nil {
+		return false, err
 	}
-	return false, err
+	isRetryableURL := contains(endpointsEligibleForRetry, req.URL.Path)
+	return isRetryableURL && isRetryableStatus(res.StatusCode), err
+}
+
+func isRetryableStatus(statusCode int) bool {
+	return (statusCode >= 500 && statusCode < 600) || contains(statusCodesEligibleForRetry, statusCode)
```
sfc-gh-pfus, line 398: Maybe better name would be `clientErrorsStatusCodesEligibleForRetry`? I thought that this is the full list at first.

retry_test.go
```diff
@@ -341,18 +342,18 @@ func TestRetryQuerySuccessWithTimeout(t *testing.T) {
...
+		success:    false,
 	}
 	urlPtr, err := url.Parse("https://fakeaccountretryfail.snowflakecomputing.com:443/queries/v1/query-request?" + requestIDKey)
 	if err != nil {
 		t.Fatal("failed to parse the test URL")
 	}
 	_, err = newRetryHTTP(context.Background(),
 		client,
-		emptyRequest, urlPtr, make(map[string]string), 60*time.Second, defaultTimeProvider, nil).doPost().setBody([]byte{0}).execute()
+		emptyRequest, urlPtr, make(map[string]string), 30*time.Second, defaultTimeProvider, nil).doPost().setBody([]byte{0}).execute()
```
sfc-gh-pfus, line 356: So this test duration is at least 30s? Maybe we can decrease it even more?

retry_test.go
```diff
@@ -483,3 +484,85 @@ func TestLoginRetry429(t *testing.T) {
 		t.Fatalf("no retry counter should be attached: %v", retryCountKey)
 	}
 }
+
+type retryableTc struct {
```
sfc-gh-pfus, line 488: Can we inline this struct?

retry.go
```diff
@@ -5,23 +5,51 @@ package gosnowflake
...
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 )
 
+const (
+	// MaxRetryCount specifies maximum number of subsequent retries
+	MaxRetryCount = 7
```
sfc-gh-pfus, line 22: Do we need it publicly available?

retry.go
sfc-gh-dheyman-1: done.

retry_test.go
sfc-gh-dheyman-1: Decreased to 15s. There have to be at least 2 retries.

retry_test.go
sfc-gh-dheyman-1: done.

retry.go
sfc-gh-dheyman-1: in fact no, we dont. Changed scope.

sfc-gh-pfus (approved):

PR #926, opened 2023-10-12
Fix TestConcurrentReadOnParams cleaning up the connection too early

sfc-gh-pfus:
### Description
This test failed randomly when concurrent threads were run. If no row was read, the connection was returned to the pool too fast, before chunk decoding (in another thread) was finished. Currently, as we read from the rows, this object is not returned to the pool too early.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,6 +48,7 @@ jobs:
               env:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
+                GORACE: history_size=7
               run: ./ci/test.sh
             - name: Upload coverage to Codecov
               uses: codecov/codecov-action@v3
--- a/connection_test.go
+++ b/connection_test.go
@@ -476,19 +476,19 @@ func TestExecWithServerSideError(t *testing.T) {
 }
 
 func TestConcurrentReadOnParams(t *testing.T) {
-	t.Skip("Fails randomly")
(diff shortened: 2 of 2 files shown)


connection_test.go
```diff
@@ -499,13 +497,19 @@ func TestConcurrentReadOnParams(t *testing.T) {
...
+					t.Error(err)
+				}
 				_ = rows.Close()
 			}
 			wg.Done()
 		}()
 	}
 	wg.Wait()
-	defer db.Close()
+	db.Close()
```
sfc-gh-dheyman-1, line 512: do we need that since we deferred db.Close in line 483?

.github/workflows/build-test.yml
```diff
@@ -48,6 +48,7 @@ jobs:
               env:
                 PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
+                GORACE: history_size=7
```
sfc-gh-dheyman-1, line 51: how is this used?

connection_test.go
sfc-gh-pfus: Good catch!

.github/workflows/build-test.yml
sfc-gh-pfus: When race occurs, only one stack was gathered (with latest access). In that way, we have access also to previous threads accessing the same space where race occurs.

sfc-gh-dheyman-1 (approved):

PR #933, opened 2023-10-13
SNOW-856228 better assertions in easy logging

sfc-gh-knozderko:
### Description

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/assert_test.go
+++ b/assert_test.go
@@ -48,7 +48,7 @@ func assertHasPrefix(t *testing.T, actual string, expectedPrefix string, descrip
 
 func joinDescriptions(descriptions ...string) string {
 	desc := ""
-	if descriptions != nil && len(descriptions) > 0 {
+	if len(descriptions) > 0 {
 		desc = " while checking: " + strings.Join(descriptions, " ")
 	}
 	return desc
--- a/assert_test.go
+++ b/assert_test.go
@@ -16,29 +16,8 @@ func assertNilF(t *testing.T, actual any, descriptions ...string) {
 	}
 }
 
-func assertNilE(t *testing.T, actual any, descriptions ...string) {
-	errMsg := validateNil(actual, descriptions...)
-	if errMsg != "" {
-		t.Error(errMsg)
-	}
-}
-
 func assertNotNilF(t *testing.T, actual any, descriptions ...string) {
 	errMsg := validateNotNil(actual, descriptions...)
-	if errMsg != "" {
-		t.Fatal(errMsg)
-	}
-}
-
-func assertNotNilE(t *testing.T, actual any, descriptions ...string) {
-	errMsg := validateNotNil(actual, descriptions...)
-	if errMsg != "" {
-		t.Error(errMsg)
-	}
-}
-
-func assertEqualF(t *testing.T, actual string, expected string, descriptions ...string) {
-	errMsg := validateEqualStrings(actual, expected, descriptions...)
 	if errMsg != "" {
 		t.Fatal(errMsg)
 	}
@@ -51,24 +30,10 @@ func assertEqualE(t *testing.T, actual string, expected string, descriptions ...
 	}
 }
 
-func assertStringContainsF(t *testing.T, actual string, expectedToContain string, descriptions ...string) {
-	errMsg := validateStringContains(actual, expectedToContain, descriptions...)
-	if errMsg != "" {
-		t.Fatal(errMsg)
-	}
-}
-
 func assertStringContainsE(t *testing.T, actual string, expectedToContain string, descriptions ...string) {
 	errMsg := validateStringContains(actual, expectedToContain, descriptions...)
 	if errMsg != "" {
 		t.Error(errMsg)
-	}
-}
-
-func assertHasPrefixF(t *testing.T, actual string, expectedPrefix string, descriptions ...string) {
-	errMsg := validateHasPrefix(actual, expectedPrefix, descriptions...)
-	if errMsg != "" {
-		t.Fatal(errMsg)
 	}
 }
 
--- a/assert_test.go
+++ b/assert_test.go
@@ -9,18 +9,18 @@ import (
 	"testing"
 )
 
-func assertNil(t *testing.T, actual interface{}, descriptions ...string) {
+func assertNil(t *testing.T, actual any, descriptions ...string) {
 	if !isNil(actual) {
 		desc := joinDescriptions(descriptions...)
-		errMsg := fmt.Sprintf("expected \"%s\" to be nil but was not%s", actual, desc)
+		errMsg := fmt.Sprintf("expected \"%s\" to be nil but was not. %s", actual, desc)
 		t.Fatal(errMsg)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


assert_test.go
```diff
@@ -0,0 +1,55 @@
+package gosnowflake
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func assertNil(t *testing.T, actual interface{}, descriptions ...string) {
```
sfc-gh-pfus, line 9: Can we replace `interface{}` with `any`?

assert_test.go
```diff
@@ -0,0 +1,55 @@
...
+	if actual != nil {
+		desc := joinDescriptions(descriptions...)
+		errMsg := fmt.Sprintf("expected \"%s\" to be nil but was not%s", actual, desc)
+		t.Fatal(errMsg)
```
sfc-gh-pfus, line 13: Do we need fatal here? Maybe error is better?

assert_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func assertNil(t *testing.T, actual interface{}, descriptions ...string) {
+	if actual != nil {
+		desc := joinDescriptions(descriptions...)
+		errMsg := fmt.Sprintf("expected \"%s\" to be nil but was not%s", actual, desc)
```
sfc-gh-pfus, line 12: Maybe it would be better to just finish the message "but was not. %s", and just return `strings.Join(descriptions)`?

assert_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func assertEqual(t *testing.T, actual string, expected string, descriptions ...string) {
+	if expected != actual {
+		desc := joinDescriptions(descriptions...)
+		errMsg := fmt.Sprintf("expected \"%s\" to be equal to \"%s\" but was not%s", actual, expected, desc)
+		t.Fatal(errMsg)
+	}
+}
+
+func assertContains(t *testing.T, actual string, expectedToContain string, descriptions ...string) {
```
sfc-gh-pfus, line 33: Maybe it would be better to rename it to `stringContains`? There is no function overloading in Go.

assert_test.go
sfc-gh-knozderko: done

assert_test.go
sfc-gh-knozderko: I renamed it to assertStringContains, because I think that it would be good to have assert in the method's name

assert_test.go
sfc-gh-knozderko: I prefer fatal, because most likely I would like to break the test execution.
Common test scenario is to execute "when" operation and in the first order to check if there is error. Only if there is no error then we assert on the result. If there was an error and we got nil as the response and if we try assert something on nil as if it was a regular object we might get panic errors.
So we could have 2 versions of these methods: one for fatal, one for error, but for now for simplicity I did only the fatal one.

assert_test.go
sfc-gh-pfus: Yes, very good, it was only a brachylogy.

assert_test.go
sfc-gh-pfus: Ok, so going with go way, what I suggest is naming functions like `assertNilF` and when it is needed we can add `assertNilE`? This way every attempt may be considered individually.

assert_test.go
sfc-gh-knozderko: I changed it, but kept method joinDescriptions extracted as a separate method

assert_test.go
sfc-gh-knozderko: I added both versions for each assert

assert_test.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func assertNilF(t *testing.T, actual any, descriptions ...string) {
+	errMsg := validateNil(actual, descriptions...)
+	if errMsg != "" {
```
sfc-gh-pfus, line 14: I can see a lot of duplications here. Maybe we can write it as:
```
errMsg := ...
fatalOnEmpty(t, errMsg)?
```

assert_test.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func assertNotNilE(t *testing.T, actual any, descriptions ...string) {
+	errMsg := validateNotNil(actual, descriptions...)
+	if errMsg != "" {
+		t.Error(errMsg)
+	}
+}
+
+func assertEqualF(t *testing.T, actual string, expected string, descriptions ...string) {
```
sfc-gh-pfus: Can we have `actual` and `expected` as `any`?

sfc-gh-knozderko:
a

assert_test.go
```diff
@@ -0,0 +1,132 @@
...
+	errMsg := validateEqualStrings(actual, expected, descriptions...)
```
sfc-gh-pfus: If above comment seems yes, there should be just `validateEqual` here, right?

assert_test.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func assertHasPrefixE(t *testing.T, actual string, expectedPrefix string, descriptions ...string) {
+	errMsg := validateHasPrefix(actual, expectedPrefix, descriptions...)
+	if errMsg != "" {
+		t.Error(errMsg)
+	}
+}
+
+func validateNil(actual any, descriptions ...string) string {
```
sfc-gh-pfus, line 47: I wonder if it is better to stick them closer to the callee. Maybe we can group asserts and their validation function together?

client_configuration_test.go
```diff
@@ -150,26 +139,18 @@ func TestParseConfigurationFails(t *testing.T) {
...
-			}
+			assertHasPrefixE(t, errMessage, expectedPrefix, "error message")
+			assertStringContainsE(t, errMessage, tc.expectedErrorMessageToContain, "error message")
 		})
 	}
 }
 
 func createFile(t *testing.T, fileName string, fileContents string, directory string) string {
 	fullFileName := path.Join(directory, fileName)
 	err := os.WriteFile(fullFileName, []byte(fileContents), 0644)
```
sfc-gh-pfus, line 153: Shouldn't we defer closing file here?

assert_test.go
sfc-gh-knozderko: changed, even in 1 liner

assert_test.go
sfc-gh-knozderko: done

assert_test.go
sfc-gh-knozderko: changed

assert_test.go
sfc-gh-knozderko: we agreed to leave it as is

client_configuration_test.go
sfc-gh-knozderko: I think that no, because we create a directory for the files as `t.TempDir()` and we can read in its description:

"The directory is automatically removed by Cleanup when the test and all its subtests complete."

sfc-gh-knozderko:
b

PR #945, opened 2023-10-25
SNOW-949229: support overwrite option in PUT

sfc-gh-pfus:
### Description
Supports overwrite in PUT commands. Before, overwrite retrieved from Snowflake was not taken into account.
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/712

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -92,7 +92,6 @@ type SnowflakeFileTransferOptions struct {
 	compressSourceFromStream bool
 
 	/* PUT */
-	DisablePutOverwrite     bool
 	putCallback             *snowflakeProgressPercentage
 	putAzureCallback        *snowflakeProgressPercentage
 	putCallbackOutputStream *io.Writer
@@ -268,7 +267,7 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
 	if sfa.data.Parallel != 0 {
 		sfa.parallel = sfa.data.Parallel
 	}
-	sfa.overwrite = !sfa.options.DisablePutOverwrite
+	sfa.overwrite = sfa.data.Overwrite
 	sfa.stageLocationType = cloudType(strings.ToUpper(sfa.data.StageInfo.LocationType))
 	sfa.stageInfo = &sfa.data.StageInfo
 	sfa.presignedURLs = make([]string, 0)
--- a/put_get_test.go
+++ b/put_get_test.go
@@ -288,6 +288,9 @@ func TestPutWithAutoCompressFalse(t *testing.T) {
 }
 
 func TestPutOverwrite(t *testing.T) {
+	if runningOnGCP() {
+		t.Skip("Overwriting is default as long as presigned URLs are enabled")
+	}
 	tmpDir := t.TempDir()
 	testData := filepath.Join(tmpDir, "data.txt")
 	f, err := os.Create(testData)
@@ -298,9 +301,6 @@ func TestPutOverwrite(t *testing.T) {
 	f.Close()
 
 	runDBTest(t, func(dbt *DBTest) {
-		if _, err = dbt.exec("use role sysadmin"); err != nil {
-			t.Skip("snowflake admin account not accessible")
-		}
 		dbt.mustExec("rm @~/test_put_overwrite")
 
 		f, _ = os.Open(testData)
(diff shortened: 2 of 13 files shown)


file_transfer_agent.go
```diff
@@ -92,7 +92,6 @@ type SnowflakeFileTransferOptions struct {
 	compressSourceFromStream bool
```
sfc-gh-pfus, line 92: Can you extend doc.go for this `WithFileTransferOptions`?

put_get_test.go
```diff
@@ -321,13 +321,18 @@ func TestPutOverwrite(t *testing.T) {
 			t.Fatalf("expected UPLOADED, got %v", s6)
 		}
 
+		rows = dbt.mustQuery("ls @~/test_put_overwrite")
+		defer rows.Close()
+		if rows.Next() {
```
sfc-gh-pfus, line 326: What if no rows are returned? `uploadTime` is incorrectly updated (with previous `s3` value), and test fails with strange to debug message. Maybe it would be better to assert on `rows.Next`?

put_get_test.go
```diff
@@ -341,11 +346,23 @@ func TestPutOverwrite(t *testing.T) {
 			t.Fatalf("expected SKIPPED, got %v", s6)
 		}
 
+		rows = dbt.mustQuery("ls @~/test_put_overwrite")
+		defer rows.Close()
+		if rows.Next() {
```
sfc-gh-pfus, line 351: Same here. If `rows.Next` returns false, we never update `s3`, so we compare the same value and test passes even it it should not.

s3_storage_client.go
```diff
@@ -80,11 +80,7 @@ func (util *snowflakeS3Client) getFileHeader(meta *fileMetadata, filename string
 		if errors.As(err, &ae) {
 			if ae.ErrorCode() == notFound {
 				meta.resStatus = notFoundFile
-				return &fileHeader{
-					digest:             "",
-					contentLength:      0,
-					encryptionMetadata: nil,
-				}, nil
+				return nil, fmt.Errorf("could not find file")
```
sfc-gh-pfus, line 83: I'm not sure if this is ok. If this is a correct situation, we shouldn't return error. Why previous solution was bad?

storage_client.go
```diff
@@ -88,7 +88,12 @@ func (rsu *remoteStorageUtil) uploadOneFile(meta *fileMetadata) error {
 	for retry := 0; retry < maxRetry; retry++ {
 		if !meta.overwrite {
 			header, err := utilClass.getFileHeader(meta, meta.dstFileName)
-			if err != nil {
+			if meta.resStatus == notFoundFile {
+				err := utilClass.uploadFile(dataFile, meta, encryptMeta, maxConcurrency, meta.options.MultiPartThreshold)
```
sfc-gh-pfus, line 92: Do we have a test covering this situation?

storage_client.go
```diff
@@ -88,7 +88,12 @@ func (rsu *remoteStorageUtil) uploadOneFile(meta *fileMetadata) error {
 	for retry := 0; retry < maxRetry; retry++ {
 		if !meta.overwrite {
 			header, err := utilClass.getFileHeader(meta, meta.dstFileName)
-			if err != nil {
+			if meta.resStatus == notFoundFile {
```
sfc-gh-pfus, line 91: According to my previous comment I see it like this:
```
if err != nil {
  // handle err
} else if meta.resStatus == notFoundFile {
  // handle upload
} else {
  // what else can be here? maybe at least log?
}

**Review by sfc-gh-dheyman-1 — APPROVED** [review]

File: storage_client.go [hunk 7]
```diff
@@ -90,6 +90,11 @@ func (rsu *remoteStorageUtil) uploadOneFile(meta *fileMetadata) error {
 			header, err := utilClass.getFileHeader(meta, meta.dstFileName)
 			if err != nil {
 				return err
+			} else if meta.resStatus == notFoundFile {
+				err := utilClass.uploadFile(dataFile, meta, encryptMeta, maxConcurrency, meta.options.MultiPartThreshold)
+				if err != nil {
+					logger.Warnf("Error uploading %v. err: %v", dataFile, err)
+				}
 			}
 			if header != nil && meta.resStatus == uploaded {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Yes, I fixed it just before your comment ;)

s3_storage_client.go
```diff
@@ -80,11 +80,7 @@ func (util *snowflakeS3Client) getFileHeader(meta *fileMetadata, filename string
 		if errors.As(err, &ae) {
 			if ae.ErrorCode() == notFound {
 				meta.resStatus = notFoundFile
-				return &fileHeader{
-					digest:             "",
-					contentLength:      0,
-					encryptionMetadata: nil,
-				}, nil
+				return nil, errors.New("could not find file")
```
sfc-gh-dstempniak, line 83: why `errors.New` not  `fmt.Errorf` like in other cases ?

sfc-gh-ext-simba-jl: I've changed it back to fmt.Errorf to match with other cases

put_get_test.go
```diff
@@ -321,51 +321,67 @@ func TestPutOverwrite(t *testing.T) {
...
 		f.Close()
-		if rows.Next() {
-			if err = rows.Scan(&s0, &s1, &s2, &s3, &s4, &s5, &s6, &s7); err != nil {
-				t.Fatal(err)
-			}
+		assertTrueF(t, rows.Next(), "expected new rows")
+		if err = rows.Scan(&s0, &s1, &s2, &s3, &s4, &s5, &s6, &s7); err != nil {
+			t.Fatal(err)
 		}
 		if s6 != skipped.String() {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Yes, it works.

PR #946, opened 2023-10-25
SNOW-856228 enable easy logging

sfc-gh-knozderko:
### Description
Enable easy logging

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -735,6 +735,10 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
 		queryContextCache:   (&queryContextCache{}).init(),
 		currentTimeProvider: defaultTimeProvider,
 	}
+	err := initEasyLogging(config.ClientConfigFile)
+	if err != nil {
+		return nil, err
+	}
 	var st http.RoundTripper = SnowflakeTransport
 	if sc.cfg.Transporter == nil {
 		if sc.cfg.InsecureMode {
--- a/easy_logging.go
+++ b/easy_logging.go
@@ -1,0 +1,138 @@
+package gosnowflake
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path"
+	"strings"
+)
+
+type initTrials struct {
+	everTriedToInitialize bool
+	initParameter         string
+	configureCounter      int
+}
+
+var easyLoggingInitTrials = initTrials{
+	everTriedToInitialize: false,
+	initParameter:         "",
+	configureCounter:      0,
+}
+
+func (i *initTrials) setInitTrial(initParameter string) {
+	i.everTriedToInitialize = true
+	i.initParameter = initParameter
+}
+
+func (i *initTrials) increaseReconfigureCounter() {
+	i.configureCounter++
+}
+
+func (i *initTrials) resetInitTrials() {
+	i.everTriedToInitialize = false
+	i.initParameter = ""
+	i.configureCounter = 0
+}
+
+func initEasyLogging(configFilePathFromConnectionString string) error {
+	if !allowedToInitialize(configFilePathFromConnectionString) {
+		return nil
+	}
+	config, err := getClientConfig(configFilePathFromConnectionString)
+	if err != nil {
+		return err
+	}
+	if config == nil {
+		easyLoggingInitTrials.setInitTrial(configFilePathFromConnectionString)
+		return nil
+	}
+	var logLevel string
+	logLevel, err = getLogLevel(config.Common.LogLevel)
+	if err != nil {
+		return err
+	}
+	var logPath string
+	logPath, err = getLogPath(config.Common.LogPath)
+	if err != nil {
+		return err
+	}
+	reconfigureEasyLogging(logLevel, logPath)
+	easyLoggingInitTrials.setInitTrial(configFilePathFromConnectionString)
+	easyLoggingInitTrials.increaseReconfigureCounter()
+	return nil
+}
+
+func reconfigureEasyLogging(logLevel string, logPath string) error {
+	err := logger.SetLogLevel(logLevel)
+	if err != nil {
+		return err
+	}
+	var output io.Writer
+	output, err = createLogWriter(logPath)
+	if err != nil {
+		return err
+	}
+	logger.SetOutput(output)
+	return nil
+}
+
+func createLogWriter(logPath string) (io.Writer, error) {
+	if strings.EqualFold(logPath, "STDOUT") {
+		return os.Stdout, nil
+	}
+	logFileName := path.Join(logPath, "snowflake.log")
+	file, err := os.OpenFile(logFileName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return nil, err
+	}
+	return io.MultiWriter(file, os.Stdout), nil
+}
+
+func allowedToInitialize(configFilePathFromConnectionString string) bool {
+	triedToInitializeWithoutConfigFile := easyLoggingInitTrials.everTriedToInitialize && easyLoggingInitTrials.initParameter == ""
+	isAllowedToInitialize := !easyLoggingInitTrials.everTriedToInitialize || (triedToInitializeWithoutConfigFile && configFilePathFromConnectionString != "")
+	if !isAllowedToInitialize && easyLoggingInitTrials.initParameter != configFilePathFromConnectionString {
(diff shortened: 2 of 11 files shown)


connection.go
```diff
@@ -735,6 +735,10 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
 		queryContextCache:   (&queryContextCache{}).init(),
 		currentTimeProvider: defaultTimeProvider,
 	}
+	err := initEasyLogging(config.ClientConfigFile)
```
sfc-gh-pfus, line 738: Do we officially call it easy logging? Or is it just client config(uration)?

connection.go
```diff
@@ -735,6 +735,10 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
+	if err != nil {
+		return nil, err
```
sfc-gh-pfus, line 740: Should we fail here? Maybe warn is enough? This is far from being a critical feature.

connection.go
sfc-gh-pfus, line 738: Should we init it on every new connection? Maybe on initializing driver is good enough?

easy_logging.go
```diff
@@ -0,0 +1,138 @@
...
+func (i *initTrials) setInitTrial(initParameter string) {
+	i.everTriedToInitialize = true
+	i.initParameter = initParameter
+}
+
+func (i *initTrials) increaseReconfigureCounter() {
+	i.configureCounter++
+}
+
+func (i *initTrials) resetInitTrials() {
```
sfc-gh-pfus, line 32: Just `reset`?

easy_logging.go
```diff
@@ -0,0 +1,138 @@
...
+	file, err := os.OpenFile(logFileName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return nil, err
+	}
+	return io.MultiWriter(file, os.Stdout), nil
+}
+
+func allowedToInitialize(configFilePathFromConnectionString string) bool {
+	triedToInitializeWithoutConfigFile := easyLoggingInitTrials.everTriedToInitialize && easyLoggingInitTrials.initParameter == ""
+	isAllowedToInitialize := !easyLoggingInitTrials.everTriedToInitialize || (triedToInitializeWithoutConfigFile && configFilePathFromConnectionString != "")
```
sfc-gh-pfus, line 94: Is it `ever` or `never`?

easy_logging.go
```diff
@@ -0,0 +1,138 @@
...
+	logFileName := path.Join(logPath, "snowflake.log")
+	file, err := os.OpenFile(logFileName, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0666)
+	if err != nil {
+		return nil, err
+	}
+	return io.MultiWriter(file, os.Stdout), nil
+}
+
+func allowedToInitialize(configFilePathFromConnectionString string) bool {
+	triedToInitializeWithoutConfigFile := easyLoggingInitTrials.everTriedToInitialize && easyLoggingInitTrials.initParameter == ""
```
sfc-gh-pfus, line 93: I'm not sure what this `initParameter` means, can you explain it a bit?

easy_logging.go
```diff
@@ -0,0 +1,138 @@
...
+		logger.Warn("LogLevel in client config not found. Using default value: OFF")
+		return levelOff, nil
+	}
+	return toLogLevel(logLevel)
+}
+
+func getLogPath(logPath string) (string, error) {
+	logPathOrDefault := logPath
+	if logPath == "" {
+		logger.Warn("LogPath in client config not found. Using temporary directory as a default value")
```
sfc-gh-pfus, line 112: Maybe we can print the tempDir here?

easy_logging.go
```diff
@@ -0,0 +1,138 @@
...
+	if !exists {
+		err = os.MkdirAll(pathWithGoSubdir, 0755)
+		if err != nil {
+			return "", err
+		}
+	}
+	return pathWithGoSubdir, nil
+}
+
+func existsDir(dirPath string) (bool, error) {
```
sfc-gh-pfus, line 129: Is it rather `dirExists`?

sfc-gh-pfus, line 129: Same function exists in tests, maybe we can remove it?

connection.go
sfc-gh-knozderko: We do logging reconfiguration so I named it accordingly. It is also named in such a way in other drivers.
Recently I also added more descriptive error information using "easy logging" phrase, because it relates to a particular feature.

connection.go
sfc-gh-knozderko: we do fail in other drivers as well. As I remember it was supposed to work this way.

connection.go
sfc-gh-knozderko: we don't init on every new connection, for most of the cases we will immediately return from this method without taking any further actions. We really init only for the first time and for the first time when client config file is given explicite. It is not enough to do it on driver's initalization because it can be reconfigured based on value given in config parameter.

easy_logging.go
sfc-gh-knozderko: ever

easy_logging.go
sfc-gh-knozderko: changed

easy_logging.go
sfc-gh-knozderko: it is Config.ClientConfigFile so the parameter the user provided in the config or dsn

easy_logging.go
sfc-gh-knozderko: done

easy_logging.go
sfc-gh-knozderko: not sure what is better, but I changed as you suggest

sfc-gh-knozderko: in what file? I don't see it.

connection.go
sfc-gh-knozderko: I changed the exception to be more about "client config" than about "easy logging"

easy_logging.go
sfc-gh-pfus: In `client_configuration.go` :P The implementation for dir and file is the same, right?

easy_logging.go
sfc-gh-pfus: Please, rename it to be more descriptive.

sfc-gh-knozderko: renamed to: clientConfigFileInput

easy_logging.go
sfc-gh-knozderko: they are different. In dirExists we additionally check if it is a dictionary.
If it is a file, but not a dictionary we return false.
If we used existsFile function we would get true.

easy_logging.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"errors"
+	"io"
+	"os"
+	"path"
+	"strings"
+)
+
+type initTrials struct {
```
sfc-gh-pfus, line 11: Is the name same in all drivers? I'm wondering if `attempts` is a better word?

easy_logging_test.go
```diff
@@ -0,0 +1,184 @@
...
+	assertNilF(t, err, "open config error")
+	err = openWithClientConfigFile(t, configFilePath)
+	assertNilF(t, err, "open config error")
+	err = openWithClientConfigFile(t, "/another-config.json")
+	assertNilF(t, err, "open config error")
+
+	assertEqualE(t, toClientConfigLevel(logger.GetLogLevel()), logLevel, "error log level check")
+	assertEqualE(t, easyLoggingInitTrials.configureCounter, 1)
+
+	cleanUp()
```
sfc-gh-pfus, line 35: It would be better, to use `t.Cleanup` before the assertions? If an assertion fails, cleanup is never executed.

easy_logging_test.go
```diff
@@ -0,0 +1,184 @@
...
+	case "WARNING":
+		return levelWarn
+	case levelOff, levelError, levelWarn, levelInfo, levelDebug, levelTrace:
+		return logLevelUpperCase
+	default:
+		return ""
+	}
+}
+
+func notEmpty(val string) bool {
```
sfc-gh-pfus, line 158: Can we inline it?

sfc-gh-pfus (approved):

easy_logging_test.go
sfc-gh-knozderko: improved

PR #948, opened 2023-10-26
SNOW-878073 Add MaxRetryCount as configurable parameter

sfc-gh-dheyman-1:
### Description
Added MaxRetryCount as configurable parameter. It specifies how many times requests can be retried.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -336,6 +336,7 @@ func TestParseDSN(t *testing.T) {
 				ClientTimeout:             defaultClientTimeout,
 				JWTClientTimeout:          defaultJWTClientTimeout,
 				IncludeRetryReason:        ConfigBoolTrue,
+				MaxRetryCount:             defaultMaxRetryCount,
 			},
 			ocspMode: ocspModeFailOpen,
 		},
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -340,6 +340,22 @@ func TestParseDSN(t *testing.T) {
 			ocspMode: ocspModeFailOpen,
 		},
 		{
+			dsn: "u:p@a?database=d&maxRetryCount=20",
+			config: &Config{
+				Account: "a", User: "u", Password: "p",
+				Protocol: "https", Host: "a.snowflakecomputing.com", Port: 443,
+				Database: "d", Schema: "",
+				ExternalBrowserTimeout:    defaultExternalBrowserTimeout,
+				OCSPFailOpen:              OCSPFailOpenTrue,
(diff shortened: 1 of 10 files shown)


dsn_test.go
```diff
@@ -339,6 +339,22 @@ func TestParseDSN(t *testing.T) {
 			},
 			ocspMode: ocspModeFailOpen,
 		},
+		{
+			dsn: "u:p@a?database=d&maxRetryCount=20",
```
sfc-gh-pfus, line 343: Can we verify that the default value is correctly set, if not configured?

retry_test.go
```diff
@@ -353,7 +353,37 @@ func TestRetryQueryFail(t *testing.T) {
 	}
 	_, err = newRetryHTTP(context.Background(),
 		client,
-		emptyRequest, urlPtr, make(map[string]string), 15*time.Second, defaultTimeProvider, nil).doPost().setBody([]byte{0}).execute()
+		emptyRequest, urlPtr, make(map[string]string), 15*time.Second, 100, defaultTimeProvider, nil).doPost().setBody([]byte{0}).execute()
+	if err == nil {
```
sfc-gh-pfus, line 357: We have an assertion library for such checks :)

retry_test.go
```diff
@@ -353,7 +353,37 @@ func TestRetryQueryFail(t *testing.T) {
...
+	values, err = url.ParseQuery(urlPtr.RawQuery)
+	if err != nil {
+		t.Fatalf("failed to parse the URL: %v", err)
+	}
+	retry, err := strconv.Atoi(values.Get(retryCountKey))
+	if err != nil {
+		t.Fatalf("failed to get retry counter: %v", err)
+	}
+	if retry < 2 {
+		t.Fatalf("not enough retry counter: %v", retry)
```
sfc-gh-pfus, line 370: Is it more like `not enough retires attempted`?

dsn_test.go
sfc-gh-dheyman-1: Added a check in other test.

retry_test.go
sfc-gh-dheyman-1: Changed nil checks in this file to use `assertNilf/assertNotNilf`

retry_test.go
sfc-gh-dheyman-1: Changed the name to be more applicable.

sfc-gh-pfus (approved):

PR #967, opened 2023-11-13
SNOW-968719: Remember queryId for failed queries

sfc-gh-pbulawa:
### Description
Remember queryId for failed queries.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -5,6 +5,7 @@ package gosnowflake
 import (
 	"context"
 	"database/sql/driver"
+	"errors"
 	"fmt"
 )
 
@@ -35,7 +36,10 @@ func (stmt *snowflakeStmt) ExecContext(ctx context.Context, args []driver.NamedV
 	logger.WithContext(stmt.sc.ctx).Infoln("Stmt.ExecContext")
 	result, err := stmt.sc.ExecContext(ctx, stmt.query, args)
 	if err != nil {
-		stmt.lastQueryID = err.(*SnowflakeError).QueryID
+		var snowflakeError *SnowflakeError
+		if errors.As(err, &snowflakeError) {
+			stmt.lastQueryID = snowflakeError.QueryID
+		}
 		return nil, err
 	}
 	r, ok := result.(SnowflakeResult)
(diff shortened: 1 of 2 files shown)


statement.go
```diff
@@ -63,6 +72,10 @@ func (stmt *snowflakeStmt) Exec(args []driver.Value) (driver.Result, error) {
 	logger.WithContext(stmt.sc.ctx).Infoln("Stmt.Exec")
 	result, err := stmt.sc.Exec(stmt.query, args)
 	if err != nil {
+		var snowflakeError *SnowflakeError
```
sfc-gh-pfus, line 75: Can we extract some function here?

statement_test.go
```diff
@@ -39,6 +39,109 @@ func openConn(t *testing.T) *sql.Conn {
 	return conn
 }
 
+func TestSetFailedQueryId(t *testing.T) {
+	ctx := context.Background()
+	conn := openConn(t)
```
sfc-gh-pfus, line 44: Can we use `runDBTest` here?

statement.go
```diff
@@ -1,10 +1,11 @@
-// Copyright (c) 2017-2022 Snowflake Computing Inc. All rights reserved.
+// Copyright (c) 2017-2023 Snowflake Computing Inc. All rights reserved.
```
sfc-gh-pfus, line 1: Can you confirm that for regular (not statement-based) queries it works?

statement.go
sfc-gh-pbulawa: Done

statement_test.go
sfc-gh-pbulawa: Done

statement.go
```diff
@@ -90,3 +95,10 @@ func (stmt *snowflakeStmt) Query(args []driver.Value) (driver.Rows, error) {
 func (stmt *snowflakeStmt) GetQueryID() string {
 	return stmt.lastQueryID
 }
+
+func (stmt *snowflakeStmt) setQueryIDFromError(err error) {
+	var snowflakeError *SnowflakeError
+	if errors.As(err, &snowflakeError) {
+		stmt.lastQueryID = snowflakeError.QueryID
+	}
```
sfc-gh-pfus, line 103: Maybe we can add log in else here, if casting fails?

sfc-gh-pfus (approved):

statement.go
sfc-gh-pbulawa: I think that this is not necessary as there is at least 1 valid case returning regular `Error` instead of `SnowflakeError` - when the provided binding value is not supported.

PR #978, opened 2023-11-29
SNOW-978164: Fix stmt.Exec for DML

sfc-gh-pbulawa:
### Description
Enable the usage of stmt.Exec/ExecContext for DML without returning the exception when used.

### Checklist
- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/monitoring.go
+++ b/monitoring.go
@@ -136,7 +136,7 @@ func (sc *snowflakeConn) checkQueryStatus(
 	if tok, _, _ := sc.rest.TokenAccessor.GetTokens(); tok != "" {
 		headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, tok)
 	}
-	resultPath := fmt.Sprintf("/monitoring/queries/%s", qid)
+	resultPath := fmt.Sprintf("%s/%s", monitoringQueriesPath, qid)
 	url := sc.rest.getFullURL(resultPath, &param)
 
 	res, err := sc.rest.FuncGet(ctx, sc.rest, url, headers, sc.rest.RequestTimeout)
--- a/restful.go
+++ b/restful.go
@@ -37,6 +37,7 @@ const (
 	tokenRequestPath         = "/session/token-request"
 	abortRequestPath         = "/queries/v1/abort-request"
 	authenticatorRequestPath = "/session/authenticator-request"
+	monitoringQueriesPath    = "/monitoring/queries"
 	sessionRequestPath       = "/session"
 	heartBeatPath            = "/session/heartbeat"
 )
--- a/retry.go
+++ b/retry.go
@@ -20,18 +20,17 @@ import (
 type waitAlgo struct {
 	mutex  *sync.Mutex // required for *rand.Rand usage
 	random *rand.Rand
+	base   time.Duration // base wait time
+	cap    time.Duration // maximum wait time
 }
 
 var random *rand.Rand
 var defaultWaitAlgo *waitAlgo
 
-var endpointsEligibleForRetry = []string{
+var authEndpoints = []string{
 	loginRequestPath,
 	tokenRequestPath,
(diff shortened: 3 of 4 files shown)


connection.go
```diff
@@ -335,7 +335,11 @@ func (sc *snowflakeConn) ExecContext(
 		return sc.handleMultiExec(ctx, data.Data)
 	}
 	logger.Debug("DDL")
-	return driver.ResultNoRows, nil
+	return &snowflakeResult{
```
sfc-gh-pfus, line 338: I think we shouldn't change this, as this is a BCR. Client code may depend on returning `ResultNoRows`.

connection_util.go
```diff
@@ -194,10 +194,13 @@ func getResultType(ctx context.Context) resultType {
 
 // isDml returns true if the statement type code is in the range of DML.
 func isDml(v int64) bool {
-	return statementTypeIDDml <= v && v <= statementTypeIDMultiTableInsert
+	return (statementTypeIDDml <= v && v <= statementTypeIDMultiTableInsert) || v == statementTypeIDSelect
```
sfc-gh-pfus, line 197: Can you explain the reason of this change? I believe that SELECT is not a DML query.

sfc-gh-pbulawa: It seems that you are right as in Snowflake it is not considered as one as per docs, however I have found a multiple mentions in the internet that it is considered a limited DML, i.e.:
- https://docs.oracle.com/cd/B13789_01/server.101/b10759/statements_1001.htm#:~:text=The%20SELECT%20statement%20is%20a,the%20results%20of%20the%20query.
- https://en.wikipedia.org/wiki/Data_manipulation_language

connection.go
sfc-gh-pbulawa: `ResultNoRows` is a struct implementing `driver.Result` methods (`LastInsertId()`, `RowsAffected()`). These methods throw an error when used. `SnowflakeResult` is also an implementation of `driver.Result` with additional features thus I do not believe that it is a BCR.

rows_test.go
```diff
@@ -74,22 +74,6 @@ func TestRowsClose(t *testing.T) {
 	})
 }
 
-func TestResultNoRows(t *testing.T) {
```
sfc-gh-pfus, line 77: Why this was removed?

statement_test.go
```diff
@@ -40,6 +40,68 @@ func openConn(t *testing.T) *sql.Conn {
...
+			_, err := tc.f(dbt)
+			if err != nil {
+				t.Error(err)
+			}
+		}
+	})
+}
+
+func TestDDLExec(t *testing.T) {
+	query := "CREATE OR REPLACE TABLE TestDDLExec (num NUMBER)"
```
sfc-gh-pfus, line 75: Can we defer dropping this table?

connection_util.go
```diff
@@ -197,7 +197,14 @@ func isDml(v int64) bool {
 	return statementTypeIDDml <= v && v <= statementTypeIDMultiTableInsert
 }
 
+func isSelect(v int64) bool {
+	return v == statementTypeIDSelect
```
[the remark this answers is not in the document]
sfc-gh-pbulawa: Added additional check, thanks!

statement_test.go
sfc-gh-pbulawa: Done

statement_test.go
```diff
@@ -40,6 +40,72 @@ func openConn(t *testing.T) *sql.Conn {
 	return conn
 }
 
+func TestExecStmt(t *testing.T) {
+	dqlQuery := "SELECT 1"
```
sfc-gh-pfus, line 44: Can we have a dml and multistatement tested here? Does it make sense?

sfc-gh-pbulawa: Done

PR #997, opened 2023-12-10
SNOW-983911 Add support for querying the vector data type

sfc-gh-rpanchapakesan:
### Description
Added support for parsing vectors from the serialized result set, so queries that return vectors can now be made. To do so, I also had to update `execResponseRowType` to add two new fields that are now being sent back, matching the [change](https://github.com/snowflakedb/snowflake-connector-python/pull/1806) I recently made to the Python connector.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -91,10 +91,10 @@ func goTypeToSnowflake(v driver.Value, tsmode snowflakeType) snowflakeType {
 }
 
 // snowflakeTypeToGo translates Snowflake data type to Go data type.
-func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
-	switch dbtype {
+func snowflakeTypeToGo(rowType execResponseRowType) reflect.Type {
+	switch getSnowflakeType(rowType.Type) {
 	case fixedType:
-		if scale == 0 {
+		if rowType.Scale == 0 {
 			return reflect.TypeOf(int64(0))
 		}
 		return reflect.TypeOf(float64(0))
@@ -108,8 +108,22 @@ func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
 		return reflect.TypeOf([]byte{})
 	case booleanType:
 		return reflect.TypeOf(true)
+	case vectorType:
+		if len(rowType.Fields) != 1 {
+			logger.Errorf("invalid response row type for vector: %+v", rowType)
+			return reflect.TypeOf("")
+		}
+		switch getSnowflakeType(rowType.Fields[0].Type) {
+		case fixedType:
+			return reflect.TypeOf([]int32{})
+		case realType:
+			return reflect.TypeOf([]float32{})
+		default:
+			logger.Errorf("invalid element type for vector: %+v", rowType.Fields[0].Type)
+			return reflect.TypeOf("")
+		}
 	}
-	logger.Errorf("unsupported dbtype is specified. %v", dbtype)
+	logger.Errorf("unsupported dbtype is specified. %v", rowType.Type)
 	return reflect.TypeOf("")
 }
 
@@ -608,6 +622,25 @@ func arrowToValue(
 			if ts != nil {
 				destcol[i] = *ts
 			}
+		}
+		return err
+	case vectorType:
+		vectorData := srcValue.(*array.FixedSizeList)
+		datatype := vectorData.DataType().(*arrow.FixedSizeListType)
+		dim := int(datatype.Len())
+		switch datatype.Elem().ID() {
+		case arrow.INT32:
+			values := vectorData.ListValues().(*array.Int32).Int32Values()
+			for i := 0; i < vectorData.Len(); i++ {
+				destcol[i] = values[i*dim : (i+1)*dim]
+			}
+		case arrow.FLOAT32:
+			values := vectorData.ListValues().(*array.Float32).Float32Values()
+			for i := 0; i < vectorData.Len(); i++ {
+				destcol[i] = values[i*dim : (i+1)*dim]
+			}
+		default:
+			return fmt.Errorf("unsupported element type %q for a vector", datatype.Elem().String())
 		}
 		return err
 	}
--- a/converter.go
+++ b/converter.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/big"
@@ -110,7 +111,7 @@ func snowflakeTypeToGo(rowType execResponseRowType) reflect.Type {
 		return reflect.TypeOf(true)
 	case vectorType:
 		if len(rowType.Fields) != 1 {
-			logger.Errorf("invalid response row type for vector: %+v", rowType)
+			logger.Errorf("invalid result metadata fields for vector: length=%d", len(rowType.Fields))
 			return reflect.TypeOf("")
 		}
 		switch getSnowflakeType(rowType.Fields[0].Type) {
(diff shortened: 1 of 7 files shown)


driver_test.go
```diff
@@ -1263,6 +1264,48 @@ func testArray(t *testing.T, json bool) {
 	})
 }
 
+func TestVector(t *testing.T) {
```
sfc-gh-pfus, line 1267: Do we need to enable some parameter for this to work?

driver_test.go
```diff
@@ -1263,6 +1264,48 @@ func testArray(t *testing.T, json bool) {
...
+		var gotInt []int32
+		if err := rowsInt.Scan(Vector(&gotInt)); err != nil {
+			t.Error(err)
+		}
+		wantInt := []int32{1, 2, 3}
+		if !reflect.DeepEqual(gotInt, wantInt) {
+			t.Errorf("incorrect vector deserialized: got %v, want %v", gotInt, wantInt)
+		}
+
+		rowsFloat := dbt.mustQuery(`select [1.1,2.2,3,4,5]::vector(float, 5)`)
```
sfc-gh-pfus, line 1292: Can we split it into separate tests for int and float?

driver_test.go
```diff
@@ -1263,6 +1264,48 @@ func testArray(t *testing.T, json bool) {
...
+func TestVector(t *testing.T) {
+	testVector(t, false)
+}
+
+func testVector(t *testing.T, json bool) {
+	runDBTest(t, func(dbt *DBTest) {
+		if json {
+			dbt.mustExec(forceJSON)
+		}
+		// dbt.mustExec(`create temporary table test_vector (i vector(int,3), f vector(float,3))`)
```
sfc-gh-pfus, line 1276: Can we remove this query?

driver_test.go
```diff
@@ -1263,6 +1264,48 @@ func testArray(t *testing.T, json bool) {
 	})
 }
 
+func TestVector(t *testing.T) {
+	testVector(t, false)
```
sfc-gh-pfus, line 1268: We are lacking the test for JSON, right?

driver_test.go
```diff
@@ -1263,6 +1264,48 @@ func testArray(t *testing.T, json bool) {
...
+}
+
+func testVector(t *testing.T, json bool) {
```
sfc-gh-pfus, line 1271: Instead of such approach, we use `t.Run` - please see some examples in our tests.

converter.go
```diff
@@ -91,10 +91,10 @@ func goTypeToSnowflake(v driver.Value, tsmode snowflakeType) snowflakeType {
 }
 
 // snowflakeTypeToGo translates Snowflake data type to Go data type.
-func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
-	switch dbtype {
+func snowflakeTypeToGo(rowType execResponseRowType) reflect.Type {
```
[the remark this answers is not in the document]
sfc-gh-rpanchapakesan: I could parse the response type in the caller of this function and pass down some type of options struct with the vector metadata. However, since this function is only called in one place right now (and in the context of a request), it felt cleaner to have the function take in `execResponseRowType`.

converter.go
```diff
@@ -108,8 +108,22 @@ func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
...
 	case booleanType:
 		return reflect.TypeOf(true)
+	case vectorType:
+		if len(rowType.Fields) != 1 {
+			logger.Errorf("invalid response row type for vector: %+v", rowType)
+			return reflect.TypeOf("")
+		}
+		switch getSnowflakeType(rowType.Fields[0].Type) {
+		case fixedType:
+			return reflect.TypeOf([]int32{})
```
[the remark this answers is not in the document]
sfc-gh-rpanchapakesan: Yes, we only currently support 32 bit values.

driver_test.go
sfc-gh-rpanchapakesan: Added tests for JSON to `old_driver_test.go`.

driver_test.go
sfc-gh-rpanchapakesan: Oh ok, I was following the example of `testArray` and `testVariant` which call `runDBTest` without `t.Run`. I've converted this test into a table test that uses `t.Run`.

driver_test.go
sfc-gh-rpanchapakesan: I wasn't sure which environment these tests were running oi. However, it looks like the CI failures are due to the parameter, so I've added a statement to enable it for this test.

converter.go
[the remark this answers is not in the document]
sfc-gh-pfus: I think that separations of layers (API/mapping) is good :) I would prefer to have API parsing closer to API functions, and this function preserve with just types at it was before.

converter.go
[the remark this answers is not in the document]
sfc-gh-pfus: Can we mention V1 in the commit message (so for now - in PR name only)?

converter.go
```diff
@@ -610,6 +646,33 @@ func arrowToValue(
 			}
 		}
 		return err
+	case vectorType:
+		vectorData := srcValue.(*array.FixedSizeList)
+		datatype := vectorData.DataType().(*arrow.FixedSizeListType)
```
sfc-gh-pfus, line 651: Does this work only for arrow? What should happen if a client uses `ALTER SESSION SET GO_QUERY_RESULT_FORMAT = JSON`?

driver_test.go
```diff
@@ -1263,6 +1264,102 @@ func testArray(t *testing.T, json bool) {
 	})
 }
 
+func TestVectorInt(t *testing.T) {
+	testVectorInt(t, false)
```
sfc-gh-pfus, line 1268: How about testing with JSON?

driver_test.go
```diff
@@ -1263,6 +1264,102 @@ func testArray(t *testing.T, json bool) {
...
+			want:  []int32{-5},
+		},
+		{
+			msg:   "3 element vector",
+			query: `select [1,2,-3]::vector(int, 3)`,
+			want:  []int32{1, 2, -3},
+		},
+	}
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.mustExec("alter session set enable_vector_data_type='Enable'")
```
sfc-gh-pfus, line 1289: We use uppercase for parameters, can you change it?

driver_test.go
sfc-gh-pfus: I believe that adding it close to the Arrow tests is better for now.

sfc-gh-pfus (changes requested):

datatype.go
```diff
@@ -131,6 +135,40 @@ func dataTypeMode(v driver.Value) (tsmode snowflakeType, err error) {
 	return tsmode, nil
 }
 
+type vectorElements interface {
```
sfc-gh-pfus, line 138: I'm wondering about one thing. In the future we are expected to implement structured types, which contains arrays. Array is a very similar structure to vector, just more generic I'd say. Maybe we can prepare the code here to use these arrays instead of native backend type as vector?

driver_test.go
```diff
@@ -1263,6 +1264,102 @@ func testArray(t *testing.T, json bool) {
...
+
+func testVectorInt(t *testing.T, json bool) {
+	tests := []struct {
+		msg   string
+		query string
+		want  []int32
+	}{
+		{
+			msg:   "1 element vector",
+			query: `select [-5]::vector(int, 1)`,
```
sfc-gh-pfus, line 1279: Can we also have a test when the response contains empty/null value?

PR #999, opened 2023-12-11
SNOW-988972 adding an example demonstrating how to insert data into VARIANT and OBJECT datatype columns with bind variables

sfc-gh-dszmolka:
### Description
SNOW-988972 adding an example demonstrating how to insert data into VARIANT and OBJECT datatype columns with bind variables

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/cmd/variant/insertvariantobject.go
+++ b/cmd/variant/insertvariantobject.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"flag"
@@ -42,6 +43,13 @@ func main() {
 	}
 	defer db.Close()
 
+	ctx := context.Background()
+	conn, err := db.Conn(ctx)
+	if err != nil {
+		log.Fatalf("Failed to acquire connection. err: %v", err)
+	}
+	defer conn.Close()
+
 	tablename := "insert_variant_object_" + strconv.FormatInt(time.Now().UnixNano(), 10)
 	param := map[string]string{"key": "value"}
 	jsonStr, err := json.Marshal(param)
@@ -62,19 +70,19 @@ func main() {
 	dropQuery := "DROP TABLE " + tablename
 
 	fmt.Printf("Creating table: %v\n", createTableQuery)
-	_, err = db.Exec(createTableQuery)
+	_, err = conn.ExecContext(ctx, createTableQuery)
 	if err != nil {
 		log.Fatalf("failed to run the query. %v, err: %v", createTableQuery, err)
 	}
 	defer func() {
 		fmt.Printf("Dropping the table: %v\n", dropQuery)
-		_, err = db.Exec(dropQuery)
+		_, err = conn.ExecContext(ctx, dropQuery)
 		if err != nil {
 			log.Fatalf("failed to run the query. %v, err: %v", dropQuery, err)
 		}
 	}()
 	fmt.Printf("Inserting VARIANT and OBJECT data into table: %v\n", insertQuery)
-	_, err = db.Exec(insertQuery,
+	_, err = conn.ExecContext(ctx, insertQuery,
 		string(jsonStr),
 		string(jsonStr),
 	)
@@ -82,13 +90,13 @@ func main() {
 		log.Fatalf("failed to run the query. %v, err: %v", insertQuery, err)
 	}
 	fmt.Printf("Now for another approach: %v\n", insertOnlyObject)
-	_, err = db.Exec(insertOnlyObject)
+	_, err = conn.ExecContext(ctx, insertOnlyObject)
 	if err != nil {
 		log.Fatalf("failed to run the query. %v, err: %v", insertOnlyObject, err)
 	}
 
 	fmt.Printf("Querying the table into which we just inserted the data: %v\n", selectQuery)
-	rows, err := db.Query(selectQuery)
+	rows, err := conn.QueryContext(ctx, selectQuery)
 	if err != nil {
 		log.Fatalf("failed to run the query. %v, err: %v", selectQuery, err)
 	}


cmd/insertvariantobject/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../../gosnowflake.mak
+CMD_TARGET=insertvariantobject
```
sfc-gh-pfus, line 2: Can we make it more generic - just naming it `variant`?

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	db, err := sql.Open("snowflake", dsn)
+	if err != nil {
+		log.Fatalf("failed to connect. %v, err: %v", dsn, err)
+	}
+	defer db.Close()
+
+	param := map[string]string{"key": "value"}
+	jsonStr, _ := json.Marshal(param)
```
sfc-gh-pfus, line 44: Why do you omit error here?

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+
+	createTableQuery := "CREATE OR REPLACE TABLE insertvariantobject (c1 VARIANT, c2 OBJECT);"
```
sfc-gh-pfus, line 46: Maybe underscores between words in table name?

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// https://docs.snowflake.com/en/sql-reference/functions/parse_json
+	// can do with TO_VARIANT(PARSE_JSON(..)) as well, but PARSE_JSON already produces VARIANT
+	insertQuery := "INSERT INTO insertvariantobject (c1, c2) SELECT PARSE_JSON(?), TO_OBJECT(PARSE_JSON(?));"
+	// https://docs.snowflake.com/en/sql-reference/data-types-semistructured#object
+	insertOnlyObject := "INSERT INTO insertvariantobject (c2) SELECT OBJECT_CONSTRUCT('name', 'Jones'::VARIANT,'age',  42::VARIANT);"
```
sfc-gh-pfus, line 52: Why do we need `::VARIANT` suffix?
Btw, space is missing before `age`.

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+	defer db.Close()
+
+	param := map[string]string{"key": "value"}
+	jsonStr, _ := json.Marshal(param)
+
+	createTableQuery := "CREATE OR REPLACE TABLE insertvariantobject (c1 VARIANT, c2 OBJECT);"
+
+	// https://docs.snowflake.com/en/sql-reference/functions/parse_json
+	// can do with TO_VARIANT(PARSE_JSON(..)) as well, but PARSE_JSON already produces VARIANT
+	insertQuery := "INSERT INTO insertvariantobject (c1, c2) SELECT PARSE_JSON(?), TO_OBJECT(PARSE_JSON(?));"
```
sfc-gh-pfus, line 50: Usually, when sending single query, we don't add semicolon at the end.

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+	// https://docs.snowflake.com/en/sql-reference/data-types-semistructured#object
+	insertOnlyObject := "INSERT INTO insertvariantobject (c2) SELECT OBJECT_CONSTRUCT('name', 'Jones'::VARIANT,'age',  42::VARIANT);"
+
+	selectQuery := "SELECT c1, c2 FROM insertvariantobject;"
+
+	fmt.Printf("Running CREATE OR REPLACE TABLE\n")
```
sfc-gh-pfus, line 56: You can use `println` instead of manually adding new line character.

Also, I believe that printing the actual query may be helpful.

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+	db.Exec(createTableQuery)
```
sfc-gh-pfus, line 57: Why omitting error? When table creation fails, we shouldn't proceed.

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+
+	fmt.Printf("Inserting VARIANT and OBJECT data into table\n")
+	_, err = db.Exec(insertQuery,
+		string(jsonStr),
+		string(jsonStr),
+	)
+	if err != nil {
+		log.Fatalf("failed to run the query. %v, err: %v", insertQuery, err)
+	}
+	_, err = db.Exec(insertOnlyObject)
```
sfc-gh-pfus, line 67: No print before?

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-pfus, line 57: If you call many operations on `db` directly, it causes authentication every time. It would be better, if we had connection created here.

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,91 @@
...
+                log.Fatalf("failed to run the query. %v, err: %v", selectQuery, err)
+        }
+        defer rows.Close()
+	var c1, c2 any
+	for rows.Next() {
+		err := rows.Scan(&c1, &c2)
+		if err != nil {
+			log.Fatalf("failed to get result. err: %v", err)
+		}
+		fmt.Printf("%v, %v\n", c1, c2)
```
sfc-gh-pfus, line 84: What do you think about printing `%T` (actual type) here?

cmd/insertvariantobject/Makefile
sfc-gh-dszmolka: i'm following naming convention of all the rest of the stuff under `cmd` but of course can change it

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: fix

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: ok

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: its coming from the SQL where you have a strong habit of adding them, but removed

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: this is literally from the Snowflake documentation and i intended to use the same example here.
thanks for pointing out the space missing ,fixed

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: thanks for the pointer! 
added the queries too

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: right, fixed

sfc-gh-dszmolka: again this is a very useful pointer, thank you !
however the example is only here because i thought it would be better as a code, instead of the documentation. by no means is it representative of any production-ready approach, it is really intended to show how you can insert VARIANT and OBJECT into your table with variable binding.

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: good catch, missed it, thank you ! fixed

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: can do that but what would be the reason for that ?

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// https://docs.snowflake.com/en/sql-reference/functions/parse_json
+	// can do with TO_VARIANT(PARSE_JSON(..)) as well, but PARSE_JSON already produces VARIANT
+	insertQuery := "INSERT INTO insert_variant_object (c1, c2) SELECT PARSE_JSON(?), TO_OBJECT(PARSE_JSON(?))"
+	// https://docs.snowflake.com/en/sql-reference/data-types-semistructured#object
+	insertOnlyObject := "INSERT INTO insert_variant_object (c2) SELECT OBJECT_CONSTRUCT('name', 'Jones'::VARIANT, 'age',  42::VARIANT)"
+
+	selectQuery := "SELECT c1, c2 FROM insert_variant_object"
+
+	fmt.Printf("Running CREATE OR REPLACE TABLE: %v\n", createTableQuery)
```
sfc-gh-pfus, line 59: Maybe just `Creating table: `?

And maybe it would be good to defer dropping table?

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-pfus: I'm curious what is the real type here? String? Some json type or sth?

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-pfus: On the other hand, I like keeping code clean everywhere it is possible. Otherwise, someone may copy/use the code incorrectly, just by looking at the example.

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-pfus: I'm also thinking, maybe just to be on the safe side - it would be better to skip `OR REPLACE` clause, if we defer dropping. Let's consider the scenarios:
1. When table does not exist - it is created and dropped by us.
2. When table exists - it is better to fail on table creation, because we don't want to replace table, that is not managed by us.
I know chances are very low, but...

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: ok, added type

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: you're right - although i do not expect anyone to actually run this code, but you got a very valid point
so i chose a different approach, suffixing the table name with `UnixNano` , hopefully reduces our chances very much :)  into having (and then, potentially dropping) a table with the exact name in the exact same database and exact same schema ..

also implemented the cleanup (drop table) which i also missed

cmd/insertvariantobject/insertvariantobject.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			log.Fatalf("failed to get result. err: %v", err)
+		}
+		fmt.Printf("%v (type: %T), %v (type: %T)\n", c1, c1, c2, c2)
+	}
+	if rows.Err() != nil {
+		fmt.Printf("ERROR: %v\n", rows.Err())
+		return
+	}
+	fmt.Printf("Dropping the table: %v\n", dropQuery)
```
sfc-gh-pfus, line 102: I think that we should defer it and move to line 69:

```
defer func() {
// dropping here
}()
```

cmd/insertvariantobject/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../../gosnowflake.mak
+CMD_TARGET=variant
```
sfc-gh-pfus, line 2: Can you rename the folder as well?

sfc-gh-dszmolka: renamed the folder

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: got it, thank you, added this way

sfc-gh-pfus (approved):

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: eventually, the drop has been moved to a function and that invoked from `defer`

cmd/insertvariantobject/insertvariantobject.go
sfc-gh-dszmolka: has been rewritten eventually to use `db.Conn`

PR #1012, opened 2023-12-21
SNOW-994698 cover corner cases where an unexpected non APIerror occurs during HeadObject

sfc-gh-dszmolka:
to see if it fixes 1008

### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/s3_storage_client.go
+++ b/s3_storage_client.go
@@ -89,6 +89,9 @@ func (util *snowflakeS3Client) getFileHeader(meta *fileMetadata, filename string
 			meta.lastError = err
 			return nil, fmt.Errorf("error while retrieving header")
 		}
+		meta.resStatus = errStatus
+		meta.lastError = err
+		return nil, fmt.Errorf("unexpected error while retrieving header: %v", err)
 	}
 
 	meta.resStatus = uploaded
--- a/s3_storage_client_test.go
+++ b/s3_storage_client_test.go
@@ -294,9 +294,10 @@ func TestGetHeaderNonApiError(t *testing.T) {
 			return nil, othErr
 		}),
 	}
+
 	header, err := new(snowflakeS3Client).getFileHeader(&meta, "file.txt")
 	assertEqualF(t, header, nil, fmt.Sprintf("expected nil header, got: %v", header))
-	assertNotNilF(t, err, fmt.Sprintf("expected err to not be nil"))
+	assertNotNilF(t, err, "expected err to not be nil")
 	assertEqualF(t, meta.resStatus, errStatus, fmt.Sprintf("expected %v result status for non-APIerror, got: %v", errStatus, meta.resStatus))
 }
 
--- a/s3_storage_client_test.go
+++ b/s3_storage_client_test.go
@@ -296,9 +296,9 @@ func TestGetHeaderNonApiError(t *testing.T) {
 	}
 
 	header, err := new(snowflakeS3Client).getFileHeader(&meta, "file.txt")
-	assertEqualF(t, header, nil, fmt.Sprintf("expected nil header, got: %v", header))
-	assertNotNilF(t, err, "expected err to not be nil")
-	assertEqualF(t, meta.resStatus, errStatus, fmt.Sprintf("expected %v result status for non-APIerror, got: %v", errStatus, meta.resStatus))
+	assertNilE(t, header, fmt.Sprintf("expected header to be nil, actual: %v", header))
+	assertNotNilE(t, err, "expected err to not be nil")
+	assertEqualE(t, meta.resStatus, errStatus, fmt.Sprintf("expected %v result status for non-APIerror, got: %v", errStatus, meta.resStatus))
 }
 
 func TestGetHeaderNotFoundError(t *testing.T) {
--- a/s3_storage_client_test.go
+++ b/s3_storage_client_test.go
@@ -286,12 +286,11 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
 }
 
 func TestGetHeaderNonApiError(t *testing.T) {
-	othErr := errors.New("something went wrong here")
 	meta := fileMetadata{
 		client:    s3.New(s3.Options{}),
 		stageInfo: &execResponseStageInfo{Location: ""},
 		mockHeader: mockHeaderAPI(func(ctx context.Context, params *s3.HeadObjectInput, optFns ...func(*s3.Options)) (*s3.HeadObjectOutput, error) {
-			return nil, othErr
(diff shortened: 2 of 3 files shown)


s3_storage_client.go
```diff
@@ -87,8 +87,11 @@ func (util *snowflakeS3Client) getFileHeader(meta *fileMetadata, filename string
 			}
 			meta.resStatus = errStatus
 			meta.lastError = err
-			return nil, fmt.Errorf("error while retrieving header")
+			return nil, fmt.Errorf("s3 api error while retrieving header")
```
sfc-gh-pfus, line 90: I think it would be better to leave it as is, some people may depend on this message.

s3_storage_client.go
```diff
@@ -87,8 +87,11 @@ func (util *snowflakeS3Client) getFileHeader(meta *fileMetadata, filename string
...
 		}
+		meta.resStatus = errStatus
+		meta.lastError = err
+		return nil, fmt.Errorf("unexpected error while retrieving header")
```
sfc-gh-pfus, line 94: I have a strong belief, that go devs depend more on this error than on the `lastError` field. So IMHO it would be better to include `err` in return value. We can either return just `return nil, err` or add a message like `return nil, fmt.Errorf("unexpected error while retrieving header. %v", err)` (not sure about syntax, please check it :) )

s3_storage_client_test.go
```diff
@@ -285,6 +285,23 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
 	}
 }
 
+func TestGetHeaderNonApiError(t *testing.T) {
+	othErr := errors.New("Something went wrong here.")
```
sfc-gh-pfus, line 289: Can we inline this var?

s3_storage_client_test.go
```diff
@@ -285,6 +285,23 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
...
+		client:    s3.New(s3.Options{}),
+		stageInfo: &execResponseStageInfo{Location: ""},
+		mockHeader: mockHeaderAPI(func(ctx context.Context, params *s3.HeadObjectInput, optFns ...func(*s3.Options)) (*s3.HeadObjectOutput, error) {
+			return nil, othErr
+		}),
+	}
+	if header, err := new(snowflakeS3Client).getFileHeader(&meta, "file.txt"); header != nil || err == nil {
+		t.Fatalf("expected null header, got: %v", header)
+	}
+	if meta.resStatus != errStatus {
```
sfc-gh-pfus, line 300: We have some assertion library - please, check `assertEqualF`.

s3_storage_client_test.go
```diff
@@ -285,6 +285,23 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
...
+	othErr := errors.New("Something went wrong here.")
+	meta := fileMetadata{
+		client:    s3.New(s3.Options{}),
+		stageInfo: &execResponseStageInfo{Location: ""},
+		mockHeader: mockHeaderAPI(func(ctx context.Context, params *s3.HeadObjectInput, optFns ...func(*s3.Options)) (*s3.HeadObjectOutput, error) {
+			return nil, othErr
+		}),
+	}
+	if header, err := new(snowflakeS3Client).getFileHeader(&meta, "file.txt"); header != nil || err == nil {
+		t.Fatalf("expected null header, got: %v", header)
```
sfc-gh-pfus, line 298: The message explains only a case when header is nil (btw nil, not null in go), the message may be confusing when header is nil, but the error is nil as well. Maybe we can split it into two assertions?

s3_storage_client.go
sfc-gh-dszmolka: ok

s3_storage_client.go
sfc-gh-dszmolka: good point, changed

s3_storage_client_test.go
sfc-gh-dszmolka: ok

s3_storage_client_test.go
sfc-gh-dszmolka: this is part of the original driver code, but will check it

s3_storage_client_test.go
sfc-gh-dszmolka: again this is part of how the driver was originally written, but still valid points, so ii'll check

s3_storage_client_test.go
sfc-gh-dszmolka: ok

s3_storage_client_test.go
sfc-gh-dszmolka: ok

s3_storage_client_test.go
```diff
@@ -285,6 +285,22 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
...
+	meta := fileMetadata{
+		client:    s3.New(s3.Options{}),
+		stageInfo: &execResponseStageInfo{Location: ""},
+		mockHeader: mockHeaderAPI(func(ctx context.Context, params *s3.HeadObjectInput, optFns ...func(*s3.Options)) (*s3.HeadObjectOutput, error) {
+			return nil, othErr
+		}),
+	}
+
+	header, err := new(snowflakeS3Client).getFileHeader(&meta, "file.txt")
+	assertEqualF(t, header, nil, fmt.Sprintf("expected nil header, got: %v", header))
```
sfc-gh-pfus, line 299: We have `assertNilF(E)` to check nullability.

s3_storage_client_test.go
```diff
@@ -285,6 +285,22 @@ func TestGetHeaderUnexpectedError(t *testing.T) {
...
+	assertNotNilF(t, err, "expected err to not be nil")
+	assertEqualF(t, meta.resStatus, errStatus, fmt.Sprintf("expected %v result status for non-APIerror, got: %v", errStatus, meta.resStatus))
```
sfc-gh-pfus, line 301: After you rewrote it to use our asserts I have one suggestion - we can use functions ending with `E` (error) instead of `F` (fatal). The difference is that when `E` fails, the test is marked as failed, but it is continuing to check remaining assertions.

sfc-gh-dszmolka: noted, thank you ! replaced with the Error version

s3_storage_client_test.go
sfc-gh-dszmolka: got it; used assertNilE and also added assertNotNilE (was only F available)

sfc-gh-pfus (approved):

PR #1018, opened 2023-12-22
SNOW-996678 Use snowflakeTransport for all clouds

sfc-gh-pfus:
### Description
Use the same transport configuration for all stages.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/gcs_storage_client.go
+++ b/gcs_storage_client.go
@@ -69,8 +69,7 @@ func (util *snowflakeGcsClient) getFileHeader(meta *fileMetadata, filename strin
 		for k, v := range gcsHeaders {
 			req.Header.Add(k, v)
 		}
-		var client gcsAPI
-		client = &http.Client{}
+		client := newGcsClient()
 		// for testing only
 		if meta.mockGcsClient != nil {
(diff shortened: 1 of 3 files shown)


gcs_storage_client.go
```diff
@@ -390,3 +387,9 @@ func (util *snowflakeGcsClient) generateFileURL(stageLocation string, filename s
 func (util *snowflakeGcsClient) isTokenExpired(resp *http.Response) bool {
 	return resp.StatusCode == 401
 }
+
+func newGcsClient() gcsAPI {
```
sfc-gh-dheyman-1, line 391: is this something GCS specific? Because it seems like the client is cloud provider agnostic, we are creating the same instance in s3_storage_client. Couldn't this method be renamed and  shared?

sfc-gh-pfus: Actually this is not that simple. Here we need to return `gcsAPI`, which is implemented by standard `http.Client`. We can't change it, because it would change the `client` variable type.

sfc-gh-dheyman-1 (approved):

PR #1021, opened 2024-01-08
SNOW-942595: Retry the fetch result request for response code queryInProgressAsyncCode when using WithFetchResultByID

sfc-gh-ext-simba-jl:
### Description
issue: [#705](https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/705)
using:
```
ctx = gosnowflake.WithFetchResultByID(ctx, queryId)
rows, err := s.db.QueryContext(ctx, "")
```
will return no error and no result after about 45 seconds.

This PR applies the same change as [PR #824](https://github.com/snowflakedb/gosnowflake/pull/824) to check for the status `queryInProgressAsyncCode` in `rowsForRunningQuery` in `monitoring.go`

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/async.go
+++ b/async.go
@@ -145,6 +145,8 @@ func getQueryResultWithRetriesForAsyncMode(
 	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
 
 	for {
+		logger.WithContext(ctx).Debugf("Retry count for get query result request in async mode: %v", retry)
+
 		resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
 		if err != nil {
 			logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
@@ -167,12 +169,11 @@ func getQueryResultWithRetriesForAsyncMode(
 			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
 			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
 			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
-			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
+			logger.WithContext(ctx).Infof("Query execution still in progress. Response code: %v, message: %v Sleep for %v ms", respd.Code, respd.Message, sleepTime)
 			time.Sleep(sleepTime)
 
 			if retry < len(retryPattern)-1 {
 				retry++
-				logger.Debugf("retry: %v", retry)
 			}
 		}
 	}
--- a/async.go
+++ b/async.go
@@ -143,6 +143,7 @@ func getQueryResultWithRetriesForAsyncMode(
 	var respd *execResponse
 	retry := 0
 	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
+	retryPatternIndex := 0
 
 	for {
 		logger.WithContext(ctx).Debugf("Retry count for get query result request in async mode: %v", retry)
@@ -168,12 +169,13 @@ func getQueryResultWithRetriesForAsyncMode(
 		} else {
 			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
 			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
-			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
+			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retryPatternIndex])
 			logger.WithContext(ctx).Infof("Query execution still in progress. Response code: %v, message: %v Sleep for %v ms", respd.Code, respd.Message, sleepTime)
 			time.Sleep(sleepTime)
+			retry++
 
-			if retry < len(retryPattern)-1 {
-				retry++
+			if retryPatternIndex < len(retryPattern)-1 {
+				retryPatternIndex++
 			}
 		}
 	}
--- a/async.go
+++ b/async.go
@@ -64,45 +64,12 @@ func (sr *snowflakeRestful) getAsync(
 	token, _, _ := sr.TokenAccessor.GetTokens()
 	headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, token)
 
-	var err error
-	var respd execResponse
-	retry := 0
-	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
-
-	for {
-		resp, err := sr.FuncGet(ctx, sr, URL, headers, timeout)
-		if err != nil {
-			logger.WithContext(ctx).Errorf("failed to get response. err: %v", err)
-			sfError.Message = err.Error()
-			errChannel <- sfError
-			return err
-		}
-		defer resp.Body.Close()
-
-		respd = execResponse{} // reset the response
-		err = json.NewDecoder(resp.Body).Decode(&respd)
-		if err != nil {
-			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
-			sfError.Message = err.Error()
-			errChannel <- sfError
-			return err
-		}
-		if respd.Code != queryInProgressAsyncCode {
-			// If the query takes longer than 45 seconds to complete the results are not returned.
-			// If the query is still in progress after 45 seconds, retry the request to the /results endpoint.
-			// For all other scenarios continue processing results response
-			break
-		} else {
-			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
-			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
-			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
-			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
-			time.Sleep(sleepTime)
-		}
-		if retry < len(retryPattern)-1 {
-			retry++
-		}
-
+	respd, err := getQueryResult(ctx, sr, URL, headers, timeout)
+	if err != nil {
+		logger.WithContext(ctx).Errorf("error: %v", err)
+		sfError.Message = err.Error()
+		errChannel <- sfError
+		return err
 	}
 
 	sc := &snowflakeConn{rest: sr, cfg: cfg, queryContextCache: (&queryContextCache{}).init(), currentTimeProvider: defaultTimeProvider}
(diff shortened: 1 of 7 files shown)


async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
 		}
 	}
 }
+
+func runLongRunningAsyncQuery(ctx context.Context, t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
```
sfc-gh-pfus, line 201: Not sure if this is the best construction. What do you think about such changes:
1. The whole test function should be put in `runDBTest` (as in all other testes).
2. Remove this function and use just `go dbt.mustQueryContext(ctx, "CALL SYSTEM$WAIT(100, 'SECONDS')")` in the test.
WDYT?

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+	})
+}
+
+func TestLongRunningAsyncQueryFetchResultByID(t *testing.T) {
+	queryIDChan := make(chan string, 1)
+	ctx := WithAsyncMode(context.Background())
+	ctx = WithQueryIDChan(ctx, queryIDChan)
+
+	goRoutineChan := make(chan string)
+	go func(grCh chan string, qIDch chan string) {
```
sfc-gh-pfus, line 212: Not sure, why do we need this code? Why can't we just use `queryID := <-queryIDChan` below?

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+func runLongRunningAsyncQuery(ctx context.Context, t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		_ = dbt.mustQueryContext(ctx, "CALL SYSTEM$WAIT(100, 'SECONDS')")
+	})
+}
+
+func TestLongRunningAsyncQueryFetchResultByID(t *testing.T) {
+	queryIDChan := make(chan string, 1)
+	ctx := WithAsyncMode(context.Background())
+	ctx = WithQueryIDChan(ctx, queryIDChan)
```
sfc-gh-pfus, line 209: A lot of non GO SQL standard features are shown here! Maybe you can extend cmd/async/async.go with some demo and doc.go with description of this contexts?

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+	// Run a long running query asynchronously
+	go runLongRunningAsyncQuery(ctx, t)
+
+	// Get the query ID without waiting for the query to finish
+	queryID := <-goRoutineChan
+	if queryID == "" {
+		t.Fatal("expected a nonempty query ID")
+	}
+
+	conn := openConn(t)
```
sfc-gh-pfus, line 226: If we wrap the whole function with `runDBTest` it won't be necessary, am I right? We can reuse the same connection?

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+	ctx = WithFetchResultByID(ctx, queryID)
+	rows, err := conn.QueryContext(ctx, "")
+	if err != nil {
+		t.Fatalf("failed to run a query. err: %v", err)
+	}
+	defer rows.Close()
+
+	var v string
+	i := 0
+	for {
```
sfc-gh-pfus, line 239: We need more assertions here to ensure that internal assertions are checked. Now, if no records are returned at all, no assertion is triggered, right? We expect very specific number of iterations in this loop, so maybe we can just have something like:
```
assertTrueF(t, rows.Next())
... remaining checks
```
instead of the loops?

sfc-gh-pfus, line 239: Maybe we can use `for rows.NextResultSet() {`?

monitoring.go
```diff
@@ -214,17 +215,40 @@ func (sc *snowflakeConn) getQueryResultResp(
...
+			// For all other scenarios continue processing results response
+			break
+		} else {
+			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
+			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
+			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
+			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
+			time.Sleep(sleepTime)
+		}
+		if retry < len(retryPattern)-1 {
```
sfc-gh-pfus, line 247: This is strongly connected to the `else` above - maybe we can move it there? If so, it would be easier to test without waiting for more retries instead of just one - for instance we could have a test for two retries.

monitoring.go
```diff
@@ -214,17 +215,40 @@ func (sc *snowflakeConn) getQueryResultResp(
...
-		return nil, err
-	}
-	defer res.Body.Close()
+
 	var respd *execResponse
-	if err = json.NewDecoder(res.Body).Decode(&respd); err != nil {
-		logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
-		return nil, err
+	retry := 0
+	retryPattern := []int32{1, 1, 2, 3, 4, 8, 10}
```
sfc-gh-pfus, line 221: As long as this logic is copied, maybe we can refactor this code to common function instead of copying?

async_test.go
sfc-gh-ext-simba-jl: You're right, I don't need the go func.

async_test.go
sfc-gh-ext-simba-jl: Updated

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
 		}
 	}
 }
+
+func runLongRunningAsyncQuery(ctx context.Context, t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		_ = dbt.mustQueryContext(ctx, "CALL SYSTEM$WAIT(100, 'SECONDS')")
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Updated. 50s should be enough. Waiting for the test result on github.

async_test.go
sfc-gh-ext-simba-jl: Which API do you want me to demo? cmd/async/async.go already demo for WithAsyncMode() and I think it's better to keep it simple as is.

async_test.go
sfc-gh-ext-simba-jl: Updated.

async_test.go
sfc-gh-ext-simba-jl: Updated with asserts everywhere in the test.

sfc-gh-ext-simba-jl: The result in `rows.Next()` not `rows.NextResultSet()`

async_test.go
```diff
@@ -196,3 +196,63 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+		queryID := <-queryIDChan
+		grCh <- queryID
+	}(goRoutineChan, queryIDChan)
+
+	// Run a long running query asynchronously
+	go runLongRunningAsyncQuery(ctx, t)
+
+	// Get the query ID without waiting for the query to finish
+	queryID := <-goRoutineChan
+	if queryID == "" {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Updated

monitoring.go
sfc-gh-ext-simba-jl: The fix in [PR #824](https://github.com/snowflakedb/gosnowflake/pull/824) intends to retry until the result is ready.

monitoring.go
sfc-gh-ext-simba-jl: Refactored to `func getQueryResult`

async_test.go
sfc-gh-pfus: I meant fetching results by query id especially (`WithFetchResultByID`). It is worth extending the doc at least, with some code snippet.

monitoring.go
sfc-gh-pfus: What do you think about `getQueryResultWithRetriesForAsyncMode`? `getQueryResult` seems to little specific.

async.go
```diff
@@ -166,3 +133,46 @@ func (sr *snowflakeRestful) getAsync(
...
+			// For all other scenarios continue processing results response
+			break
+		} else {
+			// Sleep before retrying get result request. Exponential backoff up to 5 seconds.
+			// Once 5 second backoff is reached it will keep retrying with this sleeptime.
+			sleepTime := time.Millisecond * time.Duration(500*retryPattern[retry])
+			logger.WithContext(ctx).Infof("Query execution still in progress. Sleep for %v ms", sleepTime)
+			time.Sleep(sleepTime)
+		}
+		if retry < len(retryPattern)-1 {
```
sfc-gh-pfus, line 173: Can we move this inside `else` branch?

sfc-gh-ext-simba-jl: The fix in https://github.com/snowflakedb/gosnowflake/pull/824 intends to retry until the result is ready.

sfc-gh-pfus: Yes, I agree, but if we enter the `if` branch above, the loop breaks. So this part of code is connected only to the case when response code is equal to `queryInProgressAsyncCode`, am I right? For me it would be more readable/descriptive.

async_test.go
sfc-gh-ext-simba-jl: Added description in doc.go and an example cmd/fetchresultbyid/fetchresultbyid.go

monitoring.go
sfc-gh-ext-simba-jl: Renamed

async.go
sfc-gh-ext-simba-jl: Updated.

cmd/fetchresultbyid/fetchresultbyid.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		log.Fatalf("failed to create Config, err: %v", err)
+	}
+
+	dsn, err := sf.DSN(cfg)
+	if err != nil {
+		log.Fatalf("failed to create DSN from Config: %v, err: %v", cfg, err)
+	}
+
+	query := "SELECT 1"
```
sfc-gh-pfus, line 36: Great to have this example! Can we split into two submethods - one synchronous and one async with getting query id from channel?

async.go
```diff
@@ -166,3 +133,48 @@ func (sr *snowflakeRestful) getAsync(
...
+		}
+		defer resp.Body.Close()
+
+		respd = &execResponse{} // reset the response
+		err = json.NewDecoder(resp.Body).Decode(&respd)
+		if err != nil {
+			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
+			return respd, err
+		}
+		if respd.Code != queryInProgressAsyncCode {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Added logging for retry counter.

async_test.go
```diff
@@ -196,3 +196,36 @@ func TestLongRunningAsyncQuery(t *testing.T) {
...
+		queryID := <-queryIDChan
+		assertNotNilF(t, queryID, "expected a nonempty query ID")
+
+		ctx = WithFetchResultByID(ctx, queryID)
+		rows := dbt.mustQueryContext(ctx, "")
+		defer rows.Close()
+
+		var v string
+		i := 0
+		for rows.Next() {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Updated test.

cmd/fetchresultbyid/fetchresultbyid.go
sfc-gh-ext-simba-jl: Updated.

async.go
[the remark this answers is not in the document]
sfc-gh-pfus: But in a way you did it - after reaching retry pattern length, the same counter is logged again and again? Maybe it would be better to increment it always and when reaching the retryPattern prune it to the length?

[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Updated.

sfc-gh-pfus (approved):

PR #1027, opened 2024-01-10
expose `nullType` so we can write nulls to the warehouse from the driver

madisonchamberlain:
### Description
I have a use case where I am using conn.ExecContext to insert values into a table. If the type that I am trying to write is a known null, it would be nice if I could pass the null type as an argument to my `"INSERT INTO %s (%s) VALUES (%s)"` query. In this query the first variable is tablePath, the second is columnName, the third is as many ? as is needed given the number of columns. I will then pass the snowflake type and the value. If the desired snowflake type is null, I would like to be able to pass this in. I know I can just pass in []byte which is not any of the publicly supported types, but this would make the code cleaner and less at risk of changes causing failure 

### Checklist
- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/datatype.go
+++ b/datatype.go
@@ -26,8 +26,8 @@ const (
 	binaryType
 	timeType
 	booleanType
-	// the following are not snowflake types per se but internal types
 	nullType
+	// the following are not snowflake types per se but internal types
 	sliceType
 	changeType
 	unSupportedType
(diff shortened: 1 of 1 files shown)


madisonchamberlain:
I have read the CLA Document and I hereby sign the CLA'

madisonchamberlain:
recheck

sfc-gh-pfus:
Hi @madisonchamberlain ! Thank you for the PR you provided. I'm not sure I fully understand what you mean. I [checked](https://github.com/snowflakedb/gosnowflake/blob/7328b302337856a01936511bfdb61aea3dbd370b/driver_test.go#L1206) that we can insert nils natively. Does your case is somehow more specific because you use templating?

PR #1032, opened 2024-01-17
SNOW-916951: Retry okta auth with new token

sfc-gh-ext-simba-jl:
### Description
issue: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/655

The native Okta Authentication fails in retry due to the same token is used in an sso/saml retry.
This PR fixes the retry logic by generating a new token each time it retries.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -418,6 +418,17 @@ func createRequestBody(sc *snowflakeConn, sessionParameters map[string]interface
 		requestMain.Authenticator = AuthTypeOAuth.String()
 		requestMain.Token = sc.cfg.Token
 	case AuthTypeOkta:
+		samlResponse, err := authenticateBySAML(
+			sc.ctx,
+			sc.rest,
+			sc.cfg.OktaURL,
+			sc.cfg.Application,
+			sc.cfg.Account,
+			sc.cfg.User,
+			sc.cfg.Password)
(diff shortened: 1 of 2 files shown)


auth.go
```diff
@@ -418,6 +418,17 @@ func createRequestBody(sc *snowflakeConn, sessionParameters map[string]interface
 		requestMain.Authenticator = AuthTypeOAuth.String()
 		requestMain.Token = sc.cfg.Token
 	case AuthTypeOkta:
+		samlResponse, err := authenticateBySAML(
```
sfc-gh-pfus, line 421: Let me understand it correctly. So the flow is that first we call identity provider and it returns some SAML response. Then we take response and use it to call Snowflake. The problem is that when retry happens the same SAML response cannot be used to call Snowflake, am I right? So when we create a body, we have to regenerate SAML response every time?

Does the same problem happens for authenticating using external browsers?

sfc-gh-ext-simba-jl: Yes, we will need to regenerate SAML response when retrying. The ticket is logged for okta only.

sfc-gh-pfus (approved):

PR #1040, opened 2024-01-25
SNOW-911238 Add doc about WithOriginalTimestamp

sfc-gh-pfus:
### Description
Added docs about the feature introduced in SNOW-911238.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -455,6 +455,41 @@ of the returned value:
 	    }
 	}
 
+## Arrow batches
+
+You can retrieve data in a columnar format similar to the format a server returns.
+You must use `WithArrowBatches` context, similar to the following:
+
+	    var rows driver.Rows
+		err = conn.Raw(func(x interface{}) error {
+			rows, err = x.(driver.QueryerContext).QueryContext(ctx, query, nil)
+			return err
+		})
+
+		...
+
+		batches, err := rows.(sf.SnowflakeRows).GetArrowBatches()
+
+		... // use Arrow records
+
+Limitations:
+
+1. For some queries Snowflake may decide to return data in JSON format (examples: `SHOW PARAMETERS` or `ls @stage`). You cannot use JSON with Arrow batches context.
+2. Snowflake handles timestamps in a range which is higher than available space in Arrow timestamp type. Because of that special treatment should be used (see below).
+
+### Handling timestamps in Arrow batches
+
+Snowflake returns timestamps natively (from backend to driver) in multiple formats.
+The Arrow timestamp is an 8-byte data type, which is insufficient to handle the larger date and time ranges used by Snowflake.
+Also, Snowflake supports 0-9 (nanosecond) digit precision for seconds, while Arrow supports only 3 (millisecond), 6 (microsecond), an 9 (nanosecond) precision.
+Consequently, Snowflake uses a custom timestamp format in Arrow, which differs on timestamp type and precision.
+
+If you want to use timestamps in Arrow batches, you have two options:
+
+ 1. The Go driver can reduce timestamp struct into simple Arrow Timestamp, if all timestamp values fit into Arrow timestamp field.
+ 2. You can use native Snowflake values. In that case you will receive complex structs as described above. To transform Snowflake values into the Golang time.Time struct you can use `ArrowSnowflakeTimestampToTime`.
+    To enable this feature, you must use `WithOriginalTimestamp` context.
+
 # Binding Parameters
 
 Binding allows a SQL statement to use a value that is stored in a Golang variable.


sfc-gh-dstempniak (approved):

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
 	    }
 	}
 
+## Arrow batches
+
+There is a possibility to retrieve data in a columnar format similar to native server response.
```
sfc-gh-mheavin, line 460: You can retrieve data in a columnar format similar to the format a native server returns.

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+You have to use a proper `WithArrowBatches` context, example:
```
sfc-gh-mheavin, line 461: You must use ... context, similar to the following:

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+
+		...
+
+		batches, err := rows.(sf.SnowflakeRows).GetArrowBatches()
+
+		... // use arrow records
+
+Limitations:
+
+1. When a query regards something that is not real data, it is returned usually in JSON format. It can't be used with arrow batches context. Examples: `SHOW PARAMETERS`.
```
sfc-gh-mheavin, line 477: Not sure what you are trying to say in the first sentence (not real data?).

When a query returns information instead of data, such as a SHOW PARAMETERS query, the information is usually returned as JSON. You cannot use JSON with arrow batches context.

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+2. Snowflake handles timestamps in a range which is higher than available space in Arrow timestamp type. Because of that special treatment should be used (see below).
```
sfc-gh-mheavin, line 478: Why do you use different capitalization for "arrow" (lowercase above, uppercase here)?

Consequently, you should handle timestamps as described in the following section.

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+
+### Handling timestamps in Arrow batches
+
+Snowflake returns timestamps natively (from backend to driver) in multiple formats.
+Arrow timestamp type is only 8 bytes long and is not enough to cover the whole date/time Range Snowflake handles.
```
sfc-gh-mheavin, line 483: The Arrow timestamp is an 8-byte data type, which is insufficient to handle the larger date and time ranges used by Snowflake.

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+Also, Snowflake allows for precision of every fraction of seconds up to 9 (nano), whereas Arrow timestamp allows only 3 (milli), 6 (micro) and 9 (nano) precision.
```
sfc-gh-mheavin, line 484: Also, Snowflake supports nanosecond precision for seconds, while an Arrow timestamp supports only ...

(question. If Arrow supports nano, what is the problem?)

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+Because of that, Snowflake uses custom timestamp format in Arrow which differs on timestamp type and precision.
```
sfc-gh-mheavin, line 485: Consequently, Snowflake uses a custom timestamp format in Arrow, which ....

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+
+If you want to use timestamps in arrow batches, you have two options:
+
+ 1. Go driver can reduce timestamp struct into simple Arrow Timestamp, if all timestamp values fit into Arrow timestamp field.
```
sfc-gh-mheavin, line 489: You can configure the Go driver to reduce the timestamp struct...

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+ 2. You can use native Snowflake values. In that case you will receive complex structs as described above. To transform Snowflake values into Golang time.Time you can use `ArrowSnowflakeTimestampToTime`.
```
sfc-gh-mheavin, line 490: In that case, ... To transform Snowflake values into the Golang `time.Time` struct, you can use the `Arrow..` function.

doc.go
```diff
@@ -455,6 +455,41 @@ of the returned value:
...
+    To enable this feature you have to use `WithOriginalTimestamp` context.
```
sfc-gh-mheavin, line 491: To enable this feature, you must use the...

sfc-gh-mheavin (commented):
Some suggestions and questions

doc.go
sfc-gh-pfus: Done. I didn't like the native in such phrase, so I removed it.

doc.go
sfc-gh-pfus: Wasn't sure how to call it. `SHOW PARAMETERS` returns some response, but this is not customer data, this is just a technical response. Maybe I can rephrase it with:

For some queries Snowflake may decide to return data in JSON format (examples: `SHOW PARAMETERS` or `ls @stage`). You cannot use JSON with Arrow batches context.

What do you think?

doc.go
sfc-gh-pfus: Good catch, changed to capital everywhere.

doc.go
sfc-gh-pfus: What I wanted to say is that Snowflake supports precision from 0 to 9 decimal numbers and Arrow supports only 0, 3, 6, 9.

doc.go
sfc-gh-pfus: Actually this is default behaviour, comparing to the option 2.

doc.go
sfc-gh-mheavin: Also, Snowflake supports 0-9 (nanosecond) digit precision for seconds, while Arrow supports only 3 (millisecond), 6 (microsecond), an 9 (nanosecond) precision.

doc.go
sfc-gh-mheavin: The Go driver can ....

sfc-gh-mheavin (approved):
responded to your questions

PR #1049, opened 2024-02-07
fix a tz bug when validate ErrTooHighTimestampPrecision

Yifeng-Sigma:
### Description
We got error `268003 (22007): Cannot convert timestamp 2021-01-01 01:24:54 +0000 UTC in column Date to Arrow.Timestamp data type due to too high precision. Please use context with WithOriginalTimestamp.` However this should be within the range that nanosecond supports. 

Notice that the date is the first day of the year, that points to where the problem is. The right side is always in UTC while the left side is from `arrowSnowflakeTimestampToTime()` and can have a timezone. 

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing


converter.go
```diff
@@ -1084,7 +1084,7 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
 							ar = arrow.Timestamp(ts.UnixNano())
 							// in case of overflow in arrow timestamp return error
 							// this could only happen for nanosecond case
-							if ts.Year() != ar.ToTime(arrow.Nanosecond).Year() {
+							if ts.UTC().Year() != ar.ToTime(arrow.Nanosecond).Year() {
```
sfc-gh-pfus, line 1087: Great catch indeed! Can you add a test to confirm it doesn't work without the fix and works with it?

Yifeng-Sigma: @sfc-gh-pfus done, using the existing test case, but changed to 1/1. The test failed before the fix.

sfc-gh-pfus (approved):

Yifeng-Sigma:
@sfc-gh-pfus could you please re-trigger the tests? I think they might be flaky?

sfc-gh-dheyman-1:
Hi @Yifeng-Sigma, I rerun the tests and it seems that everything is passing.

Yifeng-Sigma:
> Hi @Yifeng-Sigma, I rerun the tests and it seems that everything is passing.

@sfc-gh-dheyman-1 Thank you! can we merge this?

sfc-gh-dheyman-1:
Thank you @Yifeng-Sigma for fixing this!

PR #1052, opened 2024-02-17
add enableArrowBatchesUtf8Validation to replace invalid utf8 chars in arrow batch

Yifeng-Sigma:
### Description

For historical reasons, snowflake used to allow users to upload invalid utf-8 characters. This caused arrow record to contain invalid utf-8 characters in arrow string columns. 

However by definition arrow.String column should contain utf-8 characters only, see https://arrow.apache.org/docs/cpp/api/datatype.html, 
https://github.com/apache/arrow/blob/a03d957b5b8d0425f9d5b6c98b6ee1efa56a1248/go/arrow/datatype.go#L73-L74

This could cause other upstream services that consume the bad arrow record to break, e.g Rust arrow lib force some basic validation, and this bad arrow record will fail the validation. 

This change introduce `enableArrowBatchesUtf8Validation`, when enabled, will iterate all values of string columns and replace any invalid character. 

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -1120,7 +1120,7 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
 					if col.(*array.String).IsValid(i) {
 						stringValue := col.(*array.String).Value(i)
 						if !utf8.ValidString(stringValue) {
-							logger.Error("Invalid utf8 detected!")
+							logger.Error("Invalid UTF-8 characters detected while reading query response, column: ", srcColumnMeta.Name)
 							stringValue = strings.ToValidUTF8(stringValue, "�")
 						}
 						tb.Append(stringValue)
--- a/converter.go
+++ b/converter.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/array"
@@ -983,6 +984,15 @@ func higherPrecisionEnabled(ctx context.Context) bool {
 	return ok && d
 }
 
+func arrowBatchesUtf8ValidationEnabled(ctx context.Context) bool {
+	v := ctx.Value(enableArrowBatchesUtf8Validation)
+	if v == nil {
+		return false
+	}
+	d, ok := v.(bool)
+	return ok && d
+}
+
 func getArrowBatchesTimestampOption(ctx context.Context) snowflakeArrowBatchesTimestampOption {
 	v := ctx.Value(arrowBatchesTimestampOption)
 	if v == nil {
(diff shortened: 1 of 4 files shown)


Yifeng-Sigma:
cc: @sfc-gh-wfateem

converter_test.go
```diff
@@ -1135,6 +1137,27 @@ func TestArrowToRecord(t *testing.T) {
...
+			values:                           []string{"\xFF", "bar"},
+			enableArrowBatchesUtf8Validation: true,
+			nrows:                            2,
+			builder:                          array.NewStringBuilder(pool),
+			append:                           func(b array.Builder, vs interface{}) { b.(*array.StringBuilder).AppendValues(vs.([]string), valids) },
+			compare: func(src interface{}, convertedRec arrow.Record) int {
+				srcvs := src.([]string)
+				arr := convertedRec.Column(0).(*array.String)
+				for i := 0; i < arr.Len(); i++ {
+					if !utf8.ValidString(arr.Value(i)) || strings.ToValidUTF8(srcvs[i], "�") != string(arr.Value(i)) {
```
sfc-gh-dheyman-1, line 1154: Perhaps we could somehow check converted values against literals, to avoid reusing `strings.ToValidUTF8(srcvs[i], "�")` which is our production code?

sfc-gh-dheyman-1 (approved):

converter_test.go
Yifeng-Sigma: make sense, changed.

sfc-gh-dheyman-1:
@Yifeng-Sigma , could you also update `doc.go` and include there this new context parameter?

Yifeng-Sigma:
> @Yifeng-Sigma , could you also update `doc.go` and include there this new context parameter?

done

sfc-gh-dheyman-1 (approved):

converter.go
```diff
@@ -1098,6 +1108,26 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
...
+		case textType:
+			if arrowBatchesUtf8ValidationEnabled(ctx) && col.DataType().ID() == arrow.STRING {
+				tb := array.NewStringBuilder(pool)
+				defer tb.Release()
+
+				for i := 0; i < int(numRows); i++ {
+					if col.(*array.String).IsValid(i) {
+						stringValue := col.(*array.String).Value(i)
+						if !utf8.ValidString(stringValue) {
+							logger.Error("Invalid utf8 detected!")
```
sfc-gh-pfus, line 1123: Can we add some more context here? If I saw such message in logs without any more info I would be completely confused. How about `Invalid UTF-8 characters detected while reading query response` maybe include column name (if we have it)? WDYT?

Yifeng-Sigma: Previously I log out the whole cell content but I think that's too risky, we should not log any customer data.

I guess column name is ok?

Yifeng-Sigma: @sfc-gh-pfus , done.

sfc-gh-pfus: Yes, logging value is too risky. Column name is good enough. Thank you!

sfc-gh-pfus:
@Yifeng-Sigma thanks a lot again!

PR #1055, opened 2024-02-19
SNOW-999076: Fix application directory

sfc-gh-pbulawa:
### Description
Fix application directory which is used to get the config file.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/client_configuration_test.go
+++ b/client_configuration_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


client_configuration_test.go
```diff
@@ -71,13 +72,15 @@ func TestNotFindConfigFileWhenNotDefined(t *testing.T) {
 }
 
 func TestCreatePredefinedDirs(t *testing.T) {
+	exeDir, _ := os.Executable()
```
sfc-gh-mhofman, line 75: if this is a symlink should we look for config in the place where symlink is located or evaluate the symlink and look where it points to?

sfc-gh-knozderko (approved):

PR #1058, opened 2024-02-23
SNOW-903604 Zombie dbus processes generated by one of our dependencies github.com/99designs/keyring

sfc-gh-dszmolka:
SNOW-903604 Zombie dbus processes generated by one of our dependencies github.com/99designs/keyring

After long consideration and discussion how we want to address the issue we decided to go with only documenting for now, which is what i tried to do here, alongside with trying to detect if the OS is susceptible for the issue (not all Linux are)

If so, log a warning of the issue + workaround which is the same as documented in the README now.

Of course this is by no means a solution - long term solution will be to possibly replace the current keyring implementation.

### Description

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/util.go
+++ b/util.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"math/rand"
 	"os"
+	"os/exec"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


util.go
```diff
@@ -292,3 +293,30 @@ func contains[T comparable](s []T, e T) bool {
...
+}
+
+func canDbusLeakProcesses() (bool, string) {
+	// TODO: delete this once we replaced 99designs/keyring (SNOW-1017659) and/or keyring#103 is resolved
+	leak := false
+	message := ""
+
+	valDbus, haveDbus := os.LookupEnv("DBUS_SESSION_BUS_ADDRESS")
+	if !haveDbus || strings.Contains(valDbus, "unix:abstract") {
+		// if DBUS_SESSION_BUS_ADDRESS is not set or set to an abstract socket, it's not necessarily a problem, only
```
sfc-gh-pfus, line 313: Is there some comment missing?

sfc-gh-dszmolka: no, i was considering lines 313-314 together 😅  , comment meaning ... 'not necessarily a problem, only if the dbus-daemon is running'

sfc-gh-pfus: I'd prefer having full sentences, not joining comments with code :)

sfc-gh-dszmolka: fixed!

sfc-gh-pfus (approved):

PR #1077, opened 2024-03-20
SNOW-1256926 Add converter from snowflake date/time format to go

sfc-gh-pfus:
### Description
Provided converter from snowflake date time formats to golang formats

### Checklist
- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/datetime.go
+++ b/datetime.go
@@ -1,0 +1,62 @@
+// Copyright (c) 2024 Snowflake Computing Inc. All rights reserved.
+
+package gosnowflake
+
+import (
+	"errors"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var incorrectSecondsFractionRegex = regexp.MustCompile(`[^.,]FF`)
+var correctSecondsFractionRegex = regexp.MustCompile(`FF(?P<fraction>\d?)`)
+
+type formatReplacement struct {
+	input  string
+	output string
+}
+
+var formatReplacements = []formatReplacement{
+	{input: "YYYY", output: "2006"},
+	{input: "YY", output: "06"},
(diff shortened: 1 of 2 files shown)


datetime.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func snowflakeFormatToGoFormat(sfFormat string) (string, error) {
+	res := sfFormat
+	for _, replacement := range formatReplacements {
+		res = strings.Replace(res, replacement.input, replacement.output, -1)
+	}
+
+	if incorrectSecondsFractionRegex.MatchString(res) {
+		return "", errors.New("incorrect second fraction - golang requires fraction to be preceded by comma or decimal point")
```
sfc-gh-dheyman-1, line 44: I would add scenario for this error as well.

datetime.go
```diff
@@ -0,0 +1,62 @@
...
+	{input: "AM", output: "PM"},
+	{input: "MI", output: "04"},
+	{input: "SS", output: "05"},
+	{input: "TZH", output: "Z07"},
+	{input: "TZM", output: "00"},
+}
+
+func snowflakeFormatToGoFormat(sfFormat string) (string, error) {
+	res := sfFormat
+	for _, replacement := range formatReplacements {
```
sfc-gh-dstempniak, line 39: If the format is `DD/MM/YY` it can be parsed correctly, but it will lead to incorrect time. I wonder if we can verify it somehow.

datetime.go
sfc-gh-pfus: Right, done.

sfc-gh-dheyman-1 (approved):

PR #1080, opened 2024-03-23
Allow enableHigherPrecision to be used in arrow batches 

Yifeng-Sigma:
### enableHigherPrecision in Arrow batches
Snowflake returns BigDecimal natively (from backend to driver)
To preserve BigDecimal values within Arrow batches, set the `enableHigherPrecision` value to true in the context using `WithHigherPrecision“.

### Benefits
This offers two main benefits: it helps avoid precision loss and defers the conversion to upstream services.
Alternatively, without this setting, all non-zero scale numbers will be converted to float64, potentially resulting in loss of precision.

### Checklist

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -505,9 +505,8 @@ To address this issue and prevent potential downstream disruptions, the context 
 When enabled, this feature iterates through all values in string columns, identifying and replacing any invalid characters with `�`.
 This ensures that Arrow records conform to the UTF-8 standards, preventing validation failures in downstream services like the Rust Arrow library that impose strict validation checks.
 
-### BigDecimal in Arrow batches
-Snowflake returns BigDecimal natively (from backend to driver)
-To preserve BigDecimal values within Arrow batches, set the `arrowBatchesOriginalBigDecimal` value to true in the context using `WithArrowBatchesOriginalBigDecimal`.
+### WithHigherPrecision in Arrow batches
+To preserve BigDecimal values within Arrow batches, set the `enableHigherPrecision` value to true in the context using `WithHigherPrecision`.
 This offers two main benefits: it helps avoid precision loss and defers the conversion to upstream services.
 Alternatively, without this setting, all non-zero scale numbers will be converted to float64, potentially resulting in loss of precision.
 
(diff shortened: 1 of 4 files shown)


doc.go
```diff
@@ -505,6 +505,11 @@ To address this issue and prevent potential downstream disruptions, the context
 When enabled, this feature iterates through all values in string columns, identifying and replacing any invalid characters with `�`.
 This ensures that Arrow records conform to the UTF-8 standards, preventing validation failures in downstream services like the Rust Arrow library that impose strict validation checks.
 
+### WithHigherPrecision in Arrow batches
+To preserve BigDecimal values within Arrow batches, set the `enableHigherPrecision` value to true in the context using `WithHigherPrecision`.
+This offers two main benefits: it helps avoid precision loss and defers the conversion to upstream services.
+Alternatively, without this setting, all non-zero scale numbers will be converted to float64, potentially resulting in loss of precision.
```
sfc-gh-pfus, line 511: Or int64, am I right?

doc.go
```diff
@@ -505,6 +505,11 @@ To address this issue and prevent potential downstream disruptions, the context
 When enabled, this feature iterates through all values in string columns, identifying and replacing any invalid characters with `�`.
 This ensures that Arrow records conform to the UTF-8 standards, preventing validation failures in downstream services like the Rust Arrow library that impose strict validation checks.
 
+### WithHigherPrecision in Arrow batches
+To preserve BigDecimal values within Arrow batches, set the `enableHigherPrecision` value to true in the context using `WithHigherPrecision`.
```
sfc-gh-pfus, line 509: I'm not sure we should mention internal parameter name. Maybe just `use WithHigherPrecision` is good enough?

Yifeng-Sigma: done!

doc.go
Yifeng-Sigma: Actually that's for zero scale numbers.
Added `Zero-scale numbers (DECIMAL256, DECIMAL128) will be converted to int64, which could lead to overflow.`

sfc-gh-pfus: Great, thanks!

PR #1082, opened 2024-03-25
SNOW-1256908 Implement structured objects

sfc-gh-pfus:
### Description

SNOW-1256908 Implementation of structured objects

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/big"
@@ -107,7 +108,7 @@ func goTypeToSnowflake(v driver.Value, tsmode snowflakeType) snowflakeType {
 }
 
 // snowflakeTypeToGo translates Snowflake data type to Go data type.
-func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
+func snowflakeTypeToGo(dbtype snowflakeType, scale int64, fields []fieldMetadata) reflect.Type {
 	switch dbtype {
 	case fixedType:
 		if scale == 0 {
@@ -116,7 +117,7 @@ func snowflakeTypeToGo(dbtype snowflakeType, scale int64) reflect.Type {
 		return reflect.TypeOf(float64(0))
 	case realType:
 		return reflect.TypeOf(float64(0))
-	case textType, variantType, objectType, arrayType:
+	case textType, variantType, arrayType:
 		return reflect.TypeOf("")
 	case dateType, timeType, timestampLtzType, timestampNtzType, timestampTzType:
 		return reflect.TypeOf(time.Now())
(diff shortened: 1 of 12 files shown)


converter.go
```diff
@@ -631,6 +594,184 @@ func arrowToValue(
...
+			return f
+		}
+		return fmt.Sprintf("%.*f", scale, float64(val)/math.Pow10(scale))
+	}
+	return nil
+}
+
+func arrowInt16ToValue(srcValue *array.Int16, rowIdx int, higherPrecision bool, scale int) snowflakeValue {
+	if !srcValue.IsNull(rowIdx) {
+		val := srcValue.Value(rowIdx)
```
sfc-gh-dheyman-1, line 699: I see that there is some duplication among arrowInt**ToValue methods. Perhaps extract/generify this duplicated logic?

datetime.go
```diff
@@ -60,3 +60,34 @@ func snowflakeFormatToGoFormat(sfFormat string) (string, error) {
 	}
 	return res, nil
 }
+
+func dateTimeFormatByType(dateTimeType string, params map[string]*string) (string, error) {
```
sfc-gh-dheyman-1, line 64: test for that?

converter.go
```diff
@@ -631,6 +594,184 @@ func arrowToValue(
...
+
+func arrowTimeToValue(srcValue arrow.Array, rowIdx int, scale int) snowflakeValue {
+	t0 := time.Time{}
+	if srcValue.DataType().ID() == arrow.INT64 {
+		if !srcValue.IsNull(rowIdx) {
+			return t0.Add(time.Duration(srcValue.(*array.Int64).Value(rowIdx) * int64(math.Pow10(9-scale))))
+		}
+	} else {
+		if !srcValue.IsNull(rowIdx) {
+			return t0.Add(time.Duration(int64(srcValue.(*array.Int32).Value(rowIdx)) * int64(math.Pow10(9-scale))))
```
sfc-gh-dheyman-1, line 769: don't we have a util for this math logic?

query.go
```diff
@@ -47,13 +47,21 @@ type contextData struct {
...
+	Precision  int64           `json:"precision"`
+	Scale      int64           `json:"scale"`
+	Nullable   bool            `json:"nullable"`
+}
+
+type fieldMetadata struct {
+	Name   string          `json:"name"`
+	Type   string          `json:"type"`
+	Scale  int             `json:"scale"`
+	Fields []fieldMetadata `json:"fields"`
```
sfc-gh-pmotacki, line 64: Here should be only 4 properties

datetime.go
sfc-gh-pfus: It is tested via integration test. Do you see a value in unit testing?

query.go
sfc-gh-pfus: Yes, this is enough - we don't need remaining fields.

sfc-gh-pmotacki (approved):
LGTM

sfc-gh-dheyman-1 (approved):

PR #1087, opened 2024-03-29
SNOW-1256916 implement structured arrays

sfc-gh-pfus:
### Description

SNOW-1256916 Implemented structured arrays

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -9,6 +9,7 @@ import (
 	"database/sql/driver"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -109,7 +110,7 @@ func goTypeToSnowflake(v driver.Value, tsmode snowflakeType) snowflakeType {
 }
 
 // snowflakeTypeToGo translates Snowflake data type to Go data type.
-func snowflakeTypeToGo(dbtype snowflakeType, scale int64, fields []fieldMetadata) reflect.Type {
+func snowflakeTypeToGo(ctx context.Context, dbtype snowflakeType, scale int64, fields []fieldMetadata) reflect.Type {
 	switch dbtype {
 	case fixedType:
 		if scale == 0 {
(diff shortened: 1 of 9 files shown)


converter.go
```diff
@@ -361,11 +391,112 @@ func stringToValue(dest *driver.Value, srcColumnMeta execResponseRowType, srcVal
...
+			return err
+		}
+		*dest = v
+		return nil
 	}
 	*dest = *srcValue
 	return nil
 }
 
+func buildStructuredArray(fieldMetadata fieldMetadata, srcValue *string, params map[string]*string) (any, error) {
```
sfc-gh-dheyman-1, line 413: This is for JSON or Arrow format with JSON structured types, correct?

converter.go
```diff
@@ -521,12 +652,153 @@ func arrowToValue(rowIdx int, srcColumnMeta execResponseRowType, srcValue arrow.
...
+			}
+			return binaries, nil
+		case dateType:
+			dates := make([]time.Time, offsets[rowIdx+1]-offsets[rowIdx])
+			for j := offsets[rowIdx]; j < offsets[rowIdx+1]; j++ {
+				dates[j-offsets[rowIdx]] = arrowDateToValue(values.(*array.Date32), int(j)).(time.Time)
+			}
+			return dates, nil
+		case timeType:
+			times := make([]time.Time, offsets[rowIdx+1]-offsets[rowIdx])
```
sfc-gh-dheyman-1, line 780: there is a lot of potentially generic logic (iterating through data and offsets). Can we somehow extract this to avoid writing so many loops?

converter.go
sfc-gh-pfus: Correct.

converter.go
sfc-gh-pfus: I refactored it a bit. Not sure if it is better though.

converter.go
```diff
@@ -570,6 +806,18 @@ func arrowToValue(rowIdx int, srcColumnMeta execResponseRowType, srcValue arrow.
 	return nil, fmt.Errorf("unsupported data type")
 }
 
+func mapStructuredArrayNativeArrowRows[T any](offsets []int32, rowIdx int, values arrow.Array, f func(j int) (T, error)) (snowflakeValue, error) {
+	arr := make([]T, offsets[rowIdx+1]-offsets[rowIdx])
+	for j := offsets[rowIdx]; j < offsets[rowIdx+1]; j++ {
+		v, err := f(int(j))
```
sfc-gh-dheyman-1, line 812: I think this is better. I would just give this passed function more meaningful name instead of `f`.

sfc-gh-dheyman-1 (approved):

PR #1089, opened 2024-04-03
SNOW-1256917 Implement structured maps

sfc-gh-pfus:
### Description

SNOW-1256917 Implemented structured maps

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -164,8 +164,43 @@ func snowflakeTypeToGo(ctx context.Context, dbtype snowflakeType, scale int64, f
 			return reflect.TypeOf([]ObjectType{})
 		}
 		return nil
+	case mapType:
+		switch getSnowflakeType(fields[0].Type) {
+		case textType:
+			return snowflakeTypeToGoForMaps[string](ctx, fields[1])
+		case fixedType:
+			return snowflakeTypeToGoForMaps[int64](ctx, fields[1])
+		}
+		return reflect.TypeOf(map[any]any{})
 	}
 	logger.Errorf("unsupported dbtype is specified. %v", dbtype)
+	return reflect.TypeOf("")
+}
+
+func snowflakeTypeToGoForMaps[K comparable](ctx context.Context, valueMetadata fieldMetadata) reflect.Type {
+	switch getSnowflakeType(valueMetadata.Type) {
+	case textType:
+		return reflect.TypeOf(map[K]string{})
+	case fixedType:
+		if higherPrecisionEnabled(ctx) && valueMetadata.Scale == 0 {
+			return reflect.TypeOf(map[K]*big.Int{})
+		} else if higherPrecisionEnabled(ctx) && valueMetadata.Scale != 0 {
+			return reflect.TypeOf(map[K]*big.Float{})
+		} else if !higherPrecisionEnabled(ctx) && valueMetadata.Scale == 0 {
+			return reflect.TypeOf(map[K]int64{})
+		} else {
+			return reflect.TypeOf(map[K]float64{})
+		}
+	case realType:
+		return reflect.TypeOf(map[K]float64{})
+	case booleanType:
+		return reflect.TypeOf(map[K]bool{})
+	case binaryType:
+		return reflect.TypeOf(map[K][]byte{})
+	case timeType, dateType, timestampTzType, timestampNtzType, timestampLtzType:
+		return reflect.TypeOf(map[K]time.Time{})
+	}
+	logger.Errorf("unsupported dbtype is specified for map value")
 	return reflect.TypeOf("")
 }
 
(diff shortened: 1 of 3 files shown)


converter.go
```diff
@@ -818,24 +959,137 @@ func mapStructuredArrayNativeArrowRows[T any](offsets []int32, rowIdx int, creat
...
+	switch srcColumnMeta.Fields[1].Type {
+	case "text":
+		return mapStructuredMapNativeArrowRows(make(map[K]string), offsets, rowIdx, keyFunc, func(j int) (string, error) {
+			return items.(*array.String).Value(j), nil
+		})
+	case "boolean":
+		return mapStructuredMapNativeArrowRows(make(map[K]bool), offsets, rowIdx, keyFunc, func(j int) (bool, error) {
+			return items.(*array.Boolean).Value(j), nil
+		})
+	case "fixed":
```
sfc-gh-dheyman-1, line 988: I see that in case of "fixed" actual logic differs only by type correct? can we remove the code duplication? should be fairly easy to extract common logic

converter.go
```diff
@@ -818,24 +959,137 @@ func mapStructuredArrayNativeArrowRows[T any](offsets []int32, rowIdx int, creat
...
+		})
+	case "timestamp_ltz", "timestamp_ntz", "timestamp_tz":
+		return mapStructuredMapNativeArrowRows(make(map[K]time.Time), offsets, rowIdx, keyFunc, func(j int) (time.Time, error) {
+			return *arrowSnowflakeTimestampToTime(items, getSnowflakeType(srcColumnMeta.Fields[1].Type), srcColumnMeta.Fields[1].Scale, j, loc), nil
+		})
+	}
+	return nil, errors.New("Unsupported map value: " + srcColumnMeta.Fields[1].Type)
+}
+
+func extractNumberFromArrow(values *arrow.Array, j int, higherPrecision bool, scale int) (snowflakeValue, error) {
```
sfc-gh-dheyman-1, line 1046: can we keep second arg as `rowIdx` instead of `j`? helps in understanding whats going on.

structured_type_test.go
```diff
@@ -236,19 +237,26 @@ func TestArrayAndMetadata(t *testing.T) {
...
 					query:     "SELECT ARRAY_CONSTRUCT(1.1, 2.2)::ARRAY(NUMBER(38, 19)) as structured_type UNION SELECT ARRAY_CONSTRUCT()::ARRAY(NUMBER(38, 19))",
 					expected1: []float64{1.1, 2.2},
 					expected2: []float64{},
 					actual:    []float64{},
 				},
+				{
+					name:      "double",
+					query:     "SELECT ARRAY_CONSTRUCT(1.1, 2.2)::ARRAY(DOUBLE) as structured_type UNION SELECT ARRAY_CONSTRUCT()::ARRAY(DOUBLE)",
+					expected1: []float64{1.1, 2.2},
+					expected2: []float64{},
```
sfc-gh-dheyman-1, line 257: replacing `expected1` and `expected2` with an array of expected values would benefit this test with more flexibility of assertions. Same with `TestMapAndMetadata`

converter.go
```diff
@@ -789,6 +896,40 @@ func arrowToValue(rowIdx int, srcColumnMeta execResponseRowType, srcValue arrow.
...
+					k, err := extractInt64(keys, int(j))
+					if err != nil {
+						return 0, err
+					}
+					return k, nil
+				}
+				return buildStructuredMapFromArrow(rowIdx, srcColumnMeta, offsets, keyFunc, items, higherPrecision, loc)
+			}
+		}
+		return nil, nil
```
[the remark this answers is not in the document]
sfc-gh-pfus: You are right. I removed return - the code should end up with error return at the end of the function.

converter.go
sfc-gh-pfus: Done.

converter.go
```diff
@@ -818,24 +959,137 @@ func mapStructuredArrayNativeArrowRows[T any](offsets []int32, rowIdx int, creat
...
+			return mapStructuredMapNativeArrowRows(make(map[K]*big.Float), offsets, rowIdx, keyFunc, func(j int) (*big.Float, error) {
+				v, err := extractNumberFromArrow(&items, j, higherPrecision, srcColumnMeta.Fields[1].Scale)
+				if err != nil {
+					return nil, err
+				}
+				return v.(*big.Float), nil
+			})
+		} else if !higherPrecision && srcColumnMeta.Fields[1].Scale == 0 {
+			return mapStructuredMapNativeArrowRows(make(map[K]int64), offsets, rowIdx, keyFunc, func(j int) (int64, error) {
+				v, err := extractNumberFromArrow(&items, int(j), higherPrecision, srcColumnMeta.Fields[1].Scale)
```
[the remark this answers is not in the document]
sfc-gh-pfus: We don't.

converter.go
sfc-gh-pfus: Actually we can't because this is not `rowIdx`. This is offset indicator for specific row.

structured_type_test.go
sfc-gh-pfus: To be honest I don't understand what you mean. You mean renaming the variable or what?

sfc-gh-dheyman-1: I meant having an array instead of two fields. Just cause of scalability instead of potential introducing expected3 ;)

converter.go
sfc-gh-dheyman-1: Got it.

structured_type_test.go
sfc-gh-pfus: Hmmm, I'd like to leave it as is, as it is easier to read, but if you insist we can change it :)

structured_type_test.go
```diff
@@ -425,6 +426,238 @@ func TestArrayOfObjects(t *testing.T) {
...
+			},
+			{
+				name:      "string timestamp_ltz",
+				query:     "SELECT {'a': '2024-01-05 11:22:33'::TIMESTAMP_LTZ, 'b': '2024-01-06 11:22:33'::TIMESTAMP_LTZ}::MAP(VARCHAR, TIMESTAMP_LTZ) as structured_type UNION SELECT {'c': '2024-01-07 11:22:33'::TIMESTAMP_LTZ}::MAP(VARCHAR, TIMESTAMP_LTZ)",
+				expected1: map[string]time.Time{"a": time.Date(2024, time.January, 5, 11, 22, 33, 0, warsawTz), "b": time.Date(2024, time.January, 6, 11, 22, 33, 0, warsawTz)},
+				expected2: map[string]time.Time{"c": time.Date(2024, time.January, 7, 11, 22, 33, 0, warsawTz)},
+				actual:    make(map[string]time.Time),
+			},
+			{
+				name:      "integer timestamp_ltz",
```
sfc-gh-pmotacki, line 604: Does it should work for hierarchical structures like map of maps for example?

sfc-gh-pfus: It should, but this is a scope of separate task: https://snowflakecomputing.atlassian.net/browse/SNOW-1256899

sfc-gh-pmotacki: 🆗

sfc-gh-pmotacki (approved):

structured_type_test.go
sfc-gh-dheyman-1: Ok, if you say so.

sfc-gh-dheyman-1 (approved):

PR #1094, opened 2024-04-08
Allow caller to set log fields

madisonchamberlain:
### Description

This will allow the user to set fields on the logs. This is needed because at the moment, the driver does not tag the requestID, so its extremely difficult to debug issues due to the fact that you cant tell which request a log is associated with. Some examples of useful data we could tag with this method includes: client side requestIDs, server side requestIDs given that `gosnowflake.WithRequestID` is used, and snowflake query ID given that we are are using the result reuse feature `gosnowflake.WithFetchResultByID`.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,41 @@ Install [jq](https://stedolan.github.io/jq) so that the parameters can get parse
 make test
 ```
 
+## customizing Logging Tags
+
+If you would like to ensure that certain tags are always present in the logs, `RegisterClientLogContextHook` can be used in your init function. See example below.
+```
+import "github.com/snowflakedb/gosnowflake"
+
+func init() {
+    // each time the logger is used, the logs will contain a REQUEST_ID field with requestID the value extracted 
+    // from the context
+	gosnowflake.RegisterClientLogContextHook("REQUEST_ID", func(ctx context.Context) interface{} {
+		return requestIdFromContext(ctx)
+	})
+}
+```
+
+## Setting Log Level
+If you want to change the log level, `SetLogLevel` can be used in your init function like this:
+```
+import "github.com/snowflakedb/gosnowflake"
+
+func init() {
+    // The following line changes the log level to debug
+	_ = gosnowflake.GetLogger().SetLogLevel("debug")
+}
+```
+The following is a list of options you can pass in to set the level from least to most verbose: 
+- `"OFF"`
+- `"error"`
+- `"warn"`
+- `"print"`
+- `"trace"`
+- `"debug"`
+- `"info"`
+
+
 ## Capturing Code Coverage
 
 Configure your testing environment as described above and run ``make cov``. The coverage percentage will be printed on the console when the testing completes.
--- a/auth.go
+++ b/auth.go
@@ -225,7 +225,7 @@ func postAuth(
 	params.Add(requestGUIDKey, NewUUID().String())
 
 	fullURL := sr.getFullURL(loginRequestPath, params)
-	logger.Infof("full URL: %v", fullURL)
+	logger.WithContext(ctx).Infof("full URL: %v", fullURL)
 	resp, err := sr.FuncAuthPost(ctx, client, fullURL, headers, bodyCreator, timeout, sr.MaxRetryCount)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -95,6 +95,27 @@ Install [jq](https://stedolan.github.io/jq) so that the parameters can get parse
 make test
 ```
 
+## Customize Logging
```
sfc-gh-pfus, line 98: I'm glad you added this doc. But can we split it into two sections, maybe `setting log level` and `customizing logging tags`?

README.md
```diff
@@ -95,6 +95,27 @@ Install [jq](https://stedolan.github.io/jq) so that the parameters can get parse
...
+
+If you would like to ensure that certain tags are always present in the logs, or to set a particular log level, `RegisterClientLogContextHook` or `SetLogLevel` can be used
+in your init function. See example below.
+```
+import "github.com/snowflakedb/gosnowflake"
+
+func init() {
+    // changes the log level to debug
+	sfLogger := gosnowflake.GetLogger()
```
sfc-gh-pfus, line 107: Maybe we can write is as `gosnowflake.GetLogger().SetLogLevel("debug")`, wdyt?

README.md
```diff
@@ -95,6 +95,27 @@ Install [jq](https://stedolan.github.io/jq) so that the parameters can get parse
 make test
```
sfc-gh-pfus, line 95: General notices.
1. Why `WithContext` is not enough for you?
2. You know, that this works only if `WithContext` is used before?

madisonchamberlain: I updated all logs to include `WithContext` where a context is passed so that this will work in all possible places. I think the thing that is nice about this is that by setting the tags on init, it gives the client an easy method to ensure that all logs have the same tag, rather than needing to wrap each call. Does that seem reasonable?

sfc-gh-pfus: I'm not sure I understand. When you add `WithContext`, all fields defined in `LogKeys` variable is used. Maybe we don't need two mechanisms of this, maybe it would be better to define something like:

```
func registerContextVariableToLog(name string) {
    LogKeys = append(LogKeys, name)
}
```

WDYT?

madisonchamberlain: Ah I wasnt aware I could use this since LogKeys is currently configured to be an array of fixed length (set to 2). Can I alter it to be an array of any length?

madisonchamberlain: One other thing to note is that we typically use struct{} instead of string as the type to set context keys, since there are some benefits (compiler will check for duplicate keys, type safety etc) so maybe LogKeys could take in any?

sfc-gh-pfus: Ad 1 - yes, please do.
Ad 2 - yes, I think you can. Can you please add some tests to `TestLogEntryCtx`?

madisonchamberlain: Now that I look into it more, the usage of the contextKey as the key in the log will cause problems if we allow the LogKeys to be type any. I will update LogKeys to be a map of string to type contextKey where contextKey is any. I will add a test to confirm that it works as expected

sfc-gh-pfus: I don't think we should do that. This may be an incompatible change in public API. So we should either stick to array of strings or use something more generic.

madisonchamberlain: Ok, what do you think of not changing the type of contextKey, but having LogKeys map string -> any? I added a test to confirm that the behavior of `context2Fields` does not change if we try it with that

sfc-gh-pfus: Yes, that seems backward compatible :)

sfc-gh-pfus: Unfortunately, changing a type from list to map is still backward incompatible :) What I mean is that `LogKeys` is a publicly accessible variable, meaning that our clients may change it. So if we change the type of this field, their code won't compile anymore.

I'm wondering about root of what you need. Is this just logging specific context key/value or is this something more? If it is only logging key/value, why can't you just add additional value to `LogKeys` in your app?

madisonchamberlain: Ah I see. The use case for us is that since its the go recommendation to use structs as the contextKeys, rather than string, and since LogKeys takes in the key to search the context for, we are not able to use the map because of the type restriction. All of our keys are structs, so that was where the initial idea of having the separate data store came from

sfc-gh-pfus: Ahh, I can finally see it! So because of strong typing, we can't modify `LogKeys` type. In that case, the solution you proposed previously was proper. Apologies for not understanding it earlier!

So in my opinion the best one was with registering hook - log key and function retrieving from context. So it should be probably something like
```
func RegisterLogContextHook(contextKey, ctxExtractor func(ctx contex.Context) string)
```

madisonchamberlain: Ok thank you for all your help! Making the update now!

sfc-gh-pfus (approved):

sfc-gh-pfus (approved):

PR #1096, opened 2024-04-10
SNOW-1283946 Embed structured types

sfc-gh-pfus:
### Description

SNOW-1283946 Handle embedded structured types.

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -349,12 +349,15 @@ func stringToValue(ctx context.Context, dest *driver.Value, srcColumnMeta execRe
 			return nil
 		}
 		m := make(map[string]any)
-		decoder := json.NewDecoder(bytes.NewBufferString(*srcValue))
-		decoder.UseNumber()
+		decoder := decoderWithNumbersAsStrings(srcValue)
 		if err := decoder.Decode(&m); err != nil {
 			return err
 		}
-		*dest = buildStructuredTypeRecursive(m, srcColumnMeta.Fields, params)
+		v, err := buildStructuredTypeRecursive(ctx, m, srcColumnMeta.Fields, params)
+		if err != nil {
+			return err
+		}
+		*dest = v
 		return nil
 	case "text", "fixed", "real", "variant":
 		*dest = *srcValue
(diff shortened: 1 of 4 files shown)


converter.go
```diff
@@ -1022,6 +1003,153 @@ func arrowToValue(ctx context.Context, rowIdx int, srcColumnMeta execResponseRow
 	return nil, fmt.Errorf("unsupported data type")
 }
 
+func handleNativeArrowMap(ctx context.Context, rowIdx int, keyMetadata, valueMetadata fieldMetadata, srcValue arrow.Array, loc *time.Location, higherPrecision bool, params map[string]*string) (snowflakeValue, error) {
```
sfc-gh-dheyman-1, line 1006: i have impression we pass too many values between functions... perhaps we can introduce some kind of conversion context to wrap these that are inseparable?

converter.go
```diff
@@ -851,98 +936,16 @@ func arrowToValue(ctx context.Context, rowIdx int, srcColumnMeta execResponseRow
...
-			})
-		case objectType:
-			return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) (*structuredType, error) {
-				m := make(map[string]any, len(srcColumnMeta.Fields[0].Fields))
-				for fieldIdx, field := range srcColumnMeta.Fields[0].Fields {
-					m[field.Name] = values.(*array.Struct).Field(fieldIdx).ValueStr(int(j))
-				}
-				return buildStructuredTypeRecursive(m, srcColumnMeta.Fields[0].Fields, params), nil
-			})
+			return nil, nil
```
sfc-gh-dheyman-1, line 946: perhaps we can log that this ocurred?

converter.go
sfc-gh-dheyman-1, line 1006: `build` instead of `handle`?

structured_type_test.go
```diff
@@ -569,6 +600,114 @@ func TestArrayOfObjects(t *testing.T) {
 	})
 }
 
+func TestArrayOfArrays(t *testing.T) {
+	warsawTz, err := time.LoadLocation("Europe/Warsaw")
+	assertNilF(t, err)
+	testcases := []struct {
+		name     string
+		query    string
+		actual   any
```
sfc-gh-dheyman-1, line 609: this is more of an `actualType`, right?

converter.go
sfc-gh-pfus: Why? This is perfectly normal situation, that we have null value from DB.

converter.go
sfc-gh-pfus: I was considering it, but each function has different arguments... I wouldn't introduce it. At the end I'd like to make some refactoring (if I found some time), maybe I will be able to clean it up a bit.

sfc-gh-pfus: Ok, done. Needed to change the whole name though.

structured_type_test.go
sfc-gh-pfus: Actually no. This is a placeholder for actual value.

converter.go
sfc-gh-dheyman-1: Ok.

sfc-gh-dheyman-1: Sure, thanks.

sfc-gh-dheyman-1 (approved):

PR #1100, opened 2024-04-15
fix panic when passing nil to time.Time.In()

madisonchamberlain:
### Description

When time.Time.In is used, if a nil is passed in it will panic. ([See docs](https://www.geeksforgeeks.org/time-time-in-function-in-golang-with-examples/#:~:text=for%20display%20uses-,.%20And%20if%20%E2%80%9Cloc%E2%80%9D%20is%20nil%20then%20panic%20is%20returned.,-Pause)) In the other functions in converter.go (see stringToValue timestamp_ltz) if the loc is nil a default will be used. 

* Bug was introduced in [this](https://github.com/snowflakedb/gosnowflake/commit/3a295547f478bda564c3943b5e1f5d61a525437b) pr

### Checklist

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -674,7 +674,7 @@ func (scd *snowflakeArrowStreamChunkDownloader) Location() *time.Location {
 	if scd.sc != nil && scd.sc.cfg != nil {
 		return getCurrentLocation(scd.sc.cfg.Params)
 	}
-	return nil
+	return time.Now().Location()
 }
 func (scd *snowflakeArrowStreamChunkDownloader) TotalRows() int64 { return scd.Total }
 func (scd *snowflakeArrowStreamChunkDownloader) RowTypes() []execResponseRowType {
--- a/connection.go
+++ b/connection.go
@@ -674,7 +674,7 @@ func (scd *snowflakeArrowStreamChunkDownloader) Location() *time.Location {
 	if scd.sc != nil && scd.sc.cfg != nil {
 		return getCurrentLocation(scd.sc.cfg.Params)
 	}
-	return time.UTC
+	return nil
 }
 func (scd *snowflakeArrowStreamChunkDownloader) TotalRows() int64 { return scd.Total }
 func (scd *snowflakeArrowStreamChunkDownloader) RowTypes() []execResponseRowType {
--- a/connection.go
+++ b/connection.go
@@ -671,10 +671,10 @@ type snowflakeArrowStreamChunkDownloader struct {
 }
 
 func (scd *snowflakeArrowStreamChunkDownloader) Location() *time.Location {
-	if scd.sc != nil {
+	if scd.sc != nil && scd.sc.cfg != nil {
 		return getCurrentLocation(scd.sc.cfg.Params)
 	}
-	return nil
+	return time.UTC
 }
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -671,7 +671,7 @@ type snowflakeArrowStreamChunkDownloader struct {
 }
 
 func (scd *snowflakeArrowStreamChunkDownloader) Location() *time.Location {
-	if scd.sc != nil {
+	if scd.sc != nil && scd.sc.cfg != nil {
 		return getCurrentLocation(scd.sc.cfg.Params)
 	}
```
rickr-sigma, line 676: is it possible that Params is empty or something that will return a bad value for time.Location?

converter.go
```diff
@@ -316,6 +316,9 @@ func stringToValue(
 		}
 		loc := Location(int(offset) - 1440)
 		tt := time.Unix(sec, nsec)
+		if loc == nil {
+			loc = time.Now().Location()
+		}
```
rickr-sigma, line 321: should these just be time.Location?  i.e. avoid the Now() call.

rickr-sigma (approved):

connection.go
madisonchamberlain: if getCurrentLocation is called something non nil will be returned, updated the else option to also return time.Now.Location so we dont ever return nil

converter.go
madisonchamberlain: afaik time.Location{} if not generated by time.Now() is an empty struct which will panic without the name string

connection.go
```diff
@@ -671,10 +671,10 @@ type snowflakeArrowStreamChunkDownloader struct {
 }
 
 func (scd *snowflakeArrowStreamChunkDownloader) Location() *time.Location {
-	if scd.sc != nil {
+	if scd.sc != nil && scd.sc.cfg != nil {
```
sfc-gh-pfus, line 674: This change is not a mistake, but I'm wondering - you actually encountered this situation? By looking at the code I can see that when `snowflakeConn` is initialised, it is always initialised with not nil config.

connection.go
```diff
@@ -671,10 +671,10 @@ type snowflakeArrowStreamChunkDownloader struct {
...
 		return getCurrentLocation(scd.sc.cfg.Params)
 	}
-	return nil
+	return time.Now().Location()
```
sfc-gh-pfus, line 677: Similarly - looking at the code, we should never reach this line, scd.sc is always initalised...

This line is much more tricky anyway. This location may be completely different than snowflake session timezone, meaning it may break LTZ contracts.

converter.go
sfc-gh-pfus: `loc` is initialized two lines above - so this condition is always false, am I right?

connection.go
madisonchamberlain: I have not experienced this, I can remove it I just wanted to ensure that we dont null pointer dereference in any possible way

connection.go
madisonchamberlain: Ideally we should never hit this line but it seems that we are hitting it. I personally feel like this is a good temporary fix until we figure out why the connection is nil because having a panic seems worse to me but I am open to any other solutions

PR #1103, opened 2024-04-17
SNOW-1313544 temporal credential cache map race

sfc-gh-dszmolka:
SNOW-1313544 Parallel operations on different roles cause program to crash (race condition on temporal credential cache)
Discovered in a Terraform use-case, where there's a lot of resources handled by the Snowflake Terraform Provider 

* moving locks a bit ahead (delete on map happened before lock)
* write-lock instead of read-lock when reading the credentials

### DIFF
--- a/secure_storage_manager.go
+++ b/secure_storage_manager.go
@@ -190,13 +190,13 @@ func readTemporaryCredential(sc *snowflakeConn, credType string) string {
 // Writes to temporary credential file when OS is Linux.
 func writeTemporaryCredential(sc *snowflakeConn, credType, token string) {
 	target := convertTarget(sc.cfg.Host, sc.cfg.User, credType)
+	temporaryCredCacheLock.Lock()
 	localCredCache[target] = token
 
 	j, err := json.Marshal(localCredCache)
 	if err != nil {
 		logger.Debugf("failed to convert credential to JSON.")
 	}
-	temporaryCredCacheLock.Lock()
 	writeTemporaryCacheFile(j)
 	temporaryCredCacheLock.Unlock()
 }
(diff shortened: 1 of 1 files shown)


secure_storage_manager.go
```diff
@@ -190,30 +190,30 @@ func readTemporaryCredential(sc *snowflakeConn, credType string) string {
 // Writes to temporary credential file when OS is Linux.
 func writeTemporaryCredential(sc *snowflakeConn, credType, token string) {
 	target := convertTarget(sc.cfg.Host, sc.cfg.User, credType)
+	credCacheLock.Lock()
+	defer credCacheLock.Unlock()
 	localCredCache[target] = token
 
 	j, err := json.Marshal(localCredCache)
 	if err != nil {
 		logger.Debugf("failed to convert credential to JSON.")
```
sfc-gh-pfus, line 199: I think we should return here. After error occurs, we don't have proper `j`.

secure_storage_manager.go
```diff
@@ -190,30 +190,30 @@ func readTemporaryCredential(sc *snowflakeConn, credType string) string {
...
 	if credCacheDir == "" {
 		logger.Debug("Cache file doesn't exist. Skipping deleting credential file.")
 	} else {
+		credCacheLock.Lock()
+		defer credCacheLock.Unlock()
 		target := convertTarget(sc.cfg.Host, sc.cfg.User, credType)
 		delete(localCredCache, target)
 		j, err := json.Marshal(localCredCache)
 		if err != nil {
 			logger.Debugf("failed to convert credential to JSON.")
```
sfc-gh-pfus, line 214: Same here. Also, maybe we can log as warn?

sfc-gh-dszmolka: addd return to the error scenario + logging it as Warn instead of Debug

sfc-gh-pfus (approved):

PR #1111, opened 2024-04-19
SNOW-1264786 Support structured types in arrow batches

sfc-gh-pfus:
### Description

SNOW-1264786 Support for structured types in arrow batches mode

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -2124,11 +2124,8 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			numberOfNulls := structCol.NullN()
 			return array.NewStructArrayWithNulls(internalCols, fieldNames, nullBitmap, numberOfNulls, 0)
 		}
-		return col, nil
 	case arrayType:
-		if _, ok := col.(*array.List); ok {
-			listCol := col.(*array.List)
-			defer listCol.Release()
+		if listCol, ok := col.(*array.List); ok {
 			newCol, err = arrowToRecordSingleColumn(ctx, field.Type.(*arrow.ListType).ElemField(), listCol.ListValues(), fieldMetadata.Fields[0], higherPrecisionEnabled, timestampOption, pool, loc, numRows)
 			if err != nil {
 				return nil, err
--- a/converter.go
+++ b/converter.go
@@ -1966,200 +1966,212 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
 	ctxAlloc := compute.WithAllocator(ctx, pool)
 
 	for i, col := range record.Columns() {
-		srcColumnMeta := rowType[i]
+		fieldMetadata := rowType[i].toFieldMetadata()
 
-		// TODO: confirm that it is okay to be using higher precision logic for conversions
-		newCol := col
-		snowflakeType := getSnowflakeType(srcColumnMeta.Type)
-		switch snowflakeType {
-		case fixedType:
+		newCol, err := arrowToRecordSingleColumn(ctxAlloc, s.Field(i), col, fieldMetadata, higherPrecisionEnabled, arrowBatchesTimestampOption, pool, loc, numRows)
+		if err != nil {
+			return nil, err
+		}
+		cols = append(cols, newCol)
+	}
+	newRecord := array.NewRecord(s, cols, numRows)
+	for _, col := range newRecord.Columns() {
+		col.Release()
+	}
+	return newRecord, nil
+}
+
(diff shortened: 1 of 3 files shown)


structured_type_test.go
```diff
@@ -1345,6 +1349,237 @@ func TestWithHigherPrecision(t *testing.T) {
...
+		assertNotEqualF(t, len(batches), 0)
+		batch, err := batches[0].Fetch()
+		assertNilF(t, err)
+		assertNotEqualF(t, len(*batch), 0)
+		record := (*batch)[0]
+		defer record.Release()
+		assertEqualE(t, record.Column(0).(*array.List).ListValues().(*array.Int64).Value(0), int64(1))
+		assertEqualE(t, record.Column(0).(*array.List).ListValues().(*array.Int64).Value(1), int64(2))
+		assertEqualE(t, record.Column(0).(*array.List).ListValues().(*array.Int64).Value(2), int64(3))
+		assertEqualE(t, record.Column(0).(*array.List).ListValues().(*array.Int64).Value(3), int64(4))
```
sfc-gh-dstempniak, line 1545: Here is something strange.
We have 2 records with arrays of 3.
How we can read values for the second record only (indices 3, 4, 5 here) by row number? 
It looks like all arrays from the column are joined into one single array.
I thinks it is not ok.

converter.go
```diff
@@ -1966,200 +1966,216 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
...
+
+func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow.Array, fieldMetadata fieldMetadata, higherPrecisionEnabled bool, timestampOption snowflakeArrowBatchesTimestampOption, pool memory.Allocator, loc *time.Location, numRows int64) (arrow.Array, error) {
+	var err error
+	newCol := col
+	snowflakeType := getSnowflakeType(fieldMetadata.Type)
+	switch snowflakeType {
+	case fixedType:
+		if higherPrecisionEnabled {
+			// do nothing - return decimal as is
+			col.Retain()
```
sfc-gh-dstempniak, line 1990: Somehow I feel like this is going to cause problems

converter.go
```diff
@@ -1966,200 +1966,216 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
...
-			default:
-				if withHigherPrecision {
-					converted = false
-				} else if srcColumnMeta.Scale != 0 {
-					t = &arrow.Float64Type{}
-				} else {
-					converted = false
+	case objectType:
+		if structCol, ok := col.(*array.Struct); ok {
+			var internalCols []arrow.Array
```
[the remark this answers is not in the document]
sfc-gh-pfus: That is a very good question, but the answer is - no. It is going to be released [here](https://github.com/snowflakedb/gosnowflake/blob/3ffb2874ab51206d3182dd0243a4f56919723ba1/arrow_chunk.go#L58), with cleaning up `rawRecord`.

converter.go
```diff
@@ -1966,200 +1966,216 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
...
-				t = &arrow.TimestampType{Unit: arrow.Millisecond, TimeZone: loc.String()}
-			} else if timestampOption == UseSecondTimestamp {
-				t = &arrow.TimestampType{Unit: arrow.Second, TimeZone: loc.String()}
-			} else {
-				t = &arrow.TimestampType{Unit: arrow.Nanosecond, TimeZone: loc.String()}
+			nullBitmap := memory.NewBufferBytes(structCol.NullBitmapBytes())
+			numberOfNulls := structCol.NullN()
+			return array.NewStructArrayWithNulls(internalCols, fieldNames, nullBitmap, numberOfNulls, 0)
+		}
+		return col, nil
```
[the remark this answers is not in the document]
sfc-gh-pfus: We can omit it, but it is a bit descriptive - here, we intentionally return original column without changes.

converter.go
sfc-gh-pfus: Can you elaborate a bit? This code changed, retaining/releasing changed as well. Now, we release at the end all columns (we can't use defer because we need more than function scope). But this column is not going to be changed, so it would be released twice (with old record and new one). So we have to bump it's counter.

converter.go
```diff
@@ -1966,200 +1966,216 @@ func arrowToRecord(ctx context.Context, record arrow.Record, pool memory.Allocat
...
+	case arrayType:
+		if _, ok := col.(*array.List); ok {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Yes, we can!

structured_type_test.go
sfc-gh-pfus: Yes, this is because of Arrow Go API. But you are right - I added a test for checking offsets.

sfc-gh-dstempniak (approved):

PR #1124, opened 2024-05-07
SNOW-1256899 Add structured types docs

sfc-gh-pfus:
### Description

SNOW-1256899 Added docs of structured types usage.

### Checklist

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -391,11 +391,13 @@ data types. The columns are:
     -------------------------------------------------------------------------------------------------------------------
     BINARY               | []byte                                      | string                 | []byte
     -------------------------------------------------------------------------------------------------------------------
-    ARRAY                | string                                      | string
+    ARRAY [6]            | string / array                              | string / array
     -------------------------------------------------------------------------------------------------------------------
-    OBJECT               | string                                      | string
+    OBJECT [6]           | string / struct                             | string / struct
     -------------------------------------------------------------------------------------------------------------------
     VARIANT              | string                                      | string
+    -------------------------------------------------------------------------------------------------------------------
+    MAP                  | map                                         | map
 
     [1] Converting from a higher precision data type to a lower precision data type via the snowflakeRows.Scan()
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -412,8 +414,74 @@ data types. The columns are:
...
 
+    [6] Arrays and objects can be either semistructured or structured, see more info in section below.
+
 Note: SQL NULL values are converted to Golang nil values, and vice-versa.
 
+## Semistructured and structured types
+
+Snowflake supports two flavours of "structured data" - semistructured and structured.
+Semistructured types are variants, objects and arrays without schema.
+These types are represented by strings and should be interpreted on the client side.
```
sfc-gh-dheyman-1, line 425: I think this makes it more clear `When data is fetched, it's represented as strings and the client is responsible for its interpretation.`

doc.go
```diff
@@ -412,8 +414,74 @@ data types. The columns are:
...
+Example table definition:
+
+	CREATE TABLE semistructured (v VARIANT, o OBJECT, a ARRAY)
+
+In all rows of the table values may be slightly different.
```
sfc-gh-dheyman-1, line 430: `The data not have any corresponding schema, so values in table may be slightly different. `

doc.go
```diff
@@ -412,8 +414,74 @@ data types. The columns are:
...
+3. Use it in regular scan:
+
+	var res simpleObject
+	err := rows.Scan(&res)
+
+See StructuredObject for all available operations including null support, embedding nested structs, etc.
+
+#### Retrieving structured arrays
+
+Retrieving array of simple types works exactly the same like normal values - using scan.
```
sfc-gh-dheyman-1, line 472: `(...) using Scan method.`

doc.go
sfc-gh-pfus: Done.

doc.go
sfc-gh-pfus: Done.

doc.go
sfc-gh-pfus: Done.

PR #1134, opened 2024-05-17
[bug fix] objectType should work w/ or wo/ ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE

madisonchamberlain:
### Description

If we get a snowflake object type, but we haven't set the flag `ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE`, the code will panic because we are only setting the record on the batch if the snowflake type matches the expected arrow type. This will add support for the underlying type to be either structured or string for snowflake objectType, and if its neither, we will return an error with the underlying type rather than panic so we can fix it

### Checklist
   - `TestStructuredTypeInArrowBatchesWithoutEnablingStructureTypes`: check that we can have an object type without setting structured types flag
   - `TestNullObject`: ensure that we can have a null object. (This is what was failing in my test suite; ultimately this isn't the actual trigger of the panic, but its still good to test because it could be one day)
- [NA] Extended the README / documentation, if necessary

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -2023,6 +2023,8 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			builder.AppendValues(floatValues, nil)
 			newCol = builder.NewArray()
 			builder.Release()
+		} else {
+			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake mapType", col)
 		}
 	case timeType:
 		newCol, err = compute.CastArray(ctx, col, compute.SafeCastOptions(arrow.FixedWidthTypes.Time64ns))
@@ -2085,10 +2087,7 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 		}
 	case textType:
 		if stringCol, ok := col.(*array.String); ok {
-			newValidUtf8Array := arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
-			if newValidUtf8Array != nil {
-				newCol = newValidUtf8Array
-			}
+			newCol = arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
 		} else {
 			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake textType", col)
 		}
@@ -2112,10 +2111,7 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			numberOfNulls := structCol.NullN()
 			return array.NewStructArrayWithNulls(internalCols, fieldNames, nullBitmap, numberOfNulls, 0)
 		} else if stringCol, ok := col.(*array.String); ok {
-			newValidUtf8Array := arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
-			if newValidUtf8Array != nil {
-				newCol = newValidUtf8Array
-			}
+			newCol = arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
 		} else {
 			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake objectType", col)
 		}
@@ -2130,42 +2126,44 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			defer newData.Release()
 			return array.NewListData(newData), nil
 		} else if stringCol, ok := col.(*array.String); ok {
-			newValidUtf8Array := arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
-			if newValidUtf8Array != nil {
-				newCol = newValidUtf8Array
-			}
+			newCol = arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
 		} else {
 			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake objectType", col)
 		}
 	case mapType:
-		mapCol := col.(*array.Map)
-		keyCol, err := arrowToRecordSingleColumn(ctx, field.Type.(*arrow.MapType).KeyField(), mapCol.Keys(), fieldMetadata.Fields[0], higherPrecisionEnabled, timestampOption, pool, loc, numRows)
-		if err != nil {
-			return nil, err
-		}
-		defer keyCol.Release()
-		valueCol, err := arrowToRecordSingleColumn(ctx, field.Type.(*arrow.MapType).ItemField(), mapCol.Items(), fieldMetadata.Fields[1], higherPrecisionEnabled, timestampOption, pool, loc, numRows)
-		if err != nil {
-			return nil, err
-		}
-		defer valueCol.Release()
+		if mapCol, ok := col.(*array.Map); ok {
+			keyCol, err := arrowToRecordSingleColumn(ctx, field.Type.(*arrow.MapType).KeyField(), mapCol.Keys(), fieldMetadata.Fields[0], higherPrecisionEnabled, timestampOption, pool, loc, numRows)
+			if err != nil {
+				return nil, err
+			}
+			defer keyCol.Release()
+			valueCol, err := arrowToRecordSingleColumn(ctx, field.Type.(*arrow.MapType).ItemField(), mapCol.Items(), fieldMetadata.Fields[1], higherPrecisionEnabled, timestampOption, pool, loc, numRows)
+			if err != nil {
+				return nil, err
+			}
+			defer valueCol.Release()
 
-		structArr, err := array.NewStructArray([]arrow.Array{keyCol, valueCol}, []string{"k", "v"})
-		if err != nil {
-			return nil, err
+			structArr, err := array.NewStructArray([]arrow.Array{keyCol, valueCol}, []string{"k", "v"})
+			if err != nil {
+				return nil, err
+			}
+			defer structArr.Release()
+			newData := array.NewData(arrow.MapOf(keyCol.DataType(), valueCol.DataType()), mapCol.Len(), mapCol.Data().Buffers(), []arrow.ArrayData{structArr.Data()}, mapCol.NullN(), 0)
+			defer newData.Release()
+			return array.NewMapData(newData), nil
+		} else if stringCol, ok := col.(*array.String); ok {
+			newCol = arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
+		} else {
+			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake mapType", col)
 		}
-		defer structArr.Release()
-		newData := array.NewData(arrow.MapOf(keyCol.DataType(), valueCol.DataType()), mapCol.Len(), mapCol.Data().Buffers(), []arrow.ArrayData{structArr.Data()}, mapCol.NullN(), 0)
-		defer newData.Release()
-		return array.NewMapData(newData), nil
 	default:
 		col.Retain()
 	}
 	return newCol, nil
 }
 
-// returns a pointer to a new array which will be populated if we converted the array to valid utf8
-// or returns null, and an indicator that we should retain the original column
+// returns n arrow array which will be new and populated if we converted the array to valid utf8
+// or if we didn't covnert it, it will return the original column.
 func arrowStringRecordToColumn(
 	ctx context.Context,
 	stringCol *array.String,
@@ -2192,8 +2190,9 @@ func arrowStringRecordToColumn(
 		arr := tb.NewArray()
 		return arr
 	}
-	stringCol.Retain()
-	return nil
+	var asGenericArray arrow.Array = stringCol
+	asGenericArray.Retain()
+	return asGenericArray
 }
 
 func recordToSchema(sc *arrow.Schema, rowType []execResponseRowType, loc *time.Location, timestampOption snowflakeArrowBatchesTimestampOption, withHigherPrecision bool) (*arrow.Schema, error) {
--- a/converter.go
+++ b/converter.go
@@ -2140,6 +2140,15 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			newData := array.NewData(arrow.ListOf(newCol.DataType()), listCol.Len(), listCol.Data().Buffers(), []arrow.ArrayData{newCol.Data()}, listCol.NullN(), 0)
 			defer newData.Release()
 			return array.NewListData(newData), nil
+		} else if stringCol, ok := col.(*array.String); ok {
+			newValidUtf8Array, shouldRetainOriginalCol := arrowStringRecordToColumn(ctx, stringCol, pool, numRows, fieldMetadata)
+			if shouldRetainOriginalCol == shouldRetain {
+				col.Retain()
+			} else {
+				newCol = *newValidUtf8Array
+			}
+		} else {
+			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake objectType", col)
 		}
 	case mapType:
 		mapCol := col.(*array.Map)
--- a/converter.go
+++ b/converter.go
@@ -2123,6 +2123,27 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			nullBitmap := memory.NewBufferBytes(structCol.NullBitmapBytes())
 			numberOfNulls := structCol.NullN()
 			return array.NewStructArrayWithNulls(internalCols, fieldNames, nullBitmap, numberOfNulls, 0)
+		} else if stringCol, ok := col.(*array.String); ok {
+			if arrowBatchesUtf8ValidationEnabled(ctx) && col.DataType().ID() == arrow.STRING {
+				tb := array.NewStringBuilder(pool)
+				defer tb.Release()
+
+				for i := 0; i < int(numRows); i++ {
+					if stringCol.IsValid(i) {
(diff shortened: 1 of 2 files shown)


converter.go
```diff
@@ -2123,6 +2123,29 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			nullBitmap := memory.NewBufferBytes(structCol.NullBitmapBytes())
 			numberOfNulls := structCol.NullN()
 			return array.NewStructArrayWithNulls(internalCols, fieldNames, nullBitmap, numberOfNulls, 0)
+		} else if stringCol, ok := col.(*array.String); ok {
+			if arrowBatchesUtf8ValidationEnabled(ctx) && col.DataType().ID() == arrow.STRING {
+				tb := array.NewStringBuilder(pool)
```
sfc-gh-pfus, line 2128: It looks the same as in `case textType`, can we extract function here?

structured_type_test.go
```diff
@@ -1621,3 +1621,64 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+	skipStructuredTypesTestsOnGHActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		pool := memory.NewCheckedAllocator(memory.DefaultAllocator)
+		defer pool.AssertSize(t, 0)
+		ctx := WithArrowBatches(WithArrowAllocator(context.Background(), pool))
+
+		var err error
+		var rows driver.Rows
+		err = dbt.conn.Raw(func(sc any) error {
+			rows, err = sc.(driver.QueryerContext).QueryContext(ctx, "select null::object", nil)
```
sfc-gh-pfus, line 1635: Can we have also `null::object(v VARCHAR)`?

structured_type_test.go
```diff
@@ -1621,3 +1621,64 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+		assertNilF(t, err)
+		defer rows.Close()
+		batches, err := rows.(SnowflakeRows).GetArrowBatches()
+		assertNilF(t, err)
+		assertNotEqualF(t, len(batches), 0)
+		batch, err := batches[0].Fetch()
+		assertNilF(t, err)
+		assertNotEqualF(t, len(*batch), 0)
+		for _, record := range *batch {
+			assertEqualE(t, record.Column(0).(*array.String).Value(0), "")
```
sfc-gh-pfus, line 1647: Maybe it is worth also checking `record.Column(0).(*array.String).IsNull(0)`, what do you think?

structured_type_test.go
```diff
@@ -1621,3 +1621,64 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+			record.Release()
+		}
+	})
+}
+
+// Want to make sure this still works if we dont set ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE
```
sfc-gh-pfus, line 1653: Actually this is not the reason. The reason is that you use `object` without schema. IMHO this comment is not needed, can we remove it at all?

structured_type_test.go
```diff
@@ -1621,3 +1621,64 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+// AFAIK thats not fully rolledout yet, I cant set it on our server at the moment
+func TestStructuredTypeInArrowBatchesWithoutEnablingStructureTypes(t *testing.T) {
```
sfc-gh-pfus, line 1655: Maybe `TestStructuredTypeInArrowBatchesWithSemistructuredObject`?

structured_type_test.go
```diff
@@ -1621,3 +1621,64 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+	skipStructuredTypesTestsOnGHActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		pool := memory.NewCheckedAllocator(memory.DefaultAllocator)
+		defer pool.AssertSize(t, 0)
+		ctx := WithArrowBatches(WithArrowAllocator(context.Background(), pool))
+
+		var err error
+		var rows driver.Rows
+		err = dbt.conn.Raw(func(sc any) error {
+			rows, err = sc.(driver.QueryerContext).QueryContext(ctx, "SELECT 1, {'s':'someString'}::OBJECT(s VARCHAR)", nil)
```
sfc-gh-pfus, line 1665: Having said that, I believe that this test should use `SELECT {'s':'someString'}::OBJECT`

converter.go
madisonchamberlain: good idea, done

structured_type_test.go
madisonchamberlain: yes, now the test checks that both will work here

structured_type_test.go
madisonchamberlain: good idea, updated

structured_type_test.go
madisonchamberlain: Yes, comment has been removed!

structured_type_test.go
madisonchamberlain: done!

structured_type_test.go
madisonchamberlain: Updated! I am now running both queries, the original and the one you suggested just for more coverage

converter.go
```diff
@@ -58,6 +58,13 @@ const (
 	UseOriginalTimestamp
 )
 
+type retainOriginalCol bool
```
sfc-gh-pfus, line 61: What do you think about alternative solution - instead of returning information about retaining/reusing column, maybe we can instead always return column (whatever it is - transformed or original) and call retain in `arrowStringRecordToColumn` if needed? The code should be simpler and responsibility is on transforming function and not caller (easier to omit!). What do you think?

converter.go
```diff
@@ -2162,6 +2168,38 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
...
 
+// returns a pointer to a new array which will be populated if we converted the array to valid utf8
+// or returns null, and an indicator that we should retain the original column
+func arrowStringRecordToColumn(
+	ctx context.Context,
+	stringCol *array.String,
+	mem memory.Allocator,
+	numRows int64,
+	fieldMetadata fieldMetadata,
+) (*arrow.Array, retainOriginalCol) {
```
sfc-gh-pfus, line 2179: If I remember correctly - we don't need to return pointer here. Arrow already handles memory efficiently, and returning `arrow.Array` should not do memory copying.

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
 }
+
+func TestNullObject(t *testing.T) {
+	skipStructuredTypesTestsOnGHActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		nullObjectQueries := []string{
+			"select null::object(v VARCHAR)",
+			"select null::object",
+		}
+		for i := 0; i < len(nullObjectQueries); i++ {
```
sfc-gh-pfus, line 1634: Can we use `t.Run` here? See other examples in our repo :)

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
 		})
 	}
 }
+
+func TestNullObject(t *testing.T) {
```
sfc-gh-pfus, line 1627: Names are global here, so maybe we can have something a bit more specific? Maybe `TestSelectingNullObjectsInArrowBatches`?

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+		}
+	})
+}
+
+func TestStructuredTypeInArrowBatchesWithSemistructuredObject(t *testing.T) {
+	skipStructuredTypesTestsOnGHActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		objectQueries := []string{
+			"SELECT 1, {'s':'someString'}::OBJECT(s VARCHAR)", // returns string
+			"SELECT {'s':'someString'}::OBJECT",               // returns binary
```
sfc-gh-pfus, line 1686: Actually this also returns string :)

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+			}
+		}
+	})
+}
+
+func TestStructuredTypeInArrowBatchesWithSemistructuredObject(t *testing.T) {
+	skipStructuredTypesTestsOnGHActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		objectQueries := []string{
+			"SELECT 1, {'s':'someString'}::OBJECT(s VARCHAR)", // returns string
```
sfc-gh-pfus, line 1685: This case is not good... It will start failing when structured types are publicly available, because format of response will change. Also, this is covered by `TestStructuredTypeInArrowBatchesSimple`, so maybe it is not needed?

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+				assertEqualE(t, srtCol.Value(rowIndex), "")
+				isNull := srtCol.IsNull(rowIndex)
+				assertTrueF(t, isNull)
+				record.Release()
+			}
+		}
+	})
+}
+
+func TestStructuredTypeInArrowBatchesWithSemistructuredObject(t *testing.T) {
```
sfc-gh-pfus, line 1681: Having a comment below - maybe it is better to call this test `TestSelectingSemistructuredObjectInArrowBatches`? And instead of a duplicated test for structured object - maybe we can have a separate test for semistructured arrays like `SELECT [1, 2, 3]::ARRAY`?

structured_type_test.go
```diff
@@ -1621,3 +1623,135 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+				assertTrueF(t, isSfRows, "rows are not snowflakeRows")
+
+				batches, err := sfRows.GetArrowBatches()
+				assertNilF(t, err)
+				assertNotEqualF(t, len(batches), 0)
+				batch, err := batches[0].Fetch()
+				assertNilF(t, err)
+				assertNotEqualF(t, len(*batch), 0)
+				for _, record := range *batch {
+					// if we are running the first query there should be two columns, otherwise one
```
sfc-gh-pfus, line 1725: First column can be safely removed from query, the test should simplify then.

structured_type_test.go
madisonchamberlain: Done!

structured_type_test.go
madisonchamberlain: done

structured_type_test.go
madisonchamberlain: Oh yes, meant to delete that! Thanks

structured_type_test.go
madisonchamberlain: I suppose the difference is also that the original tests sets the `ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE` flag and this test does not. In theory we should be able to just run the new test I added with and without the flag value, but I cant confirm because our test snowflake server wont let us set the flag. Should we combine them? I just cant verify the test runs on my end

structured_type_test.go
madisonchamberlain: Done, thanks

converter.go
madisonchamberlain: Confirmed that you are correct, updated to not return a pointer

structured_type_test.go
madisonchamberlain: I do think its good to have `SELECT [1, 2, 3]::ARRAY` and `SELECT {'s':'someString'}::OBJECT` because both of these fail on different branches of `arrowToRecordSingleColumn` (so good call to add this case because I had not fixed it!), are you ok with me leaving both? I think its good to have the higher code coverage

converter.go
```diff
@@ -2162,6 +2164,38 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
...
+				tb.Append(stringValue)
+			} else {
+				tb.AppendNull()
+			}
+		}
+		arr := tb.NewArray()
+		return arr
+	}
+	stringCol.Retain()
+	return nil
```
sfc-gh-pfus, line 2196: If we return `stringCol` here instead of `nil` we can omit nil checking when this method is called, what do you think?

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
 		})
 	}
 }
+
+func TestSelectingNullObjectsInArrowBatches(t *testing.T) {
+	testcases := []struct {
+		name  string
```
sfc-gh-pfus, line 1629: I think we can omit this struct, we can just have:

```
testcases := []string{
  "query 1",
  "query 2",
}
```
And use it in name.

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+				batches, err := sfRows.GetArrowBatches()
+				assertNilF(t, err)
+				assertNotEqualF(t, len(batches), 0)
+				batch, err := batches[0].Fetch()
+				assertNilF(t, err)
+				assertNotEqualF(t, len(*batch), 0)
+				for _, record := range *batch {
+					// check number of cols/rows so we dont get index out of range
+					assertEqualF(t, record.NumRows(), int64(1), "wrong number of rows")
+					assertEqualF(t, record.NumRows(), int64(1), "wrong number of cols")
```
sfc-gh-pfus, line 1672: `NumRows` or something related to columns? :)

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+
+func TestSelectingSemistructuredObjectInArrowBatches(t *testing.T) {
+	testcases := []struct {
+		name               string
+		query              string
+		expected           string
+		withUtf8Validation bool
+	}{
+		{
+			name:               "test structured type in arrow batch with utf8 validation, snowflakeType = objectType",
```
sfc-gh-pfus, line 1699: `structured type` -> `semistructured object`?

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+			withUtf8Validation: true,
+			expected:           `{"s":"someString"}`,
+			query:              "SELECT {'s':'someString'}::OBJECT",
+		},
+		{
+			name:               "test structured type in arrow batch without utf8 validation, snowflakeType = objectType",
```
sfc-gh-pfus, line 1705: As above.

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+			withUtf8Validation: false,
+			expected:           `{"s":"someString"}`,
+			query:              "SELECT {'s':'someString'}::OBJECT",
+		},
+		{
+			name:               "test structured type in arrow batch without utf8 validation, snowflakeType = arrayType",
```
sfc-gh-pfus, line 1711: `structured type` -> `semistructured array`?

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+			withUtf8Validation: false,
+			expected:           `[1,2,3]`,
+			query:              "SELECT [1, 2, 3]::ARRAY",
+		},
+		{
+			name:               "test structured type in arrow batch with utf8 validation, snowflakeType = arrayType",
+			withUtf8Validation: true,
+			expected:           `[1,2,3]`,
+			query:              "SELECT [1, 2, 3]::ARRAY",
```
sfc-gh-pfus, line 1720: As above.

structured_type_test.go
```diff
@@ -1621,3 +1623,151 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+					assertEqualF(t, record.NumCols(), int64(1), "unexpected number of columns")
+					assertEqualF(t, record.NumRows(), int64(1), "unexpected number of rows")
+
+					curColIndex := 0
+					rowIndex := 0
+
+					strCol, isStr := record.Column(curColIndex).(*array.String)
+					assertTrueF(t, isStr, "wrong type for column")
+
+					actual := strings.ReplaceAll(strings.ReplaceAll(strCol.Value(rowIndex), " ", ""), "\n", "")
```
sfc-gh-pfus: There is a brand new (from today!) function called `assertEqualIgnoringWhitespace` - may be useful!

structured_type_test.go
sfc-gh-pfus: Just to explain - this parameter is required for structured types. As long as you use semistructured types (like `::OBJECT` without schema instead of `::OBJECT(v VARCHAR, i INTEGER)` with schema) this parameter does not matter. But I like the tests in current version, so I'm resolving this comment!

structured_type_test.go
madisonchamberlain: Oh cool! I will rebase and use it

madisonchamberlain:
I cant get this to correctly rebase off my fork so I moved it all here https://github.com/snowflakedb/gosnowflake/pull/1137

converter.go
madisonchamberlain: Done!

PR #1137, opened 2024-05-21
[bug fix] objectType should work w/ or wo/ ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE 

madisonchamberlain:
### Description

If we get a snowflake object type, but we haven't set the flag `ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE`, the code will panic because we are only setting the record on the batch if the snowflake type matches the expected arrow type. This will add support for the underlying type to be either structured or string for snowflake objectType, and if its neither, we will return an error with the underlying type rather than panic so we can fix it

### Checklist
   - `TestStructuredTypeInArrowBatchesWithoutEnablingStructureTypes`: check that we can have an object type without setting structured types flag
   - `TestNullObject`: ensure that we can have a null object. (This is what was failing in my test suite; ultimately this isn't the actual trigger of the panic, but its still good to test because it could be one day)
- [NA] Extended the README / documentation, if necessary

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -2023,6 +2023,8 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			builder.AppendValues(floatValues, nil)
 			newCol = builder.NewArray()
 			builder.Release()
+		} else {
+			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake mapType", col)
 		}
 	case timeType:
 		newCol, err = compute.CastArray(ctx, col, compute.SafeCastOptions(arrow.FixedWidthTypes.Time64ns))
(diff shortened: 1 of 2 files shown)


converter.go
```diff
@@ -2023,6 +2023,8 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
 			builder.AppendValues(floatValues, nil)
 			newCol = builder.NewArray()
 			builder.Release()
+		} else {
+			return nil, fmt.Errorf("unsupported arrow type %T when trying to convert a snowflake mapType", col)
```
sfc-gh-pfus, line 2027: It's rather `fixedType`?

converter.go
```diff
@@ -2134,34 +2125,76 @@ func arrowToRecordSingleColumn(ctx context.Context, field arrow.Field, col arrow
...
+				}
+				tb.Append(stringValue)
+			} else {
+				tb.AppendNull()
+			}
+		}
+		arr := tb.NewArray()
+		return arr
+	}
+	var asGenericArray arrow.Array = stringCol
```
sfc-gh-pfus, line 2193: Can't we just `return stringCol`?

structured_type_read_test.go
```diff
@@ -1621,3 +1622,156 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+					assertTrueF(t, isNull)
+					record.Release()
+				}
+
+			})
+		}
+	})
+}
+
+func TestSelectingSemistructuredObjectInArrowBatches(t *testing.T) {
```
sfc-gh-pfus, line 1681: If a test contains arrays as well as objects, maybe we can call it `semistructuredTypes`?

structured_type_read_test.go
```diff
@@ -1621,3 +1622,156 @@ func forAllStructureTypeFormats(dbt *DBTest, f func(t *testing.T, format string)
...
+				for _, record := range *batch {
+					assertEqualF(t, record.NumCols(), int64(1), "unexpected number of columns")
+					assertEqualF(t, record.NumRows(), int64(1), "unexpected number of rows")
+
+					curColIndex := 0
+					rowIndex := 0
+
+					// The underlying data may be struct, or it could be string. Either way is ok but lets not fail
+					stringCol, isString := record.Column(curColIndex).(*array.String)
+					structCol, isStruct := record.Column(curColIndex).(*array.Struct)
```
sfc-gh-pfus, line 1754: Is this case possible? Test cases (and test name!) suggest that there are only semistructured types, should always be strings.

madisonchamberlain: Yeah it should not be possible but I didnt want to block development in the case that there was some case where the server could return a struct even though it was supposed to be a string. But yeah not needed, I will remove it

structured_type_read_test.go
madisonchamberlain: Updated!

converter.go
madisonchamberlain: yes, done

sfc-gh-pfus (approved):

PR #1146, opened 2024-05-27
SNOW-1432112 verify / sanitize account name before creating JWT

sfc-gh-dszmolka:
Attempt to fix the issue caused by certain applications putting
* regioned (e.g. myaccount.eu-central-1)
* privatelink (e.g. myorg-myaccount.privatelink)

values in the `Account` field of the configuration, which could result in unsuccessful keypair authentication, as this value was relayed into the JWT unchanged and led to an invalid JWT

The change has been also manually tested by actually generating JWT and actually logging into live Snowflake account, whose names conform the pattern seen in the newly added part in auth_test.go

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -480,7 +480,7 @@ func prepareJWTToken(config *Config) (string, error) {
 	}
 	hash := sha256.Sum256(pubBytes)
 
-	accountName := strings.ToUpper(config.Account)
+	accountName := getAccountForJWT(config.Account)
 	userName := strings.ToUpper(config.User)
 
 	issueAtTime := time.Now().UTC()
@@ -571,4 +571,12 @@ func fillCachedIDToken(sc *snowflakeConn) {
 
 func fillCachedMfaToken(sc *snowflakeConn) {
 	getCredential(sc, mfaToken)
+}
+
+func getAccountForJWT(accountName string) string {
+	posDot := strings.Index(accountName, ".")
+	if posDot > 0 {
+		return strings.ToUpper(accountName[:posDot])
+	}
+	return strings.ToUpper(accountName)
 }
--- a/auth.go
+++ b/auth.go
@@ -480,7 +480,7 @@ func prepareJWTToken(config *Config) (string, error) {
 	}
 	hash := sha256.Sum256(pubBytes)
 
-	accountName := getAccountForJWT(config.Account)
+	accountName := extractAccountName(config.Account)
 	userName := strings.ToUpper(config.User)
 
 	issueAtTime := time.Now().UTC()
@@ -571,12 +571,4 @@ func fillCachedIDToken(sc *snowflakeConn) {
 
 func fillCachedMfaToken(sc *snowflakeConn) {
 	getCredential(sc, mfaToken)
-}
-
-func getAccountForJWT(accountName string) string {
-	posDot := strings.Index(accountName, ".")
-	if posDot > 0 {
-		return strings.ToUpper(accountName[:posDot])
-	}
-	return strings.ToUpper(accountName)
 }
--- a/auth_test.go
+++ b/auth_test.go
@@ -927,7 +927,7 @@ func TestOktaRetryWithNewToken(t *testing.T) {
 }
 
 func TestGetAccountForJWT(t *testing.T) {
-	accounts := map[string]string{
+	testcases := map[string]string{
 		"myaccount":                          "MYACCOUNT",
 		"myaccount.eu-central-1":             "MYACCOUNT",
 		"myaccount.eu-central-1.privatelink": "MYACCOUNT",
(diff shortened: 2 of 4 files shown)


auth.go
```diff
@@ -572,3 +572,11 @@ func fillCachedIDToken(sc *snowflakeConn) {
 func fillCachedMfaToken(sc *snowflakeConn) {
 	getCredential(sc, mfaToken)
 }
+
+func getAccountForJWT(accountName string) string {
```
sfc-gh-pfus, line 576: This function looks more general, maybe we can:
1. rename it to `getAccountNameOnly` or something like this (this name is bad, so if you have better idea - feel free!)
2. move it to dsn.go
3. I'm considering having it as a member function of `Config`, but probably this is not for the best...
WDYT?

auth_test.go
```diff
@@ -925,3 +925,21 @@ func TestOktaRetryWithNewToken(t *testing.T) {
...
+		"myorg-myaccount":                    "MYORG-MYACCOUNT",
+		"myorg-myaccount.privatelink":        "MYORG-MYACCOUNT",
+		"myorg-my-account":                   "MYORG-MY-ACCOUNT",
+		"myorg-my-account.privatelink":       "MYORG-MY-ACCOUNT",
+		"myorg-my_account":                   "MYORG-MY_ACCOUNT",
+		"myorg-my_account.privatelink":       "MYORG-MY_ACCOUNT",
+	}
+
+	for account, expected := range accounts {
+		assertEqualF(t, getAccountForJWT(account), expected)
```
sfc-gh-pfus, line 943: Can we wrap it in `t.Run` with `account` name?

auth_test.go
```diff
@@ -925,3 +925,21 @@ func TestOktaRetryWithNewToken(t *testing.T) {
 	assertEqualF(t, authResponse.MfaToken, expectedMfaToken)
 	assertEqualF(t, authResponse.SessionInfo.DatabaseName, expectedDatabaseName)
 }
+
+func TestGetAccountForJWT(t *testing.T) {
+	accounts := map[string]string{
```
sfc-gh-pfus, line 930: `testcases` is a name we usually use in this driver.

auth_test.go
```diff
@@ -925,3 +925,21 @@ func TestOktaRetryWithNewToken(t *testing.T) {
...
+		"myaccount":                          "MYACCOUNT",
+		"myaccount.eu-central-1":             "MYACCOUNT",
+		"myaccount.eu-central-1.privatelink": "MYACCOUNT",
+		"myorg-myaccount":                    "MYORG-MYACCOUNT",
+		"myorg-myaccount.privatelink":        "MYORG-MYACCOUNT",
+		"myorg-my-account":                   "MYORG-MY-ACCOUNT",
+		"myorg-my-account.privatelink":       "MYORG-MY-ACCOUNT",
+		"myorg-my_account":                   "MYORG-MY_ACCOUNT",
+		"myorg-my_account.privatelink":       "MYORG-MY_ACCOUNT",
```
sfc-gh-pfus, line 939: Is empty account possible? I believe it isn't?

sfc-gh-dszmolka: i also think it cannot be empty, fillMissingConfigParameters seems to take care

auth_test.go
sfc-gh-dszmolka: renamed

auth_test.go
sfc-gh-dszmolka: good idea, thank you ! changed

auth_test.go
```diff
@@ -925,3 +925,27 @@ func TestOktaRetryWithNewToken(t *testing.T) {
...
+	}
+
+	for account, expected := range testcases {
+		t.Run(fmt.Sprintf("Validating account part for identifier %v", account),
```
sfc-gh-pfus, line 943: I think that just using the account variable is good enough. We know the test name (it is printed above the subtest name), but if we introduce more letters it is less likely it will fit into IDE window presenting names :)

sfc-gh-dszmolka: changed accordingly

auth_test.go
```diff
@@ -925,3 +925,27 @@ func TestOktaRetryWithNewToken(t *testing.T) {
...
+		"myorg-myaccount":                    "MYORG-MYACCOUNT",
+		"myorg-myaccount.privatelink":        "MYORG-MYACCOUNT",
+		"myorg-my-account":                   "MYORG-MY-ACCOUNT",
+		"myorg-my-account.privatelink":       "MYORG-MY-ACCOUNT",
+		"myorg-my_account":                   "MYORG-MY_ACCOUNT",
+		"myorg-my_account.privatelink":       "MYORG-MY_ACCOUNT",
+	}
+
+	for account, expected := range testcases {
+		t.Run(fmt.Sprintf("%v", account),
```
sfc-gh-pfus, line 943: So we don't need `sprintf` at all`, just `t.Run(account, func...`

sfc-gh-dszmolka: oh, okay. changed.

auth.go
sfc-gh-dheyman-1: I'd have to agree with @sfc-gh-pfus - signature suggests that the it's logic is strongly coupled with JWT concept, while it's not. My suggestion is `extractAccountName(rawAccount string)`.
Still, mad respect for @sfc-gh-dszmolka for standing up for himself!

sfc-gh-dszmolka: the goal is really to strongly couple with JWT generation, the fix is targeting this area as the issue stems from wrong JWT

renamed and moved the function per the suggestions.

sfc-gh-pfus (approved):

sfc-gh-dheyman-1 (approved):

PR #1153, opened 2024-06-07
SNOW-1163210 Add tests for Max Lob Size

sfc-gh-dheyman-1:
### Description

SNOW-1163210 Add tests for Max Lob Size

### Checklist

### DIFF
--- a/bindings_test.go
+++ b/bindings_test.go
@@ -37,18 +37,25 @@ const (
 	insertSQLBulkArrayDateTimeTimestamp      = "insert into test_bulk_array_DateTimeTimestamp values(?, ?, ?, ?, ?)"
 	selectAllSQLBulkArrayDateTimeTimestamp   = "select * from test_bulk_array_DateTimeTimestamp ORDER BY 1"
 
-	enableFeatureMaxLOBSize = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY='ENABLED'"
-	unsetFeatureMaxLOBSize  = "ALTER SESSION UNSET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY"
+	enableFeatureMaxLOBSize      = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY='ENABLED'"
+	unsetFeatureMaxLOBSize       = "ALTER SESSION UNSET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY"
+	enableLargeVarcharAndBinary  = "ALTER SESSION SET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=TRUE"
+	disableLargeVarcharAndBinary = "ALTER SESSION SET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=FALSE"
+	unsetLargeVarcharAndBinary   = "ALTER SESSION UNSET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT"
 
-	// For max LOB size tests
-	// maxLOBSize = 128 * 1024 * 1024 // new max LOB size
-	maxLOBSize = 16 * 1024 * 1024 // current max LOB size
-	largeSize  = maxLOBSize / 2
-	mediumSize = largeSize / 2
+	maxVarcharAndBinarySizeParam = "varchar_and_binary_max_size_in_result"
+
 	originSize = 16 * 1024 * 1024
 	smallSize  = 16
 	// range to use for generating random numbers
 	lobRandomRange = 100000
+)
+
+var (
+	// maxLOBSize = 128 * 1024 * 1024 // new max LOB size
+	maxLOBSize = 16 * 1024 * 1024 // current max LOB size
+	largeSize  = maxLOBSize / 2
+	mediumSize = largeSize / 2
 )
 
 func TestBindingFloat64(t *testing.T) {
@@ -220,9 +227,9 @@ func TestBindingTimePtrInStruct(t *testing.T) {
 			id      *int
 			timeVal *time.Time
 		}
-		var expectedID int = 1
-		var expectedTime time.Time = time.Now()
-		var testStruct timePtrStruct = timePtrStruct{id: &expectedID, timeVal: &expectedTime}
+		var expectedID = 1
+		var expectedTime = time.Now()
+		var testStruct = timePtrStruct{id: &expectedID, timeVal: &expectedTime}
 		dbt.mustExec("CREATE OR REPLACE TABLE timeStructTest (id int, tz timestamp_tz)")
 
 		runInsertQuery := false
@@ -267,9 +274,9 @@ func TestBindingTimeInStruct(t *testing.T) {
 			id      int
 			timeVal time.Time
 		}
-		var expectedID int = 1
-		var expectedTime time.Time = time.Now()
-		var testStruct timeStruct = timeStruct{id: expectedID, timeVal: expectedTime}
+		var expectedID = 1
+		var expectedTime = time.Now()
+		var testStruct = timeStruct{id: expectedID, timeVal: expectedTime}
 		dbt.mustExec("CREATE OR REPLACE TABLE timeStructTest (id int, tz timestamp_tz)")
 
 		runInsertQuery := false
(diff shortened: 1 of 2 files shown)


bindings_test.go
```diff
@@ -37,20 +37,27 @@ const (
 	insertSQLBulkArrayDateTimeTimestamp      = "insert into test_bulk_array_DateTimeTimestamp values(?, ?, ?, ?, ?)"
 	selectAllSQLBulkArrayDateTimeTimestamp   = "select * from test_bulk_array_DateTimeTimestamp ORDER BY 1"
 
-	enableFeatureMaxLOBSize = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY='ENABLED'"
-	unsetFeatureMaxLOBSize  = "ALTER SESSION UNSET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY"
+	enableFeatureMaxLOBSize      = "ALTER SESSION SET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY='ENABLED'"
+	unsetFeatureMaxLOBSize       = "ALTER SESSION UNSET FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY"
+	enableLargeVarcharAndBinary  = "ALTER SESSION SET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=TRUE"
+	disableLargeVarcharAndBinary = "ALTER SESSION SET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT=FALSE"
+	unsetLargeVarcharAndBinary   = "ALTER SESSION UNSET ENABLE_LARGE_VARCHAR_AND_BINARY_IN_RESULT"
```
sfc-gh-pfus, line 44: What is the difference between this one and the line above?

bindings_test.go
```diff
@@ -220,9 +227,9 @@ func TestBindingTimePtrInStruct(t *testing.T) {
 			id      *int
 			timeVal *time.Time
 		}
-		var expectedID int = 1
-		var expectedTime time.Time = time.Now()
-		var testStruct timePtrStruct = timePtrStruct{id: &expectedID, timeVal: &expectedTime}
+		var expectedID = 1
```
sfc-gh-pfus, line 230: Maybe just `expected := 1`?

bindings_test.go
```diff
@@ -1151,6 +1167,20 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
 	})
 }
 
+func TestMaxLobSizeSwitch(t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.exec(disableLargeVarcharAndBinary)
```
sfc-gh-pfus, line 1172: What if we run in an env:
1. where this parameter is disabled by default?
2. where this parameter is not available?

bindings_test.go
```diff
@@ -1151,6 +1167,20 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
...
+		rows, err := dbt.query("select randstr(20000000, random())")
+		defer rows.Close()
+		assertStringContainsF(t, err.Error(), "Actual length 20000000 exceeds supported length")
+
+		dbt.exec(enableLargeVarcharAndBinary)
+		rows, err = dbt.query("select randstr(20000000, random())")
+		assertNilF(t, err)
+		dbt.exec(unsetLargeVarcharAndBinary)
```
sfc-gh-pfus, line 1180: Shouldn't be `defer`red?

driver_test.go
```diff
@@ -167,6 +167,16 @@ type DBTest struct {
 	conn *sql.Conn
 }
 
+func (dbt *DBTest) connParams() map[string]*string {
+	var params map[string]*string
+	dbt.conn.Raw(func(driverConn any) error {
```
sfc-gh-pfus, line 172: `error` is returned here. Maybe we should catch it and run `assertNilF(dbt.t, err)`?

bindings_test.go
```diff
@@ -1120,18 +1127,27 @@ func TestLOBRetrievalWithJSON(t *testing.T) {
...
-	// the LOB sizes to be tested
-	testSizes := [5]int{smallSize, originSize, mediumSize, largeSize, maxLOBSize}
-	var res string
-
 	runDBTest(t, func(dbt *DBTest) {
+		parameters := dbt.connParams()
+		varcharBinaryMaxSizeRaw := parameters[maxVarcharAndBinarySizeParam]
+		if varcharBinaryMaxSizeRaw != nil && *varcharBinaryMaxSizeRaw != "" {
+			varcharBinaryMaxSize, err := strconv.Atoi(*varcharBinaryMaxSizeRaw)
+			if err == nil {
```
sfc-gh-pfus, line 1135: Why do we need this ifs in which case it can happen? Some specific env?

bindings_test.go
```diff
@@ -1151,6 +1167,20 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
 	})
 }
 
+func TestMaxLobSizeSwitch(t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.exec(disableLargeVarcharAndBinary)
+		rows, err := dbt.query("select randstr(20000000, random())")
+		defer rows.Close()
```
sfc-gh-pfus, line 1174: We shouldn't close if we expect error.

bindings_test.go
sfc-gh-dheyman-1: Then we can use dbt.mustExec and it will throw error when this occurs :) Changed this.

bindings_test.go
sfc-gh-dheyman-1: what do you mean?

sfc-gh-pfus: Let's suppose you set this parameter above. Then, we reuse this session (because we have a pool). So if the test fail for instance a line above, this parameter will never be unset. Maybe we can add unsetting it as `defer` function at the beginning, wdyt?

bindings_test.go
sfc-gh-dheyman-1: For now we have no guarantee that this parameter will be returned from server via /login response, since it depends on FeatureWithClientEnabled matrix.

bindings_test.go
sfc-gh-dheyman-1: changed.

driver_test.go
sfc-gh-dheyman-1: nice catch, added.

bindings_test.go
sfc-gh-pfus: Ok, but the value is either not returned (two lines above) or it is returned and has proper format, right? It should never be returned but in a form in which you can't convert it to int, right?

bindings_test.go
sfc-gh-dheyman-1: I guess, unset changes this parameter to default, right?

bindings_test.go
sfc-gh-dheyman-1: sure, refactored.

bindings_test.go
sfc-gh-dheyman-1: Makes sense, added defer for unsetting the param.

bindings_test.go
```diff
@@ -1151,6 +1173,21 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
 	})
 }
 
+func TestMaxLobSizeSwitch(t *testing.T) {
+	skipMaxLobSizeTestOnGithubActions(t)
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.mustExec(disableLargeVarcharAndBinary)
+		defer dbt.mustExec(unsetLargeVarcharAndBinary)
+		if _, err := dbt.query("select randstr(20000000, random())"); err != nil {
+			assertStringContainsF(t, err.Error(), "Actual length 20000000 exceeds supported length")
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: Actually not a bad idea. I Pushed this change.

bindings_test.go
```diff
@@ -1120,18 +1133,27 @@ func TestLOBRetrievalWithJSON(t *testing.T) {
...
-	testSizes := [5]int{smallSize, originSize, mediumSize, largeSize, maxLOBSize}
-	var res string
-
 	runDBTest(t, func(dbt *DBTest) {
-		dbt.exec(enableFeatureMaxLOBSize)
+		parameters := dbt.connParams()
+		varcharBinaryMaxSizeRaw := parameters[maxVarcharAndBinarySizeParam]
+		if varcharBinaryMaxSizeRaw != nil && *varcharBinaryMaxSizeRaw != "" {
+			varcharBinaryMaxSize, err := strconv.ParseFloat(*varcharBinaryMaxSizeRaw, 64)
+			if err != nil {
```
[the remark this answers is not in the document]
sfc-gh-dheyman-1: Removed.

bindings_test.go
sfc-gh-dheyman-1: This is already changed.

PR #1156, opened 2024-06-12
Always retain `fixedType`, add unit test which fails if not

madisonchamberlain:
### Description

Fallback to retaining `fixedType` chunk if its not one of the types in the if/if else statements. Add a test which ensures that this works. 

### Checklist


chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+func TestRetainChunkWOHighPrecision(t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		query := "select 0"
+
+		ctx := WithArrowBatches(context.Background())
+
+		var rows driver.Rows
+		var err error
+
+		// must use conn.Raw so we can get back driver rows (an interface)
```
sfc-gh-pfus, line 649: In this repo we know this, we don't need to put this comment :)

chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+		// which can be cast to snowflakeRows which exposes GetArrowBatch
+		err = dbt.conn.Raw(func(x interface{}) error {
+			queryer, implementsQueryContext := x.(driver.QueryerContext)
+			assertTrueF(t, implementsQueryContext, "snowflake connection driver does not implement queryerContext")
```
sfc-gh-pfus, line 653: Same here, this is fine to do casting without this second parameter. Let it burn otherwise!

chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+
+			rows, err = queryer.QueryContext(WithArrowBatches(ctx), query, nil)
```
sfc-gh-pfus, line 655: This context is already arrow batches aware, right?

chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+			return err
+		})
+
+		assertNilF(t, err, "error running select query")
+
+		sfRows, isSfRows := rows.(SnowflakeRows)
```
sfc-gh-pfus, line 661: Same here, assume casting succeeds.

chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+
+		arrowBatches, err := sfRows.GetArrowBatches()
+		assertNilF(t, err, "error getting arrow batches")
+		assertEqualF(t, len(arrowBatches), 1, "should have one batch")
+
+		records, err := arrowBatches[0].Fetch()
+		assertNilF(t, err, "error getting batch")
+		assertNotNilF(t, records, "records should not be nil")
+
+		recs := *records
```
sfc-gh-pfus, line 672: Do we need this additional variable?

chunk_test.go
```diff
@@ -636,3 +636,51 @@ func TestQueryArrowStreamDescribeOnly(t *testing.T) {
...
+		numRecords := len(recs)
+		assertEqualF(t, numRecords, 1, "should have exactly one record")
+
+		record := recs[0]
+		assertEqualF(t, len(record.Columns()), 1, "should have exactly one column")
+
+		column := record.Column(0)
```
sfc-gh-pfus, line 679: Maybe we can have `record.Column(0).(*array.Int8)`?

chunk_test.go
madisonchamberlain: true, deleted!

chunk_test.go
madisonchamberlain: Yes, updated

chunk_test.go
madisonchamberlain: done!

chunk_test.go
madisonchamberlain: Nope, deleted

chunk_test.go
madisonchamberlain: good idea, updated

sfc-gh-pfus (approved):

PR #1157, opened 2024-06-13
Proposed goroutine wrapper solution

madisonchamberlain:
### Description

As discussed in slack, I would like to have the ability to wrap a panic in a go routine, rather than letting it crash our service. We want to simply use the recover function to grab the panic, and then send that panic to sentry (the service we use to track errors). For us this would make debugging easier and more productive. 

I have added a variable `GoroutineWrapper` which is a function that takes in a function. I call this variable around all of the goroutines that are spawned throughout the codebase. On init, a caller can provide a function which does what it wants to wrap goroutines. If nothing is specified, we fallback on `defaultDoesNothing` which just calls the goroutine with no wrapping/handling as is happening now. Please let me know your thoughts, I am happy to change anything! 

### Checklist
- [ ] Extended the README / documentation, if necessary - I will do this once we are in a state of knowing for sure how we want this to look/if this is something we want to do

### DIFF
--- a/authexternalbrowser.go
+++ b/authexternalbrowser.go
@@ -280,49 +280,44 @@ func doAuthenticateByExternalBrowser(
 		logger.WithContext(ctx).Errorf("unable to accept connection. err: %v", err)
 		log.Fatal(err)
 	}
-	go GoroutineWrapper(
-		ctx,
-		func() {
-			go func(c net.Conn) {
-				var buf bytes.Buffer
-				total := 0
-				encodedSamlResponse := ""
-				var errAccept error
-				for {
-					b := make([]byte, bufSize)
-					n, err := c.Read(b)
-					if err != nil {
-						if err != io.EOF {
-							logger.WithContext(ctx).Infof("error reading from socket. err: %v", err)
-							errAccept = &SnowflakeError{
-								Number:      ErrFailedToGetExternalBrowserResponse,
-								SQLState:    SQLStateConnectionRejected,
-								Message:     errMsgFailedToGetExternalBrowserResponse,
-								MessageArgs: []interface{}{err},
-							}
-						}
-						break
+	go func(c net.Conn) {
+		var buf bytes.Buffer
+		total := 0
+		encodedSamlResponse := ""
+		var errAccept error
+		for {
+			b := make([]byte, bufSize)
+			n, err := c.Read(b)
+			if err != nil {
+				if err != io.EOF {
+					logger.WithContext(ctx).Infof("error reading from socket. err: %v", err)
+					errAccept = &SnowflakeError{
+						Number:      ErrFailedToGetExternalBrowserResponse,
+						SQLState:    SQLStateConnectionRejected,
+						Message:     errMsgFailedToGetExternalBrowserResponse,
+						MessageArgs: []interface{}{err},
 					}
-					total += n
-					buf.Write(b)
-					if n < bufSize {
-						// We successfully read all data
-						s := string(buf.Bytes()[:total])
-						encodedSamlResponse, errAccept = getTokenFromResponse(s)
-						break
-					}
-					buf.Grow(bufSize)
 				}
-				if encodedSamlResponse != "" {
-					httpResponse := buildResponse(application)
-					c.Write(httpResponse.Bytes())
-				}
-				c.Close()
-				encodedSamlResponseChan <- encodedSamlResponse
-				errChan <- errAccept
-			}(conn)
-		},
-	)
+				break
+			}
+			total += n
+			buf.Write(b)
+			if n < bufSize {
+				// We successfully read all data
+				s := string(buf.Bytes()[:total])
+				encodedSamlResponse, errAccept = getTokenFromResponse(s)
+				break
+			}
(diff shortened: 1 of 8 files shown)


function_wrapper_test.go
```diff
@@ -0,0 +1,56 @@
...
+		close(goWrapperCalled)
+	}
+	
+	f()
+}
+
+func TestGoWrapper(t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		initGoWrapperCalledChannel()
+		GoroutineWrapper = closeGoWrapperCalledChannel
```
sfc-gh-pfus, line 45: We modify a global variable here - maybe it's good to remember old value and revert it in `t.Cleanup`?

function_wrapper_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+		numrows := 100000
+		withCancelCtx, cancel := context.WithCancel(context.Background())
+		// using async mode because I know that will trigger a goroutine to be fired off
```
sfc-gh-pfus, line 49: I'm curious, what happens here? :)

function_wrapper_test.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+var (
+	goWrapperCalled chan struct{}
+	once       sync.Once
+)
+
+// Initialize the goWrapperCalled channel which we will use for testing the goroutine wrap function
+func initGoWrapperCalledChannel() {
+	once.Do(func() {
```
sfc-gh-pfus, line 17: All this is a bit complex - maybe we can just remember some global variable when test wrapper is invoked instead of using and closing channels?

function_wrappers.go
```diff
@@ -0,0 +1,10 @@
+package gosnowflake
+
+// Define a type for the function that wraps goroutines
+type GoroutineWrapperFunc func(f func())
+
+// Global variable to hold the function pointer
+var defaultDoesNothing = func(f func()){
```
sfc-gh-pfus, line 7: Maybe we can call it `NoopGoroutineWrapper`?

function_wrapper_test.go
madisonchamberlain: Ok, will do

function_wrapper_test.go
madisonchamberlain: good idea

authexternalbrowser.go
```diff
@@ -277,44 +279,48 @@ func doAuthenticateByExternalBrowser(
...
-				if err != io.EOF {
-					logger.WithContext(ctx).Infof("error reading from socket. err: %v", err)
-					errAccept = &SnowflakeError{
-						Number:      ErrFailedToGetExternalBrowserResponse,
-						SQLState:    SQLStateConnectionRejected,
-						Message:     errMsgFailedToGetExternalBrowserResponse,
-						MessageArgs: []interface{}{err},
+	go GoroutineWrapper(
+		func() {
+			go func(c net.Conn) {
```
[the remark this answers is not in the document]
madisonchamberlain: Oh yes, updated

function_wrapper_test.go
madisonchamberlain: Like why am I using `WithAsyncMode` to kick off a goroutine rather than just doing it in this function? I wanted to demonstrate that the changes I made in the code would really wrap the goroutine like I intended, and I knew that using this flag would trigger the goroutine on line ~40 of async.go. Alternatively though, if we just want to test that the wrapper func works like expected when we use it to wrap a go routine we can just do something like this. Whatever you think is preferred 

```
func TestGoWrapper(t *testing.T) {
	t.Cleanup(
		func() {
			// set the global wrapper called variable back to false
			goWrapperCalled = false
		},
	)

	// set the global variable which holds our desired wrapper to the one which switches 
	// goWrapperCalled to true
	GoroutineWrapper = closeGoWrapperCalledChannel

	// define a channel so we will know when our goroutine is done 
	ch := make(chan struct{})

	// Start a goroutine that closes the channel after a delay, wrapped by our wrapper
	go GoroutineWrapper(
		context.Background(),
		func() {
			time.Sleep(100 * time.Millisecond)
			close(ch)
		},
	)

	// wait for our pretend work to be done
	<-ch

	// check that our wrapper was called 
	assertTrueF(t, goWrapperCalled, "channel should be closed, indicating our wrapper worked")
}
```

sfc-gh-pfus (approved):

function_wrapper_test.go
```diff
@@ -0,0 +1,39 @@
...
+	goWrapperCalled = true
+
+	f()
+}
+
+func TestGoWrapper(t *testing.T) {
+	runDBTest(t, func(dbt *DBTest) {
+		t.Cleanup(
+			func() {
+				goWrapperCalled = false
```
sfc-gh-pfus, line 25: We should also revert to `noopGroutineWrapper`, right? Or even better:

```
oldGoroutineWrapper := GoroutineWrapper
t.Cleanup(func() {
  GoroutineWrapper = oldGoroutineWrapper
})

sfc-gh-pfus (approved):

PR #1160, opened 2024-06-18
SNOW-1346547: add check for session expiry when getting query result

sfc-gh-ext-simba-jl:
### Description

drivers issue 957: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/957

The root cause is that the heartbeat process has renewed the session token [here](https://github.com/snowflakedb/gosnowflake/blob/v1.10.1/heartbeat.go#L79) but the new token in the get query result request's header was not updated in [getQueryResultWithRetriesForAsyncMode()](https://github.com/snowflakedb/gosnowflake/blob/822-support-put-cancel/async.go#L136).

This PR adds a check for session expiry in `getQueryResultWithRetriesForAsyncMode()` and update the token in the header if needed. Similar to ODBC's implementation [here](https://github.com/snowflakedb/snowflake-odbc/blob/master/Source/Platform/Statement.cpp#L385), where there should be a check for the session expired code 390112 during the getting query result request and allows a retry after updating the session token.

### Checklist


async.go
```diff
@@ -160,7 +161,34 @@ func getQueryResultWithRetriesForAsyncMode(
 			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
 			return respd, err
 		}
-		if respd.Code != queryInProgressAsyncCode {
+		if respd.Code == sessionExpiredCode {
+			// Update the session token in the header and retry
+			token, _, _ := sr.TokenAccessor.GetTokens()
+			if headers[headerAuthorizationKey] != fmt.Sprintf(headerSnowflakeToken, token) {
```
sfc-gh-pfus, line 167: So this `if` handles the case when we use external token accessor? Maybe we should ensure, that token that we set is not empty, when custom token accessor is not set?

async.go
```diff
@@ -160,7 +161,34 @@ func getQueryResultWithRetriesForAsyncMode(
...
+				headers[headerAuthorizationKey] = fmt.Sprintf(headerSnowflakeToken, token)
+				logger.WithContext(ctx).Info("Session token has been updated.")
+				retry++
```
sfc-gh-pfus, line 170: Why do we bump this retry instead of `retryCountForSessionRenewal`?

async.go
```diff
@@ -160,7 +161,34 @@ func getQueryResultWithRetriesForAsyncMode(
...
+			if err = sr.renewExpiredSessionToken(ctx, timeout, token); err != nil {
+				logger.WithContext(ctx).Errorf("failed to renew session token. err: %v", err)
+				return respd, err
+			}
+			retryCountForSessionRenewal++
+
+			// If this is the first response, go back to retry the query
+			// since it failed due to session expiration
+			logger.WithContext(ctx).Infof("retry count for session renewal: %v", retryCountForSessionRenewal)
+			if retryCountForSessionRenewal < 2 {
```
sfc-gh-pfus, line 184: Do we need this bumps at all? Maybe it's enough to assume why attempt of retry, so everything below line 178 can be removed? Especially that it is already retried on HTTP retry level. Why do we need this retries in your opinion?

connection_test.go
```diff
@@ -139,6 +139,53 @@ func TestGetQueryResultUsesTokenFromTokenAccessor(t *testing.T) {
 	}
 }
 
+func TestGetQueryResultTokenExpiry(t *testing.T) {
```
sfc-gh-pfus, line 142: It's also worth having another test for custom token accessor?

connection_test.go
```diff
@@ -139,6 +139,53 @@ func TestGetQueryResultUsesTokenFromTokenAccessor(t *testing.T) {
...
+	ta := getSimpleTokenAccessor()
+	token := "snowflake-test-token"
+	ta.SetTokens(token, "", 1)
+	funcGetMock := func(_ context.Context, _ *snowflakeRestful, _ *url.URL,
+		headers map[string]string, _ time.Duration) (*http.Response, error) {
+		dd := &execResponseData{}
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Fixed

connection_test.go
```diff
@@ -139,6 +139,53 @@ func TestGetQueryResultUsesTokenFromTokenAccessor(t *testing.T) {
...
+		er := &execResponse{
+			Data:    *dd,
+			Message: "",
+			Code:    sessionExpiredCode,
+			Success: true,
+		}
+		ba, err := json.Marshal(er)
+		if err != nil {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jl: Fixed

async.go
sfc-gh-ext-simba-jl: Added a check for non-empty token and the related test `TestGetQueryResultTokenNotSet`.

connection_test.go
sfc-gh-ext-simba-jl: Tests for simpleTokenAccessor with and without token set should be enough.

async.go
sfc-gh-ext-simba-jl: This is just to update the token that has been renewed by heartbeat already. `retryCountForSessionRenewal` is to check actual token renewal done in this function.

async.go
sfc-gh-ext-simba-jl: Everything below line 178 is used for checking token expiry in normal case where the token expires during `getQueryResultWithRetriesForAsyncMode()`

sfc-gh-pfus (approved):

sfc-gh-ext-simba-jl:
Added a check for session expiry in getQueryResultWithRetriesForAsyncMode() and update the token in the header if needed.

PR #1171, opened 2024-07-02
SNOW-1501890 Implement structured maps

sfc-gh-pfus:
### Description

SNOW-1501890 Implemented structured maps

### Checklist

### DIFF
--- a/converter.go
+++ b/converter.go
@@ -27,6 +27,7 @@ import (
 )
 
 const format = "2006-01-02 15:04:05.999999999"
+const numberDefaultPrecision = 38
 
 type timezoneType int
 
@@ -621,7 +622,7 @@ func goTypeToFieldMetadata(typ reflect.Type, tsmode snowflakeType, params map[st
 	case reflect.Int, reflect.Int8, reflect.Uint8, reflect.Int16, reflect.Int32, reflect.Int64:
 		return fieldMetadata{
 			Type:      "FIXED",
-			Precision: 38,
+			Precision: numberDefaultPrecision,
 			Nullable:  true,
 		}, nil
 	case reflect.Float32, reflect.Float64:
@@ -643,7 +644,7 @@ func goTypeToFieldMetadata(typ reflect.Type, tsmode snowflakeType, params map[st
 		} else if typ.AssignableTo(reflect.TypeOf(sql.NullByte{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt16{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt32{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt64{})) {
 			return fieldMetadata{
 				Type:      "FIXED",
-				Precision: 38,
+				Precision: numberDefaultPrecision,
 				Nullable:  true,
 			}, nil
 		} else if typ.AssignableTo(reflect.TypeOf(sql.NullFloat64{})) {
@@ -669,7 +670,6 @@ func goTypeToFieldMetadata(typ reflect.Type, tsmode snowflakeType, params map[st
 		} else if tsmode == timestampNtzType {
 			return fieldMetadata{
 				Type:     "TIMESTAMP_NTZ",
-				Scale:    9,
 				Nullable: true,
 			}, nil
 		} else if tsmode == timestampLtzType {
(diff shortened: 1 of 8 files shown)


converter.go
```diff
@@ -388,6 +387,345 @@
...
+			}, nil
+		} else if tsmode == timestampTzType {
+			return fieldMetadata{
+				Type:     "TIMESTAMP_TZ",
+				Nullable: true,
+			}, nil
+		} else if tsmode == timestampNtzType {
+			return fieldMetadata{
+				Type:     "TIMESTAMP_NTZ",
+				Scale:    9,
```
sfc-gh-dheyman-1, line 672: Is this a default? Why 9?

converter.go
```diff
@@ -388,6 +387,345 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
+			}, nil
+		} else if typ.AssignableTo(reflect.TypeOf(sql.NullBool{})) {
+			return fieldMetadata{
+				Type:     "BOOLEAN",
+				Nullable: true,
+			}, nil
+		} else if typ.AssignableTo(reflect.TypeOf(sql.NullByte{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt16{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt32{})) || typ.AssignableTo(reflect.TypeOf(sql.NullInt64{})) {
+			return fieldMetadata{
+				Type:      "FIXED",
+				Precision: 38,
```
sfc-gh-dheyman-1, line 646: Can we extract this 38 to a const? It's used repetitively

converter.go
```diff
@@ -388,6 +387,345 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
+		jsonBytes, err = json.Marshal(m)
+		if err != nil {
+			return bindingValue{}, err
+		}
+	} else if typOf.Elem().AssignableTo(reflect.TypeOf(sql.NullBool{})) {
+		m := make(map[string]*bool, valOf.Len())
+		iter := valOf.MapRange()
+		for iter.Next() {
+			val := iter.Value().Interface().(sql.NullBool)
+			if val.Valid {
```
sfc-gh-dheyman-1, line 487: lots of duplicated code, sadly not much can be done here, nor there is much sense to do it :/

converter.go
```diff
@@ -388,6 +387,345 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
 	return bindingValue{&resString, "json", nil}, nil
 }
 
+func mapToString(v driver.Value, tsmode snowflakeType, params map[string]*string) (bindingValue, error) {
```
sfc-gh-dheyman-1, line 390: anything we can do to make this func shorter?

datetime.go
```diff
@@ -34,6 +35,18 @@ var formatReplacements = []formatReplacement{
 	{input: "TZM", output: "00"},
 }
 
+func timeToString(t time.Time, dateTimeType string, params map[string]*string) (string, error) {
+	sfFormat, err := dateTimeInputFormatByType(dateTimeType, params)
+	if err != nil {
+		return "", err
```
sfc-gh-dheyman-1, line 41: shouldn't we log these error?

structured_type.go
```diff
@@ -75,9 +75,19 @@ type StructuredObjectWriterContext interface {
 	WriteNullTime(fieldName string, value sql.NullTime, tsmode []byte) error
 	WriteStruct(fieldName string, value StructuredObjectWriter) error
 	WriteNullableStruct(fieldName string, value StructuredObjectWriter, typ reflect.Type) error
+	// WriteRaw is used for inserting slices and maps only.
+	WriteRaw(fieldName string, value any, tsmode ...[]byte) error
```
sfc-gh-dheyman-1, line 79: if only slices and maps, can we type this properly and not use "any"?

converter.go
sfc-gh-pfus: It seems that this is not required at all.

converter.go
sfc-gh-pfus: Done.

converter.go
sfc-gh-pfus: Yeah, there is no common interface :(

datetime.go
sfc-gh-pfus: Why? We don't log everywhere. Root caller should log error propagated to the top. Otherwise we would have multiple logs with the same text.

structured_type.go
sfc-gh-pfus: No, we can't. Generics are not supported in interfaces. Go rulez...

sfc-gh-pmotacki (approved):

sfc-gh-dheyman-1 (approved):

PR #1174, opened 2024-07-08
SNOW-116310: Adjust BINARY col size to match assertions

sfc-gh-dheyman-1:
### Description

SNOW-116310 Adjusted BINARY col size for select statement used in tests, to match assertions.

### Checklist

### DIFF
--- a/put_get_test.go
+++ b/put_get_test.go
@@ -642,7 +642,6 @@ func TestPutLargeFile(t *testing.T) {
 
 func TestPutGetMaxLOBSize(t *testing.T) {
 	// the LOB sizes to be tested
-	// set "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true", when testing for increased max LOB size
 	testCases := [5]int{smallSize, originSize, mediumSize, largeSize, maxLOBSize}
 
 	runDBTest(t, func(dbt *DBTest) {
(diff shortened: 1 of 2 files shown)


sfc-gh-astachowski (approved):

put_get_test.go
```diff
@@ -642,6 +642,7 @@ func TestPutLargeFile(t *testing.T) {
 
 func TestPutGetMaxLOBSize(t *testing.T) {
 	// the LOB sizes to be tested
+	// set "alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true", when testing for increased max LOB size
```
sfc-gh-pfus, line 645: Instead of this comment maybe you can comment `dbt.mustExec`?

sfc-gh-pfus (approved):

put_get_test.go
sfc-gh-dheyman-1: ok, changed the comment.

sfc-gh-pfus: I have one more thought - shouldn't we test with this param set to true? I believe you don't want to enable it by default, because on may envs we don't have permissions to set it? So maybe we can do something like:
```
_, err := dbt.Exec("alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true")
if err != nil {
   // log err
  // t.Skip
}
```

put_get_test.go
```diff
@@ -645,6 +645,13 @@ func TestPutGetMaxLOBSize(t *testing.T) {
 	testCases := [5]int{smallSize, originSize, mediumSize, largeSize, maxLOBSize}
 
 	runDBTest(t, func(dbt *DBTest) {
+		if maxLOBSize > originSize { // for increased max LOB size
+			_, err := dbt.exec("alter session set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN = true")
+			if err != nil {
+				dbt.Errorf("Unable to set ALLOW_LARGE_LOBS_IN_EXTERNAL_SCAN parameter for increased max LOB size")
```
sfc-gh-pfus, line 651: Why error? Should we error this one or only skip test if current env does not support setting it?

sfc-gh-dheyman-1: Yes, we should throw an error, because this test is ran by default on 16MB max lob sizes (with possibility of manual override to 128 to run it locally) If someone intentionally switches it to 128MB, then that person should see error.

sfc-gh-pfus (approved):

PR #1175, opened 2024-07-08
SNOW-1524314 Hide structured types behind context

sfc-gh-pfus:
### Description

SNOW-1524314 Hides structured types behind context. Without using this context, structured types are going to be returned as semistructured types - as strings.

### Checklist


converter.go
```diff
@@ -185,6 +186,9 @@ func snowflakeTypeToGo(ctx context.Context, dbtype snowflakeType, scale int64, f
 		}
 		return nil
 	case mapType:
+		if !structuredTypesEnabled {
```
sfc-gh-pmotacki, line 189: Is it sure thing that map won't be returned without enabling structured types? Doesn't fields need to be checked here?

converter.go
```diff
@@ -1208,6 +1224,9 @@ func arrowToValue(ctx context.Context, rowIdx int, srcColumnMeta fieldMetadata,
 			}
 		} else {
 			// structured map as native arrow
+			if !structuredTypesEnabled {
+				return nil, errors.New("structured types must be enabled to use with native arrow")
```
sfc-gh-dheyman-1, line 1228: this string can be extracted to const

sfc-gh-dheyman-1 (approved):

converter.go
sfc-gh-pfus: Ok, done.

sfc-gh-pmotacki (approved):

PR #1177, opened 2024-07-10
SNOW-1524204 handle session variables in DSN

sfc-gh-dszmolka:
Session variables (https://docs.snowflake.com/en/sql-reference/session-variables) must be specified as `$variablename=value` . 

Currently due to their urlencoding, they end up being sent as `%24variablename=value` to the server side, which in turn won't recognize it as a session variable and won't use it.

Fix attempts to introduce a urldecoding function to those custom Params which are session variables to be actually sent as valid session variable format.
Besides the unit test, fix was manually tested to confirm session variables are sent correctly, and a subsequent `SHOW VARIABLES;` query in the same session actually shows the variable.

**edit**: since then, an end-to-end test has been also added; which actually connects to an account, and verifies if the session variables are matching what is sent


dsn_test.go
```diff
@@ -1639,3 +1639,19 @@ func TestExtractAccountName(t *testing.T) {
...
+	testcases := map[string]string{
+		"query_tag":             "query_tag",
+		"%24my_custom_variable": "$my_custom_variable",
+	}
+
+	for param, expected := range testcases {
+		t.Run(param, func(t *testing.T) {
+			decodedParam := urlDecodeIfNeeded(param)
+			if decodedParam != expected {
+				t.Fatalf("urlDecodeIfNeeded returned unexpected response (%v), should be %v", decodedParam, expected)
```
sfc-gh-pfus, line 1653: Can we use `assertNilE` here?

dsn_test.go
```diff
@@ -1639,3 +1639,19 @@ func TestExtractAccountName(t *testing.T) {
 		})
 	}
 }
+
+func TestDecodeDSNParam(t *testing.T) {
```
sfc-gh-pfus, line 1643: If the purpose of this function is to unit test a single method, maybe we can name it `TestUrlDecodeIfNeeded`?

dsn_test.go
```diff
@@ -1639,3 +1639,19 @@ func TestExtractAccountName(t *testing.T) {
...
+
+	for param, expected := range testcases {
+		t.Run(param, func(t *testing.T) {
+			decodedParam := urlDecodeIfNeeded(param)
+			if decodedParam != expected {
+				t.Fatalf("urlDecodeIfNeeded returned unexpected response (%v), should be %v", decodedParam, expected)
+			}
+		})
+	}
+}
```
sfc-gh-pfus, line 1657: Can we have an e2e test as well setting some session parameter?

dsn_test.go
sfc-gh-dszmolka: that is indeed the purpose, renaming

dsn_test.go
sfc-gh-dszmolka: i'm probably missing something here :( `urlDecodeIfNeeded` doesn't return `nil` so not sure why we want to check that?

dsn_test.go
sfc-gh-dszmolka: added an end-to-end test

dsn_test.go
sfc-gh-pfus: Sorry, I meant `assertEqualE` :)

dsn_test.go
```diff
@@ -1639,3 +1640,58 @@ func TestExtractAccountName(t *testing.T) {
...
+	}
+	defer db.Close()
+	query := "SHOW /* gosnowflake TestUrlDecodeIfNeededE2E */ VARIABLES;"
+	rows, err := db.Query(query)
+	if err != nil {
+		t.Fatalf("TestUrlDecodeIfNeededE2E failed to run SHOW VARIABLES query, error: %v", err)
+	}
+	defer rows.Close()
+	var v1, v2, v3, v4, v5, v6, v7 any
+	for rows.Next() {
```
sfc-gh-pfus, line 1686: I think it's better to assert on this (`assertTrueF(t, rows.Next())`). If no rows are returned, the test will end up correctly, but skipping this section at all.

dsn_test.go
sfc-gh-dszmolka: oh, got it, replaced

dsn_test.go
```diff
@@ -1639,3 +1640,58 @@ func TestExtractAccountName(t *testing.T) {
...
+	mydsn, err := DSN(cfg)
+	if err != nil {
+		t.Fatalf("TestUrlDecodeIfNeededE2E failed to create DSN from Config: %v, err: %v", cfg, err)
+	}
+	db, err := sql.Open("snowflake", mydsn)
+	if err != nil {
+		t.Fatalf("TestUrlDecodeIfNeededE2E failed to connect. %v, err: %v", mydsn, err)
+	}
+	defer db.Close()
+	query := "SHOW /* gosnowflake TestUrlDecodeIfNeededE2E */ VARIABLES;"
```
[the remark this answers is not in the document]
sfc-gh-dszmolka: we don't actually need it, i was thinking about if there's any reason we want to identify where is coming from , using only server-side tools (like query history etc), this would allow for it

but indeed isn't necessary, so removing it

dsn_test.go
sfc-gh-dszmolka: the test expects to read the same session variable back from the rows returned, so if no rows are returned that's a bad thing and test should fail (existing assertion `assertDeepEqualE` should also cover that already)

dsn_test.go
```diff
@@ -1639,3 +1640,58 @@ func TestExtractAccountName(t *testing.T) {
...
+
+	cfg := &Config{
+		Account:  os.Getenv("SNOWFLAKE_TEST_ACCOUNT"),
+		User:     os.Getenv("SNOWFLAKE_TEST_USER"),
+		Password: os.Getenv("SNOWFLAKE_TEST_PASSWORD"),
+		Params:   map[string]*string{"$" + customVarName: &customVarValue, "query_tag": &myQueryTag},
+	}
+	mydsn, err := DSN(cfg)
+	if err != nil {
+		t.Fatalf("TestUrlDecodeIfNeededE2E failed to create DSN from Config: %v, err: %v", cfg, err)
```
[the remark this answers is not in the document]
sfc-gh-dszmolka: yes, good point, replaced all instances of `err != nil` to `assertNilE`

dsn_test.go
sfc-gh-dszmolka: (got this now what you meant, after the internal discussion , sent a commit accordingly)

sfc-gh-pfus (approved):

sfc-gh-dheyman-1 (approved):

dsn_test.go
```diff
@@ -1817,3 +1818,52 @@ func TestExtractAccountName(t *testing.T) {
...
+			assertEqualE(t, decodedParam, expected)
+		})
+	}
+}
+
+func TestUrlDecodeIfNeededE2E(t *testing.T) {
+	customVarName := "CUSTOM_VARIABLE"
+	customVarValue := "test"
+	myQueryTag := "mytag"
+	testPort, defaultPort := strconv.Atoi(os.Getenv("SNOWFLAKE_TEST_PORT"))
```
sfc-gh-pfus, line 1839: I'm not sure I get it. Why this is a default port? I believe that the second value is an error? So maybe we can call it as a regular `err` which is a standard behaviour in Go?

sfc-gh-pfus (approved):

PR #1179, opened 2024-07-12
SNOW-1416000 Structure types array null handling

sfc-gh-astachowski:
### Description

SNOW-1416000 Added new context option that makes values within arrays be parsed to sql.NullXYZ types, allowing for handling nulls in the data.

### Checklist


converter.go
```diff
@@ -1634,96 +1637,306 @@ func buildListFromNativeArrow(ctx context.Context, rowIdx int, fieldMetadata fie
 		case *array.Decimal128:
 			if higherPrecision && fieldMetadata.Scale == 0 {
 				return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) (*big.Int, error) {
-					return arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale).(*big.Int), nil
+					bigInt := arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale)
+					if bigInt == nil {
```
sfc-gh-pfus, line 1641: Why do we need this change?

converter.go
```diff
@@ -1634,96 +1637,306 @@ func buildListFromNativeArrow(ctx context.Context, rowIdx int, fieldMetadata fie
...
+						return nil, nil
+					}
+					return bigInt.(*big.Int), nil
+
 				})
 			} else if higherPrecision && fieldMetadata.Scale != 0 {
 				return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) (*big.Float, error) {
-					return arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale).(*big.Float), nil
+					bigFloat := arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale)
```
sfc-gh-pfus, line 1649: Same here.

converter.go
```diff
@@ -1231,6 +1231,9 @@ func buildMapValues[K comparable, Vnullable any, VnotNullable any](mapValuesNull
 	}
 	result := make(map[K]VnotNullable, len(m))
 	for k, v := range m {
+		if v == nil {
```
sfc-gh-pfus, line 1234: But nils are ok, as long as we talk about pointer types, right?

converter.go
```diff
@@ -1634,96 +1637,306 @@ func buildListFromNativeArrow(ctx context.Context, rowIdx int, fieldMetadata fie
...
 			} else if !higherPrecision && fieldMetadata.Scale == 0 {
+				if arrayValuesNullableEnabled(ctx) {
+					return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) (sql.NullInt64, error) {
+						v := arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale)
+						if v == nil {
+							return sql.NullInt64{Valid: false}, nil
+						}
+						val, err := strconv.ParseInt(v.(string), 10, 64)
+						if err != nil {
+							return sql.NullInt64{Valid: false}, nil
```
sfc-gh-pfus, line 1666: We should return error here, right?

converter.go
```diff
@@ -1634,96 +1637,306 @@ func buildListFromNativeArrow(ctx context.Context, rowIdx int, fieldMetadata fie
...
 			} else {
+				if arrayValuesNullableEnabled(ctx) {
+					return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) (sql.NullFloat64, error) {
+						v := arrowDecimal128ToValue(typedValues, j, higherPrecision, fieldMetadata.Scale)
+						if v == nil {
+							return sql.NullFloat64{Valid: false}, nil
+						}
+						val, err := strconv.ParseFloat(v.(string), 64)
+						if err != nil {
+							return sql.NullFloat64{Valid: false}, nil
```
sfc-gh-pfus, line 1688: Same.

converter.go
```diff
@@ -1634,96 +1637,306 @@ func buildListFromNativeArrow(ctx context.Context, rowIdx int, fieldMetadata fie
...
+			}
+			return resBool.(bool), nil
+
 		})
+
 	case binaryType:
 		return mapStructuredArrayNativeArrowRows(offsets, rowIdx, func(j int) ([]byte, error) {
-			return arrowBinaryToValue(values.(*array.Binary), j).([]byte), nil
+			res := arrowBinaryToValue(values.(*array.Binary), j)
+			if res == nil {
```
sfc-gh-pfus, line 1832: Needed? Can be inlined?

converter.go
```diff
@@ -1736,7 +1949,11 @@ func buildArrowListRecursive[T any](ctx context.Context, rowIdx int, fieldMetada
 		if err != nil {
 			return nil, err
 		}
+		if arrowList == nil {
```
sfc-gh-pfus, line 1952: Is this variable untyped? Is this condition tested in tests?

structured_type.go
```diff
@@ -1150,22 +1150,25 @@ func (st *structuredType) fieldMetadataByFieldName(fieldName string) (fieldMetad
 	return fieldMetadata{}, errors.New("no metadata for field " + fieldName)
 }
 
-func structuredTypesEnabled(ctx context.Context) bool {
-	v := ctx.Value(enableStructuredTypes)
+func optionEnabled(ctx context.Context, option contextKey) bool {
```
sfc-gh-pfus, line 1153: Good change! <3

converter.go
sfc-gh-astachowski: Thats the same issue as we've discussed with the []byte. The assertion on nil results in an unnecessary panic.

converter.go
sfc-gh-astachowski: Needed because of the panic with assertion we disciussed. Can't inline it, because it's used in the "else" statement that can't be a proper else because of linter

converter.go
sfc-gh-astachowski: There is for sure a scenario where `buildListFromNativeArrow` returns `nil, nil`, so this would panic if we didn't catch that. Lack of this condition probably means that this wasn't tested, I can try and create a case where this comes up

converter.go
sfc-gh-astachowski: Good catch

converter.go
sfc-gh-pfus: Ok, so I'm not getting it. This one won't work correctly IMHO. This function returns `snowflakeValue` which is an effective alias for `interface{}`. So here we never have `nil` - it's always "typed" nil, and this condition is never true... Try running `TestWithHigherPrecision` and stop there in debug mode :)

sfc-gh-astachowski: I've added a test  `TestArraysWithNullValuesHigherPrecision` to demonstrate. Try running it now and with the previous code

sfc-gh-pfus: Ok, I see. This smells, because `arrowDecimal128ToValue` may return string... It's ok then.

sfc-gh-astachowski: What does returning a string change in this context?

converter.go
sfc-gh-astachowski: Please have a look at how I handled this

converter.go
sfc-gh-astachowski: Upon further inspection, this comes up in an already tested case. This is similar to all of the above discussions - nil can't be asserted to be of a certain type

converter.go
sfc-gh-pfus: If we didn't return string, we could return typed `*big.Int` instead of `any` - and this if wouldn't be needed at all.

sfc-gh-dheyman-1 (approved):

sfc-gh-pfus (approved):

sfc-gh-astachowski:

PR #1191, opened 2024-08-01
SNOW-1524257 Implement GCM encryption

sfc-gh-pfus:
### Description

SNOW-1524257 Implemented GCM encryption to be ready for CSP integrations.

### Checklist

### DIFF
--- a/encrypt_util.go
+++ b/encrypt_util.go
@@ -15,19 +15,10 @@ import (
 	"strconv"
 )
 
-type snowflakeFileEncryption struct {
-	QueryStageMasterKey string `json:"queryStageMasterKey,omitempty"`
-	QueryID             string `json:"queryId,omitempty"`
-	SMKID               int64  `json:"smkId,omitempty"`
-}
-
-// PUT requests return a single encryptionMaterial object whereas GET requests
-// return a slice (array) of encryptionMaterial objects, both under the field
-// 'encryptionMaterial'
-type encryptionWrapper struct {
-	snowflakeFileEncryption
-	EncryptionMaterials []snowflakeFileEncryption
-}
+var (
+	defaultKeyAad  = make([]byte, 0)
+	defaultDataAad = make([]byte, 0)
+)
 
 // override default behavior for wrapper
 func (ew *encryptionWrapper) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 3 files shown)


encrypt_util.go
```diff
@@ -298,6 +284,149 @@ func decryptStream(
...
+	}
+	return cipher.NewGCMWithNonceSize(block, 16)
+}
+
+func encryptFileGCM(
+	sfe *snowflakeFileEncryption,
+	filename string,
+	tmpDir string) (
+	*gcmEncryptMetadata, string, error) {
+	tmpOutputFile, err := os.CreateTemp(tmpDir, baseName(filename)+"#")
```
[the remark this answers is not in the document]
sfc-gh-pfus: It is propagated from other layer, this is not encryption code responsibility.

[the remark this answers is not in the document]
sfc-gh-dheyman-1: IMHO if it has responsibility to write file using os package it could easily perform this check without breaking SRP. Just a suggestion though ;)

sfc-gh-dheyman-1 (approved):

encrypt_util.go
sfc-gh-astachowski, line 316: Why do we add "#" here?

sfc-gh-pfus: To be honest - I don't know, but we do the same in CBC encryption. The case is that such magic things might have solved some problems on some specific OSes or something in the past and I'm afraid to replace it.

sfc-gh-mkubik (approved):

encrypt_util.go
sfc-gh-mkubik: After trying to figure this out for some time I guess it's just a way to prevent collisions in filenames. The tempfile uses the same filename as the infile so I guess it's to prevent overriding the infile in case the tmpDir is empty or something

sfc-gh-pfus: Yes, I generally agree, but whether it has to be that way or not is another question. I think that in 99% we can replace it with `.bak` or anything, but I'd leave it as is.

sfc-gh-astachowski: It will most likely not cause any problems, but in case some one, for god only know what reason, has files called my_file and my_file# it might cause a collision?

sfc-gh-astachowski: It feels like we should at least check if a file of this name exists

sfc-gh-pfus: It doesn't work like that - the second argument is called `pattern`, and this is only a prefix. This function always creates a new file.

PR #1211, opened 2024-09-19
[SNOW-1669514] Honor the Valuer/Stringer methods to resolve #1209

### LINKED ISSUE #1209: SNOW-1669514: uuid.UUID Valuer Interface No Longer Honored
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1.11.1
   
Mac, Linux x64 amd and m1
   
1.22.7

8.34.0

5. What did you do?
Updated from 10.1 to 11.1.

Values updated when using an in clause containing a struct that IS a string. We also had to implement a custom serializer to convert UUIDs to strings prior to making their way to sql. It seems like String() isn't honored anymore on struct?

Binding array of structs in a query's IN clause should return results. It seems like something is getting double escaped somewhere. We are using google's UUIDs which would resolve as an driver.Value of array b/c it's driver value is a string. This results in it flowing throug converter.arrayToString:
[SAM CLI:59345] time="2024-09-17T21:03:11Z" level=debug msg="TYPE: uuid.UUID, 51d3526b-bf57-46bb-8258-1344e6e8d910" func="gosnowflake.(*defaultLogger).Debugf" file="log.go:176"

When I run the UUID through arrayToString I get a bingingValue of json, and the value is a quoted string:
{0x1400022b640 json <nil>}
"49dbf3b3-7a72-42f2-9183-d6e24d00f292"

Is this then getting double escaped somewhere because it thinks the string is json?

   What should have happened and what happened instead?
Queries find the value by string.

8. Can you set logging to DEBUG and collect the logs?
[SAM CLI:59345] time="2024-09-17T21:03:11Z" level=info msg="Exec: \"UPDATE \\\"some_table\\\" SET \\\"is_read\\\"=? WHERE \\\"id\\\" IN (?) AND \\\"some_table\\\".\\\"deleted_at\\\" IS NULL AND \\\"user_id\\\" = ?\", [{ 1 true} { 2 51d3526b-bf57-46bb-8258-1344e6e8d910} { 3 _USERID}]" func="gosnowflake.(*snowflakeConn).ExecContext" file="connection.go:312"

   
9. What is your Snowflake account identifier, if any? (Optional)

ChronosMasterOfAllTime:
### Description

SNOW-1669514 - Solve the issue with UUID getting double quoted due to new array function #1209 

This adds the ability to check the valuer/stringer interfaces exists

### Checklist

Proof test fails.
```sh
export SKIP_SETUP=true && go test -run ^TestValueToString$ github.com/snowflakedb/gosnowflake
--- FAIL: TestValueToString (0.00s)
    --- FAIL: TestValueToString/UUID_-_should_return_string (0.00s)
        assert_test.go:103: expected "json" to be empty, but was not. . Thrown from gosnowflake/converter_test.go:290 +0xd4
            testing.tRunner(0xc0000bf520, 0x100e3a478)
        assert_test.go:103: expected ""1414b0f8-e56a-4e14-995b-774d87ea9d73"" to be equal to "1414b0f8-e56a-4e14-995b-774d87ea9d73" but was not. . Thrown from gosnowflake/converter_test.go:291 +0x197
            testing.tRunner(0xc0000bf520, 0x100e3a478)
```


converter.go
```diff
@@ -401,6 +385,28 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
+		}
+
+		if v, ok := value.(string); ok {
+			return bindingValue{&v, "", nil}, nil
+		}
+
+		return bindingValue{}, nil
+	} else if hasStringMethod(v1) { // alternate approach; check for stringer method. Guarantees it's String() and returns string
+		method := v1.MethodByName("String")
+		result := method.Call(nil) // Call with no arguments
```
AdamDrewsTR, line 402: Will this panic if it has the wrong signature?

ChronosMasterOfAllTime: No; the hasStringMethod validates the IO signature

ChronosMasterOfAllTime: https://github.com/snowflakedb/gosnowflake/pull/1211/files#diff-ecad48054a05ca5aecf42b278530331772747ed7211a193aacd94214cb53aad0R767

ChronosMasterOfAllTime: Simplified it to use `fmt.Stringer` interface check. Resolving.

converter.go
```diff
@@ -385,6 +385,21 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
+		if v, ok := value.(string); ok {
+			return bindingValue{&v, "", nil}, nil
+		}
+
+		return bindingValue{}, nil
+	} else if stringer, ok := v1.Interface().(fmt.Stringer); ok { // alternate approach; check for stringer method. Guarantees it's String() and returns string
+		value := stringer.String()
+		return bindingValue{&value, "", nil}, nil
 	}
 	res, err := json.Marshal(v)
```
AdamDrewsTR, line 404: After this, the bindingValue is set to "json". Should that be the case, or should it be string? Since it is already Marshaled, why set it to json again? That will cause it to be marshaled a second time. It feels like it should not marshal here if it also sets the binding value to json.

converter.go
```diff
@@ -239,11 +240,21 @@ func valueToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
 		return bindingValue{nil, "", nil}, nil
 	}
 	v1 := reflect.Indirect(reflect.ValueOf(v))
+
+	if valuer, ok := v.(driver.Valuer); ok { // check for driver.Valuer satisfaction and honor that first
+		if value, err := valuer.Value(); err == nil && value != nil {
+			// if the output value is a valid string, return that
+			if strVal, ok := value.(string); ok {
+				return bindingValue{&strVal, "", nil}, nil
+			}
```
ChronosMasterOfAllTime, line 254: Leaving this here as a suggestion; did we want to let a marshaler handle the driver's value if not a string by default?

```suggestion
			} else if marshaled, err := json.Marshal(value); err == nil {
			    value := string(marshaled)
			    return bindingValue{&value, jsonFormatStr, nil}, nil
			}
```

converter.go
```diff
@@ -28,6 +28,7 @@ import (
 
 const format = "2006-01-02 15:04:05.999999999"
 const numberDefaultPrecision = 38
+const jsonFormatStr = "json"
```
ChronosMasterOfAllTime, line 31: @sfc-gh-dszmolka  can you please explain how the format string is being used by the ExecBindQuery, or what is the purpose of this field?

sfc-gh-pfus: Great question! This field is extremely important, when dealing with structured objects/arrays/maps. If a field is structured, we need to inform snowflake backend, that it should parse this value and treat it as structured type instead of plain string.

converter.go
sfc-gh-pfus: I'm not sure about that. This is only a writing part, what about reading into same type? I don't think we support such approach at all.

converter.go
sfc-gh-pfus: As mentioned above - this is only a hint for backend, how to treat this string value - either as a string literally or as a structured type described in json.

converter.go
```diff
@@ -375,23 +386,27 @@ func arrayToString(v driver.Value, tsmode snowflakeType, params map[string]*stri
...
+			return bindingValue{&res, jsonFormatStr, &schemaForBytes}, nil
 		}
 		res := "["
 		for _, b := range barr {
 			res += fmt.Sprint(b) + ","
 		}
 		res = res[0:len(res)-1] + "]"
-		return bindingValue{&res, "json", &schemaForBytes}, nil
+		return bindingValue{&res, jsonFormatStr, &schemaForBytes}, nil
+	} else if stringer, ok := v1.Interface().(fmt.Stringer); ok && v1.Kind() == reflect.Array && v1.Type().Elem().Kind() == reflect.Uint8 && v1.Len() == 16 {
```
sfc-gh-pfus, line 397: I don't like such special cases. It may fail depending on the type defined in the table (varchar vs array vs array(byte) vs binary). Can you add such integration tests?

sfc-gh-pfus (commented):
Thank you for your contribution! Can you take a look at my comments?

converter.go
ChronosMasterOfAllTime: @sfc-gh-pfus What's the best approach for adding integration tests? I cannot run those locally, unless your team has a method for running them

converter.go
ChronosMasterOfAllTime: Ok will close  this suggestion out.

converter.go
ChronosMasterOfAllTime: This is testing for a very specific type of `[16]byte`

sfc-gh-pfus: I believe that you have to use your snowflake account to run integration test, or at least use a [trial](https://signup.snowflake.com/). I believe that if you want to add an integration test, you would have to write a test, which creates a table, performs an insert and performs a select.

Yes, this is something I'm afraid of. Let's imagine you have a UUID (which technically is an array of bytes) and you want to insert it into varchar table. With your solution you would use it like this (guessing without integration test):
```
myUuid := NewUUID()
db.Exec("INSERT ... VALUES (?)", myUuid)
```
Am I right? In that case, the actual value goes to backend as a string representation (eg `646313f9-d353-41bc-8ef4-21482bab7020`) and it is properly inserted into varchar column. So far so good.

But now, let's suppose, the column type is not varchar, but ARRAY or BINARY. From the driver perspective, you still send use the same construct:
```
myUuid := NewUUID()
db.Exec("INSERT ... VALUES (?)", myUuid)
```
But if we send it as the same value (`646313f9-d353-41bc-8ef4-21482bab7020`) backend won't be able to interpret it as array (expects `[16,84,21,47,77...]`) nor binary (expects `646313f9d35341bc8ef421482bab7020`).

But I may be wrong and maybe backend will be able to somehow magically parse each option, this is why I suggested adding integration test for all cases :) 

Generally UUID is not something that is supported by this driver (if it worked before it was only by accident) and if we want to add support for it, we should do it globally, not only in this case. Also, we have to think globally - how other drivers should support similar cases.

One more note - 16 elem byte array is not reserved for uuids, it can be also some checksum, something connected to cryptography (keys, IVs etc) and so on.

To solve your particular case - I believe, that if you have a stringer, you can replace `db.Exec("INSERT ...", u)` with `db.Exec("INSERT ..., u.String())`, right? Your option is nicer and I really would like it, if only it wasn't so troublesome :(

ChronosMasterOfAllTime: UUIDs on other drivers (SQLite, Postgres, etc.) is already supported.

Right; this issue more is intrinsic to using an ORM on top of the snowflake driver. Because we cannot control the ORM's behavior how it's going to do the final insert statement, we rely on the driver to do the correct thing. I can remove the UUID check in such a way OR I can add additional logic checks that looks for specific methods being available such as Parse,ParseUUID etc.

sfc-gh-pfus: As far as I know postgres has support for UUID type, whereas in snowflake there is not such a special type :) You can insert uuid to varchar, array or binary columns.

Ok, ORMs are interesting topic. Do you have any particular one in mind? Is there any ORM binding to snowflake which properly generates snowflake query syntax?

My suggestion for now is to either:
1. We can think internally about this topic and present more requirements of how to support UUIDs and you can implement more complete solution :) As I mentioned - it should work for more than a varchar probably, we would have to think if it should be covered in structured types etc. And it should obviously work with reading as well. On your side - your input from ORM use cases would also be valuable in that area, since I don't have any knowledge if gosnowflake works with any ORMs. This is probably faster way.
2. You can work with you SAM/SE or file a feature request via Salesforce and we can plan it for the future. Unfortunately, this way may be much longer, I can't make promises when we can work on it. Priorities are managed mainly by product management, not devs, and there are key factors taken into account like performance, security, number of customers interested etc.

ChronosMasterOfAllTime: I will work on a more complete solution

This behavior was introduced in version 1.11.0 by #1154 

Prior to that we never ran into this issue. Most people would write UUIDs as strings to their DB I would think.

We are using GORM in tandem with https://github.com/gorm-snowflake/gorm-snowflake. This is a repository we manage.

ChronosMasterOfAllTime: @sfc-gh-pfus when you say read/write integration tests, do you mean the structured types tests for read/write or something else?

I also made it so we're abso-positively sure that it's a UUID implementer of RFC 4122. Are you good with this?

sfc-gh-pfus: I mean firstly tests in which you create a table, insert uuid into varchar column and then read it into uuid value in client code again. driver_test.go seems to be a good place for this. Secondly - yes, we should extend structured types tests. This is slightly optional, but if it is possible to add it, it would be great.

Checking your implementation of RFC - it is a great idea! This way we can have only varchar-supported :) So summing things up - your writing part seem mostly fine so far, now let's ensure that reading such UUIDs also work. Good job!

ChronosMasterOfAllTime: Are we OK if I test using this?

https://snowflake.localstack.cloud/introduction/

My team prefers I dont use our real Snowflake database 😄

ChronosMasterOfAllTime:
@sfc-gh-pfus added tests to structured R/W and Driver. Will attempt to run the tests

bind_uploader.go
```diff
@@ -118,7 +118,7 @@ func (bu *bindUploader) createStageIfNeeded() error {
 		return (&SnowflakeError{
 			Number:   code,
 			SQLState: data.Data.SQLState,
-			Message:  err.Error(),
+			Message:  fmt.Sprint(err),
```
sfc-gh-pfus, line 121: Any comment why this is needed?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/aws/smithy-go v1.20.2
 	github.com/gabriel-vasile/mimetype v1.4.2
 	github.com/golang-jwt/jwt/v5 v5.2.1
+	github.com/google/uuid v1.6.0
```
sfc-gh-pfus, line 18: I prefer we didn't include a new library. If it is for tests only - I think we can live with our impl only.

structured_type_read_test.go
```diff
@@ -224,7 +236,8 @@ func TestObjectWithAllTypesAsObject(t *testing.T) {
 	runDBTest(t, func(dbt *DBTest) {
 		dbt.mustExec("ALTER SESSION SET TIMEZONE = 'Europe/Warsaw'")
 		forAllStructureTypeFormats(dbt, func(t *testing.T, format string) {
-			rows := dbt.mustQueryContextT(ctx, t, "SELECT 1, {'s': 'some string', 'b': 1, 'i16': 2, 'i32': 3, 'i64': 9223372036854775807, 'f32': '1.1', 'f64': 2.2, 'nfraction': 3.3, 'bo': true, 'bi': TO_BINARY('616263', 'HEX'), 'date': '2024-03-21'::DATE, 'time': '13:03:02'::TIME, 'ltz': '2021-07-21 11:22:33'::TIMESTAMP_LTZ, 'tz': '2022-08-31 13:43:22 +0200'::TIMESTAMP_TZ, 'ntz': '2023-05-22 01:17:19'::TIMESTAMP_NTZ, 'so': {'s': 'child', 'i': 9}, 'sArr': ARRAY_CONSTRUCT('x', 'y', 'z'), 'f64Arr': ARRAY_CONSTRUCT(1.1, 2.2, 3.3), 'someMap': {'x': true, 'y': false}}::OBJECT(s VARCHAR, b TINYINT, i16 SMALLINT, i32 INTEGER, i64 BIGINT, f32 FLOAT, f64 DOUBLE, nfraction NUMBER(38, 19), bo BOOLEAN, bi BINARY, date DATE, time TIME, ltz TIMESTAMP_LTZ, tz TIMESTAMP_TZ, ntz TIMESTAMP_NTZ, so OBJECT(s VARCHAR, i INTEGER), sArr ARRAY(VARCHAR), f64Arr ARRAY(DOUBLE), someMap MAP(VARCHAR, BOOLEAN))")
+			uid := NewUUID()
```
sfc-gh-pfus, line 239: Nit. Double u?

sfc-gh-pfus:
@ChronosMasterOfAllTime good job! But correct me if I'm wrong, but as far as I can see, there are only tests for embedding uuid in structured types and not as inserts into/selects from a regular column? Can you add such a test case as well (probably somewhere in driver_test.go)?

ChronosMasterOfAllTime:
> @ChronosMasterOfAllTime good job! But correct me if I'm wrong, but as far as I can see, there are only tests for embedding uuid in structured types and not as inserts into/selects from a regular column? Can you add such a test case as well (probably somewhere in driver_test.go)?

There should be a test added in driver_test.go

structured_type_read_test.go
ChronosMasterOfAllTime: was avoiding package naming `uuid` collision with google's UUID

go.mod
ChronosMasterOfAllTime: I was using it as it has the SQL scanner/valuer interfaces; if you're ok with it I can extend your existing UUID definition for testing purposes and add the scanner valuer methods

structured_type_read_test.go
ChronosMasterOfAllTime: Was trying to avoid collision with any packages name `uuid`

go.mod
ChronosMasterOfAllTime: Removed

bind_uploader.go
ChronosMasterOfAllTime: Added a comment because I noticed err will not ever be addressed since there's a bunch of short circuits on err beforehand.

ChronosMasterOfAllTime:
@sfc-gh-pfus need support on this one:
 (I was running structured_write_test.go from the master branch first to ensure it worked correctly)

Getting the following error:
```
--- FAIL: TestBindingObjectWithSchema (0.67s)
    driver_test.go:276: error on exec context [[query too large to print]]: 002040 (42601): SQL compilation error:
        Unsupported data type 'OBJECT(s VARCHAR(16777216), b NUMBER(38,0), i16 NUMBER(38,0), i32 NUMBER(38,0), i64 NUMBER(38,0), f32 FLOAT, f64 FLOAT, nfraction NUMBER(38,9), bo BOOLEAN, bi BINARY(8388608), date DATE, time TIME(9), ltz TIMESTAMP_LTZ(9), ntz TIMESTAMP_NTZ(9), tz TIMESTAMP_TZ(9), so OBJECT(s VARCHAR(16777216), i NUMBER(38,0)), sArr ARRAY(VARCHAR(16777216)), f64Arr ARRAY(FLOAT), someMap MAP(VARCHAR(16777216), BOOLEAN))'.
```

bind_uploader.go
```diff
@@ -111,15 +111,18 @@ func (bu *bindUploader) createStageIfNeeded() error {
...
+		code, atoiErr := strconv.Atoi(data.Code)
+		if atoiErr != nil {
+			return atoiErr
 		}
 		return (&SnowflakeError{
 			Number:   code,
 			SQLState: data.Data.SQLState,
-			Message:  err.Error(),
-			QueryID:  data.Data.QueryID,
+			// This err will always be nil because we are returning after the exec if err is not nil
```
sfc-gh-pfus, line 121: By looking at the code it seems that indeed passing err here is incorrect.  But there is a `data.Message` field which I believe is best here.

driver_test.go
```diff
@@ -34,6 +34,7 @@ var (
 	protocol         string
 	customPrivateKey bool            // Whether user has specified the private key path
 	testPrivKey      *rsa.PrivateKey // Valid private key used for all test cases
+	debugMode        bool
```
sfc-gh-pfus, line 37: I don't think we need it :) Can you remove it?

driver_test.go
```diff
@@ -862,6 +865,108 @@ func testString(t *testing.T, json bool) {
...
+
+// Value implements sql.Valuer so that UUIDs can be written to databases
+// transparently. Currently, UUIDs map to strings. Please consult
+// database-specific driver documentation for matching types.
+func (uuid testUUID) Value() (driver.Value, error) {
+	return uuid.String(), nil
+}
+
+func TestUUID(t *testing.T) {
+	testUuid(t, false)
```
sfc-gh-pfus, line 929: How about testing with `true`? It should work with arrow response as well. So maybe `TestUUIDInJson` and `TestUUIDInArrow`?

driver_test.go
```diff
@@ -862,6 +865,108 @@ func testString(t *testing.T, json bool) {
...
+		if json {
+			dbt.mustExec(forceJSON)
+		}
+
+		types := []string{"CHAR(255)", "VARCHAR(255)", "TEXT", "STRING"}
+
+		in := make([]testUUID, len(types))
+
+		for i := range types {
+			in[i] = newTestUUID()
```
sfc-gh-pfus, line 943: This reminded me about one more test case we need - can we have a test that inserts 100k UUIDs and uses `sf.Array`? It is a bit different flow, it goes through a stage bind.

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/goccy/go-json v0.10.2 // indirect
 	github.com/godbus/dbus v0.0.0-20190726142602-4481cbc300e2 // indirect
 	github.com/google/flatbuffers v23.5.26+incompatible // indirect
+	github.com/google/uuid v1.6.0 // indirect
```
sfc-gh-pfus, line 41: Some leftover?

sfc-gh-pfus:
> @sfc-gh-pfus need support on this one: (I was running structured_write_test.go from the master branch first to ensure it worked correctly)
> 
> Getting the following error:
> 
> ```
> --- FAIL: TestBindingObjectWithSchema (0.67s)
>     driver_test.go:276: error on exec context [[query too large to print]]: 002040 (42601): SQL compilation error:
>         Unsupported data type 'OBJECT(s VARCHAR(16777216), b NUMBER(38,0), i16 NUMBER(38,0), i32 NUMBER(38,0), i64 NUMBER(38,0), f32 FLOAT, f64 FLOAT, nfraction NUMBER(38,9), bo BOOLEAN, bi BINARY(8388608), date DATE, time TIME(9), ltz TIMESTAMP_LTZ(9), ntz TIMESTAMP_NTZ(9), tz TIMESTAMP_TZ(9), so OBJECT(s VARCHAR(16777216), i NUMBER(38,0)), sArr ARRAY(VARCHAR(16777216)), f64Arr ARRAY(FLOAT), someMap MAP(VARCHAR(16777216), BOOLEAN))'.
> ```

I tried it locally and it worked. I started build on Github Actions, let's check it. By looking at the message - I believe it because you didn't add this UUID to object definition in table, but in the current version it is present, so let's hope it works.

ChronosMasterOfAllTime:
What I mean is I tried this on a clean branch and got the same error. Is it something to do with my snowflake instance? (I finally got the ability to test against our QA env)

sfc-gh-pfus:
So probably yes, probably you don't have access to some features. But it works on CI, so it is good enough.

go.mod
ChronosMasterOfAllTime: Fixed after rebase

bind_uploader.go
ChronosMasterOfAllTime: ok cool that answers my question in the comment 😄

driver_test.go
ChronosMasterOfAllTime: I was using this to print the full query in driver_test as an override to the `query too large to print` locally. You should only ever get this behavior by setting `SNOWFLAKE_TEST_DEBUG=true`

driver_test.go
ChronosMasterOfAllTime: can you provide an example test use case where I can pattern?

bindings_test.go
```diff
@@ -382,6 +383,62 @@ func TestBindingInterfaceString(t *testing.T) {
...
+			var (
+				id  int
+				out testUUID
+			)
+			if err := rows.Scan(&id, &out); err != nil {
+				t.Fatal(err)
+			}
+
+			var found bool
+			for _, u := range uuids {
```
sfc-gh-pfus, line 428: This makes it `O(n^2)` complexity :) Two suggestions to fix this - either change the table to contain incremented value and order by id instead of UUIDs or sort `uuids` array :)

sfc-gh-pfus (commented):
One last comment and we are good to go!

bindings_test.go
ChronosMasterOfAllTime: This is fixed @sfc-gh-pfus

sfc-gh-pfus (approved):

sfc-gh-pfus:
@ChronosMasterOfAllTime I started tests and some of them failed. Mostly because of the syntax (a colon instead of a comma), but some failed on incorrectly reading UUIDs:

```
=== RUN   TestObjectWithAllTypesSimpleScan/JSON
    assert_test.go:103: expected "00000000-0000-0000-0000-000000000000" to be equal to "d0af8810-9997-42e9-5ff1-8101a9d79732" but was not. . Thrown from /home/runner/work/gosnowflake/gosnowflake/structured_type_read_test.go:600 +0x105c
        github.com/snowflakedb/gosnowflake.forAllStructureTypeFormats.func4(0x0?)
```

There are still some flaky tests, so ignore them for now, but these one that appear in each result should be fixed :)
Also, our linter failed - but only with minor suggestions.

ChronosMasterOfAllTime:
> > Where can I see the results and where can I see the linter checks?
> 
> Below the list of the comments and above "add a comment" section on this page. I checked that it is visible for non-Snowflake users.

Now I see the checks. Thank you!

ChronosMasterOfAllTime:
@sfc-gh-pfus I was out last week and taking a look now

sfc-gh-pfus:
Hi, I had an impression these parameters should be enabled by default now for everyone. Can you check if the tests pass without enabling them? In the meantime - I started an initiative to verify these params on our side.

ChronosMasterOfAllTime:
> Hi, I had an impression these parameters should be enabled by default now for everyone. Can you check if the tests pass without enabling them? In the meantime - I started an initiative to verify these params on our side.

Tests will not pass because it returns all objects as strings to the driver and the Fields for each row will not be set. Because the session cannot be set it doesn't set the context for it either (verified with delve).

sfc-gh-pfus:
Sure, so these params are indeed needed. I'll let you know my findings if these parameters should be publicly enabled or not and what are the next steps.

sfc-gh-pfus:
Hi @ChronosMasterOfAllTime ! I verified with proper teams that structured types should not be publicly available yet. I fixed the tests that failed on your PR. I cannot apply these changes to your PR directly (as it is in your repo), so can you take them and apply in your PR?
https://github.com/snowflakedb/gosnowflake/pull/1263/commits/1ca7cc389a777be347c04d81a298b4a014833e5a

sfc-gh-pfus:
@ChronosMasterOfAllTime we are really close! Only our linter complains about two names, can you fix it? You can ignore Jenkins checks for now (we have some internal problems with our infra).

ChronosMasterOfAllTime:
@sfc-gh-pfus doh! linter errors fixed

sfc-gh-dheyman (approved):

ChronosMasterOfAllTime:
@sfc-gh-dheyman I ran the linter locally and noticed the `fmt` job had some whitespace changes; added in this commit https://github.com/snowflakedb/gosnowflake/pull/1211/commits/b07b7096a80318773f49005bee8481f9022a1149

sfc-gh-pfus:
Part of the journey is the end. @ChronosMasterOfAllTime, thank you very much for all the work you put into this PR and for your huge amount of patience!

PR #1212, opened 2024-09-20
[SNOW-1671768] Ignore max retries error when getting acelerate config

sfc-gh-astachowski:
### Description

Ignore and log all errors thrown while getting accelerate configuration from bucket.

### Checklist

### DIFF
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"io"
 	"math"
 	"net/url"
@@ -628,6 +629,10 @@ func (sfa *snowflakeFileTransferAgent) transferAccelerateConfig() error {
 					// https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-s3.html
 					return nil
 				}
+			}
+			var mae *retry.MaxAttemptsError
+			if errors.As(err, &mae) {
+				return nil
 			}
 			return err
 		}
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -9,9 +9,7 @@ import (
 	"context"
 	"database/sql/driver"
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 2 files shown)


sfc-gh-hyu (commented):
My general preference is that: we only enable bucket acceleration configuration is we are able to get bucket configuration and bucket configuration indeed returns that acceleration is enabled for the bucket, in all other scenarios, whatever error returned by http, we just fallback to treating that bucket acceleration is not enabled. This is a much robust way. And this is how python connector handler the error as well. Check the code here https://github.com/snowflakedb/snowflake-connector-python/blob/3c8255935b9aebc2edee8de3913959f86596739f/src/snowflake/connector/s3_storage_client.py#L583

sfc-gh-astachowski:
@sfc-gh-hyu Ok, changed to ignore all errors in that part of the code, logging them all at warn level.

sfc-gh-pmotacki (approved):
LGTM

sfc-gh-hyu (approved):

file_transfer_agent.go
```diff
@@ -591,46 +589,47 @@ func (sfa *snowflakeFileTransferAgent) updateFileMetadataWithPresignedURL() erro
...
+			QueryID:  sfa.data.QueryID,
+			Message:  errMsgFailedToConvertToS3Client,
+		}).exceptionTelemetry(sfa.sc)
+	}
+	ret, err := client.GetBucketAccelerateConfiguration(context.Background(), &s3.GetBucketAccelerateConfigurationInput{
+		Bucket: &s3Loc.bucketName,
+	})
+	sfa.useAccelerateEndpoint = ret != nil && ret.Status == "Enabled"
+	if err != nil {
+		logger.WithContext(sfa.sc.ctx).Warnln("An error occurred when getting accelerate config:", err)
```
sfc-gh-mkubik, line 624: Let's update the PR description to reflect that all errors are being logged and ignored instead of just the old set of ignorable errors + max retries 

> Added max retries error to errors ignored when getting acceleration config.

sfc-gh-mkubik (approved):

PR #1220, opened 2024-10-11
NO-SNOW fix filestransfer example

sfc-gh-pfus:
### Description

NO-SNOW Fixed filestransfer example - incorrect destination file path.

### DIFF
--- a/cmd/filestransfer/filestransfer.go
+++ b/cmd/filestransfer/filestransfer.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 
 	sf "github.com/snowflakedb/gosnowflake"
 )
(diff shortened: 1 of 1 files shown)


cmd/filestransfer/filestransfer.go
```diff
@@ -133,7 +134,7 @@ func main() {
 	fmt.Printf("File successfully downloaded from internal stage area to %v\n", tmpDir)
 
 	//Reading from downloaded file
-	file, err := os.Open(fmt.Sprintf("%v/data_to_upload.csv.gz", tmpDir))
+	file, err := os.Open(fmt.Sprintf("%v/%v.gz", tmpDir, filepath.Base(tmpFilePath)))
```
sfc-gh-mkubik, line 137: would it make sense to do some `filepath.Join` instead of formatting with `/` since we already import the filepath module or it's irrelevant as `os.Open` is already platform agnostic?

sfc-gh-mkubik (approved):

cmd/filestransfer/filestransfer.go
sfc-gh-pfus: Good idea, fixed.

sfc-gh-pmotacki (approved):

sfc-gh-dheyman-1 (approved):

PR #1222, opened 2024-10-15
SNOW-1661955 extend logs for QueryArrowStream

sfc-gh-pfus:
### Description

SNOW-1661955 Extend logs for QueryArrowStream

### Checklist

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -187,7 +187,7 @@ func (sc *snowflakeConn) exec(
 		}
 	}
 
-	logger.WithContext(ctx).Info("Exec/Query SUCCESS")
+	logger.WithContext(ctx).Infof("Exec/Query SUCCESS with total=%v, returned=%v", data.Data.Total, data.Data.Returned)
 	if data.Data.FinalDatabaseName != "" {
 		sc.cfg.Database = data.Data.FinalDatabaseName
 	}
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -751,12 +756,14 @@ func (scd *snowflakeArrowStreamChunkDownloader) GetBatches() (out []ArrowStreamB
...
 			scd:     scd,
 		}
+		logger.Debugf("batch %v, numrows: %v", i, toFill[i].numrows)
 		totalCounted += int64(scd.ChunkMetas[i].RowCount)
 	}
 
 	if len(rowSetBytes) > 0 {
 		// if we had a first batch, fill in the numrows
 		out[0].numrows = scd.Total - totalCounted
+		logger.Debugf("first batch, numrows: %v", out[0].numrows)
```
sfc-gh-astachowski, line 766: I feel like this log combined with the one in the loop above is a bit confusing, with "first batch" being the last one logged and distinct from "batch 0". Can we move them around a bit or at least change the message to be clearer?

sfc-gh-pfus: Hmmm, first batch is a standard name in all drivers to describe content received with original response. This one is just special. I believe we can leave it as is.

sfc-gh-astachowski (approved):

sfc-gh-dheyman-1 (approved):

PR #1228, opened 2024-10-23
Add golangci-lint

sfc-gh-pfus:
### Description

### Checklist


driver.go
```diff
@@ -83,9 +85,9 @@ func init() {
 	if !skipRegisteration() {
```
sfc-gh-dszmolka, line 85: do we want to fix this typo as well perhaps? :)

sfc-gh-pfus: Well caught! So I fixed the function name, but the same typo exists in the env var, which I can't fix :(

encrypt_util.go
```diff
@@ -272,7 +288,9 @@ func decryptStreamCBC(
 		totalFileSize += n
 		chunk = chunk[:n]
 		mode.CryptBlocks(chunk, chunk)
-		out.Write(chunk)
+		if _, err = out.Write(chunk); err != nil {
+			return 0, err
```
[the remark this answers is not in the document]
sfc-gh-pfus: I added a doc, but this is a general notice, tbh.

authexternalbrowser.go
```diff
@@ -311,10 +310,17 @@ func doAuthenticateByExternalBrowser(
 			buf.Grow(bufSize)
 		}
 		if encodedSamlResponse != "" {
-			httpResponse := buildResponse(application)
-			c.Write(httpResponse.Bytes())
+			httpResponse, err := buildResponse(application)
+			if err != nil && errAccept == nil {
+				errAccept = err
+			}
+			if _, err = c.Write(httpResponse.Bytes()); err != nil {
```
sfc-gh-astachowski, line 317: Why do we want to overwrite `errAccept` in this case, but not in the previous one? (there is an `errAccept == nil` check in the previous `if`)

connection_test.go
```diff
@@ -291,8 +291,9 @@ func TestServiceName(t *testing.T) {
 
 	expectServiceName := serviceNameStub
 	for i := 0; i < 5; i++ {
-		sc.exec(context.Background(), "", false, /* noResult */
+		_, err := sc.exec(context.Background(), "", false, /* noResult */
```
sfc-gh-astachowski, line 294: The comma should probably be after the comment

file_transfer_agent_test.go
```diff
@@ -846,21 +853,26 @@ func TestReadonlyTmpDirPathShouldFail(t *testing.T) {
...
-	f.WriteString("test1,test2\ntest3,test4\n")
-	f.Close()
+	_, err = f.WriteString("test1,test2\ntest3,test4\n")
+	assertNilF(t, err)
+	assertNilF(t, f.Close())
 
-	err = os.Chmod(tmpDir, 0400)
+	err = os.Chmod(tmpDir, 0500)
 	if err != nil {
 		t.Fatalf("cannot mark directory as readonly: %v", err)
```
sfc-gh-astachowski, line 871: This message should be updated

file_transfer_agent_test.go
```diff
@@ -968,7 +981,9 @@ func testUploadDownloadOneFile(t *testing.T, isStream bool) {
 	if isStream {
 		fileStream, _ := os.Open(uploadFile)
 		ctx := WithFileStream(context.Background(), fileStream)
-		uploadMeta.srcStream = getFileStream(ctx)
+		if uploadMeta.srcStream, err = getFileStream(ctx); err != nil {
+			assertNilF(t, err)
```
sfc-gh-astachowski, line 985: We can skip the `if` and just do the assertion

file_util.go
```diff
@@ -98,8 +117,10 @@ func (util *snowflakeFileUtil) getDigestAndSizeForFile(fileName string) (string,
 		total += int64(n)
 		m.Write(chunk[:n])
 	}
-	f.Seek(0, io.SeekStart)
-	return base64.StdEncoding.EncodeToString(m.Sum(nil)), total, nil
+	if _, err = f.Seek(0, io.SeekStart); err != nil {
+		return "", -1, err
+	}
+	return base64.StdEncoding.EncodeToString(m.Sum(nil)), total, err
```
sfc-gh-astachowski, line 123: if `err != nil` we won't reach this code, so might as well return `nil` for clarity

local_storage_client_test.go
```diff
@@ -75,7 +76,9 @@ func TestLocalUpload(t *testing.T) {
 	}
 	fileStream, _ := os.Open(fname)
 	ctx := WithFileStream(context.Background(), fileStream)
-	uploadMeta.srcStream = getFileStream(ctx)
+	if uploadMeta.srcStream, err = getFileStream(ctx); err != nil {
```
sfc-gh-astachowski, line 79: Another unnecessary `if`

storage_client.go
```diff
@@ -145,7 +145,8 @@ func (rsu *remoteStorageUtil) uploadOneFileWithRetry(meta *fileMetadata) error {
 		if meta.resStatus == uploaded || meta.resStatus == skipped {
 			for j := 0; j < 10; j++ {
 				status := meta.resStatus
-				utilClass.getFileHeader(meta, meta.dstFileName)
+				_, err := utilClass.getFileHeader(meta, meta.dstFileName)
+				logger.Infof("error while getting file %v header. %v", meta.dstFileSize, err)
```
sfc-gh-astachowski, line 149: Missing `if err != nil` I think?

transaction_test.go
```diff
@@ -60,7 +64,6 @@ func TestTransactionContext(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer tx.Rollback()
```
sfc-gh-astachowski, line 63: Why delete this?

authexternalbrowser.go
sfc-gh-pfus: Very good catch, indeed! Fixed.

connection_test.go
sfc-gh-pfus: Fixed.

file_transfer_agent_test.go
sfc-gh-pfus: Actually this is still valid. The only case is that we need `-x` to not only read files in a dir, but also to access files inside it. But the main case is still - this dir is readonly, not writable.

file_transfer_agent_test.go
sfc-gh-pfus: Right, done.

file_util.go
sfc-gh-pfus: Ha, definitely not :) This is something called naked return - we name variables that are returned. Without that, if we rich this line 123, err set in a deferred block would be swallowed.

storage_client.go
sfc-gh-pfus: Right, added.

transaction_test.go
sfc-gh-pfus: Because the transaction is committed a few lines below.

file_util.go
sfc-gh-astachowski: Oh, that makes a lot of sense. I stand corrected 😄

connection_test.go
sfc-gh-astachowski: I don't think it is

sfc-gh-dheyman-1 (approved):

sfc-gh-pbulawa (approved):

PR #1245, opened 2024-11-16
Driver test to read parameters file

rickr-sigma:
### Description

The README states to use a parameters.json file for the test connection configuration.  This works fine when running `make test`, but if you want to run a test in an ide or individual ones this file is ignored and you need to manually set the env vars.

### Checklist

### DIFF
--- a/driver_test.go
+++ b/driver_test.go
@@ -7,6 +7,7 @@ import (
 	"crypto/rsa"
 	"database/sql"
 	"database/sql/driver"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


sfc-gh-pfus:
Hi @rickr-sigma ! Thanks for expressing your interest in that area. Unfortunately, we don't want to change it. Readme states that this file is used for `make test` and it is fine. It is the same way as it works in all our drivers. Sorry for the inconvenience that it makes for you!

PR #1278, opened 2024-12-16
SNOW-1855886 gracefully ignore when default region us-west-2 is used in DSN or NewConnector

sfc-gh-dszmolka:
### Description

`us-west-2` region is the default region in Snowflake, and a special one. We already had checks for the case where someone accidentally configures it in `Region`, but for `Account` we did not have any, which led to an `Account` (mis)configured as `myaccount.us-west-2` instead of omitting the region (as is the default) and using `myaccount`, not being able to connect to Snowflake and instead failing with HTTP404.

This change attempts to step over such simple misconfiguration gracefully and 
* ignores the extra `.us-west-2` from Account instead of allowing the connection to proceed, which then surely will fail on the Snowflake backend 
* logs a line about the misconfiguration on Info level, giving it a chance for the user to notice and correct it

### Checklist

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -139,11 +139,17 @@ func (c *Config) ocspMode() string {
 
 // DSN constructs a DSN for Snowflake db.
 func DSN(cfg *Config) (dsn string, err error) {
-	if cfg.Region == "us-west-2" {
+	if strings.ToLower(cfg.Region) == "us-west-2" {
 		cfg.Region = ""
+		logger.Info("Ignoring Region configuration in DSN as us-west-2 is the default one.")
 	}
 	// in case account includes region
 	region, posDot := extractRegionFromAccount(cfg.Account)
+	if strings.ToLower(region) == "us-west-2" {
+		region = ""
+		cfg.Account = cfg.Account[:posDot]
+		logger.Info("Ignoring default region .us-west-2 in DSN from Account configuration.")
+	}
 	if region != "" {
 		if cfg.Region != "" {
 			return "", errRegionConflict()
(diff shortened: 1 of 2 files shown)


dsn.go
```diff
@@ -139,11 +139,17 @@ func (c *Config) ocspMode() string {
 
 // DSN constructs a DSN for Snowflake db.
 func DSN(cfg *Config) (dsn string, err error) {
-	if cfg.Region == "us-west-2" {
+	if strings.ToLower(cfg.Region) == "us-west-2" {
 		cfg.Region = ""
+		logger.Info("Ignoring Region configuration in DSN as us-west-2 is the default one.")
```
sfc-gh-pfus, line 144: I don't think it is worth being logged at info, debug should suffice.

sfc-gh-pfus (approved):

dsn.go
sfc-gh-dszmolka: the goal here is for the user is to see it and correct it to get rid of the log message. 
initially i thought debug too, but then it would be probably hidden and never corrected.

dsn_test.go
```diff
@@ -1233,6 +1233,31 @@ func TestDSN(t *testing.T) {
...
+				Account:  "account_us-west-2",
+			},
+			dsn: "u:p@account_us-west-2.snowflakecomputing.com:443?ocspFailOpen=true&validateDefaultParameters=true",
+		},
+		{
+			cfg: &Config{
+				User:     "u",
+				Password: "p",
+				Account:  "account",
+				Host:     "account.us-west-2.snowflakecomputing.com",
```
sfc-gh-dprzybysz, line 1257: we support also `account.us-west-2.aws` - there also dsn should be shortened?

dsn.go
sfc-gh-pfus: And I agree with the cases below when customer includes region in account name, but here I'd say it is fine to give specific region as a separate parameter. But I'm not pushing you, I'll let you decide the final option.

dsn_test.go
sfc-gh-dszmolka: then the next question is that we should correct every notation in every driver, which i agree is the ideal final goal

dsn.go
sfc-gh-dszmolka: removed this logline for when cfg.Region is us-west-2

dsn_test.go
sfc-gh-dszmolka: added `SNOW-1858448` to track the more comprehensive effort

sfc-gh-dprzybysz (approved):

Issue #616, opened 2022-07-11
Make logrus dependency optional or drop it all together


Thomasdezeeuw:
The dependency on logrus creates two problems:
1.  I'm currently getting a lot of logs in a different format than my log parser is expected by using Snowflake.
2. Creating a customer logging implementation that doesn't use logrus is basically not possible.

Would it be possible to make the logrus dependency optional or maybe remove it all together?

Thomasdezeeuw:
Any feedback on this? If accept I could start a pr.

sfc-gh-dszmolka:
hi, thank you for submitting this request with us. we'll take a look - as with other enhancements, can't promise any timeline on the ETTR so thank you for bearing with us !
a PR of course would be very helpful to get things moving ; however i'm not sure if simply removing `logrus` would be feasible here or more like a breaking change.

bombsimon:
I would also be interested in dropping logging. I don't think it's very common to do arbitrary logging from libraries like this. I think removing the logging and propagating errors to the caller would make more sense. If it's not an error I don't think it needs to be logged.

If however you still want to use logging from this codebase, maybe considering something more flexible like [logr](https://github.com/go-logr/logr#implementations-non-exhaustive) which has an extensive list of implementations or create your own slimmed down interface uses can implement themselves (info- and error severity should be enough).

And if that also isn't of interest, maybe you  could expose the global logger [here](https://github.com/snowflakedb/gosnowflake/blob/4bc6cd40d71eb854e53fe6e7a65261f184e43865/driver.go#L57) so users can overwrite it?

sfc-gh-pfus:
Hi all! We started considering this issue! 

I have a strong Java background and in Java we have a wonderful SLF4J library which is de facto a standard for Java logging API. SLF4J is not a logger itself, it is more like an API which is implemented by multiple logging libraries like logback or log4j. Still, this is great, because all libraries use only SLF API and not come with logging backend of their choice. Not sure if this is it, but it seems that in Go similar library is https://github.com/go-logr/logr

Another solution might be to use https://go.dev/blog/slog . Unfortunately, this is brand new thing, which was introduced in Go 1.21. We have to support older go versions for a few couple of quarters (fingers crossed that less than more).

If you are interested in such changes, please add a reaction.
For those in favour in going to logr - react with 🚀 
For those in favour in going to slog (will take more time) - react with 🎉 
For those in favour of something else - please add your suggestions.

Thanks in advance!

[...]

PR #1286, opened 2025-01-06
SNOW-625282: Make logrus dependency optional or drop it all together

rickr-sigma:
### Description

The following changes decouple logrs from SFLogger, with the eventual goal to allow per context logging control (in a followup PR) by a user provided SFLogger. 

This PR is not (yet) complete but is presented as a proof-of-concept.  Further cleanup will be performed if this approach is deemed acceptable.  One thing to be aware of is that certain interfaces are no longer supported in particular the functional variants of logging, and the WithTime, WithFields, WithError, Panic and Fatal which were all ONLY being exercised in unit tests. 

### Checklist


driver_test.go
```diff
@@ -50,11 +51,16 @@ const (
...
+	params := parseParameters("parameters.json", "testconnection")
+
 	// get environment variables
 	env := func(key, defaultValue string) string {
 		if value := os.Getenv(key); value != "" {
 			return value
 		}
+		if value, ok := params[key].(string); ok && value != "" {
+			return value
+		}
```
rickr-sigma, line 63: this change is not necessary for the PR but it does allow parameters to be read from the json file as per the README.

easy_logging_test.go
```diff
@@ -149,7 +149,6 @@ func TestLogToConfiguredFile(t *testing.T) {
 
 	logger.Error("Error message")
 	logger.Warn("Warning message")
-	logger.Warning("Warning message")
```
rickr-sigma, line 152: Warning is basically an alias for Warn so removing it since it clutters the interface otherwise.

log.go
```diff
@@ -257,88 +238,12 @@ func (log *defaultLogger) Warn(args ...interface{}) {
...
-	if log.enabled {
-		log.inner.Panic(args...)
-	}
-}
-
-func (log *defaultLogger) TraceFn(fn rlog.LogFunction) {
-	if log.enabled {
-		log.inner.TraceFn(fn)
-	}
-}
```
rickr-sigma, line 288: all Fn variants removed; again unused in prod code.

log.go
```diff
@@ -369,69 +274,17 @@ func (log *defaultLogger) Warnln(args ...interface{}) {
...
-
 // SetOutput sets the logger output.
 func (log *defaultLogger) SetOutput(output io.Writer) {
 	log.inner.SetOutput(output)
 }
 
-func (log *defaultLogger) SetReportCaller(reportCaller bool) {
-	log.inner.SetReportCaller(reportCaller)
-}
-
```
rickr-sigma, line 434: SetFormatter / SetReportCaller only have 1 usage and so can be made internal.

rickr-sigma:

log.go
```diff
@@ -40,12 +39,39 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	Println(args ...interface{})
+	Warnln(args ...interface{})
+	Errorln(args ...interface{})
+}
+
 // SFLogger Snowflake logger interface to expose FieldLogger defined in logrus
 type SFLogger interface {
-	rlog.Ext1FieldLogger
+	LogEntry
+	WithContext(ctx ...context.Context) LogEntry
```
sfc-gh-pfus, line 71: Is it going to be backward compatible? These are exported objects, so we should keep backward compatibility as strictly as we can.

log.go
```diff
@@ -56,6 +82,8 @@ func SFCallerPrettyfier(frame *runtime.Frame) (string, string) {
 	return path.Base(frame.Function), fmt.Sprintf("%s:%d", path.Base(frame.File), frame.Line)
 }
 
+var _ SFLogger = &defaultLogger{} // ensure defaultLogger isa SFLogger.
```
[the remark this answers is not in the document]
rickr-sigma: its a no-op in the final code, but allows the compiler to check if defaultLogger implements all the methods of SFLogger.  https://go.dev/doc/effective_go#blank_implements

log.go
rickr-sigma: Good point.  The following methods will prevent us from de-coupling;
```
	WithField(key string, value interface{}) *Entry
	WithFields(fields Fields) *Entry
	WithError(err error) *Entry
```
since they return a struct.  Are we ok with NOT supporting them?

easy_logging_test.go
rickr-sigma: comments indicate we should maintain backwards compatibility as much as possible so adding these back.

log.go
```diff
@@ -195,38 +202,12 @@ func (log *defaultLogger) Warnf(format string, args ...interface{}) {
...
-	if log.enabled {
-		log.inner.Fatalf(format, args...)
-	}
-}
-
-func (log *defaultLogger) Panicf(format string, args ...interface{}) {
-	if log.enabled {
-		log.inner.Panicf(format, args...)
-	}
-}
```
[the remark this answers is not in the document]
rickr-sigma: as per backward compatibility comment these have been added back.

driver_test.go
rickr-sigma: removed this change

log.go
```diff
@@ -40,12 +40,52 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+type LogEntry interface {
+	Tracef(format string, args ...interface{})
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Printf(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Panicf(format string, args ...interface{})
```
rickr-sigma, line 55: actually thinking about this further I don't think we want to expose/support Fatal and Panic;  logrus has code that calls os.Exit by default which will pull down the entire process.  As a driver integrating with other code this would be bad behavior.

sfc-gh-pfus: As long as I agree with you, I think we should keep them for backward compatibility. But I added the idea to our list of BCRs - when we finally release v2, I'd like to have all these functions removed, and leave only logging configuration (like setting levels etc).

log.go
```diff
@@ -40,12 +40,52 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	LogEntry
+	WithField(key string, value interface{}) LogEntry
+	WithFields(fields Fields) LogEntry
+	WithError(err error) LogEntry
+	WithTime(t time.Time) LogEntry
+
 	SetLogLevel(level string) error
 	GetLogLevel() string
-	WithContext(ctx context.Context) *rlog.Entry
+	WithContext(ctx ...context.Context) LogEntry
```
sfc-gh-pfus, line 88: It is backward compatible, isn't it? Previously this function returned a struct, now it returns the interface which is implemented by logrus struct, right?

rickr-sigma: This is probably the biggest change.  Every field in the rlog.Entry struct is exported and modifiable.  With a LogEntry you can only emit a log line via trace/debug/info/... .  

The rationale behind this change is that ultimately we want to be able to have callers of the driver be able to have full control over the logger and this is only possible when interfaces are exposed, or at the very least its a lot easier. 

If we're still looking for a means of exposing the underlying logrus Entry we could ensure that a cast from LogEntry works. I'll wire up a unit test to make sure this is straightforward enough.   It doesn't provide 100% code compatibility, but at least it provides an out for anyone that has directly manipulated the logrus Entries.

log.go
rickr-sigma: added them back for now.

log.go
rickr-sigma: Added ConvertibleEntry interface so that callers can convert the underlying LogEntry into a logrus Entry struct.  Test case TestLogEntryConversion() shows how to do this.

rickr-sigma:
@sfc-gh-pfus please have another look, as this version now maximizes backwards compatibility.   

The breaking change is now narrowed down to WithContext/WithField/WithFields/WithError which now return LogEntry rather than the underlying logrus Entry struct.   

The function ToEntry() on the ConvertibleEntry interface can be used to obtain the wrapped logrus Entry struct from a LogEntry but it requires a change on the caller side (i.e. 1 additional method call).  See TestLogEntryConversion for an example.

log.go
```diff
@@ -40,12 +40,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
 // logger.WithContext is used
 var LogKeys = [...]contextKey{SFSessionIDKey, SFSessionUserKey}
 
-// SFLogger Snowflake logger interface to expose FieldLogger defined in logrus
+// Fields
```
sfc-gh-pfus, line 43: Any comment? Why do we need another public var?

rickr-sigma: LogKeys ?  It was added in 22Dec20 looks like a convenience for the iteration.  At the very least we could make it package local or even fold it into the context2Fields() func, which contains the only use.

sfc-gh-pfus: No, I meant `Fields` :)

rickr-sigma: the Fields type matches the definition of logrus Fields and exists for backward compatibility.  I.e. the existing withXXX methods on FieldLogger are; 
```
	WithField(key string, value interface{}) *Entry
	WithFields(fields Fields) *Entry
	WithError(err error) *Entry
```
so keeping the interfaces as close as possible we have SFLogger containing;
```
	WithField(key string, value interface{}) LogEntry
	WithFields(fields Fields) LogEntry
	WithError(err error) LogEntry
```

sfc-gh-pfus: Ok, so just leave a comment above the var :) Also, if we want to release it as a non compatible change, we don't have to write in a comment that it is a backward compatibility to logrus.

Regardless the variable itself, I see two use cases.
1. Global map with fields actually global, like instance name/host, OS or app version - that is fine.
2. `WithFields` used to propagate fields to the actual log entry. I can't see a good examples, but my intuition says that in some cases it may be helpful, so probably also good to have it.

frbvianna:
Hi @rickr-sigma, do you plan to continue working on this PR? Looks to be in a good direction for https://github.com/snowflakedb/gosnowflake/issues/616

rickr-sigma:
> Hi @rickr-sigma, do you plan to continue working on this PR? Looks to be in a good direction for #616

Absolutely. I'll sync and resolve the conflicts sometime this week.

PR #1288, opened 2025-01-12
SNOW-1859664 use correct transport for calling cloud providers

sfc-gh-dszmolka:
### Description
See linked issue. The problem is, that since 1.7.2 even though the user specifies `insecureMode` (pre-1.12.1) and/or `disableOCSPchecks` (post-1.12.1); the client we create for executing cloud storage operations; always uses `SnowflakeTransport` regardless of the OCSP verification settings. 
This Transport always performs OCSP validation, regardless of the user's wishes, whether they want OCSP validation to be performed or not.

This is unexpected. What is even worse if there's any configuration issues on the user's infrastructure (e.g. the privatelink OCSP endpoint isn't configured in the DNS), we will still try the OCSP validation and even with the new OCSP strategy; with privatelink (since the 'OCSP Responder' is in this case, behind the Snowflake account), we'll retry a lot of times adding extra minutes of delay before failing with mysterious error messages like
```
operation error S3: PutObject, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequestID: , HostID: , request send failed, Put <remoteLocation>?x-id=PutObject\": EOF"
```

and since 1.7.2 due to improved error handling, we actually stopped ignoring these errors and are now surfacing to the user so they can fix it. But performing OCSP in an environment which is explicitly configured not to perform OCSP was not intended.

The change here aims to make the `Config` part of the cloud storage client operation; use a Transport depending on the `Config`, thus the OCSP validation not performed if the user explicitly configured the driver with `disableOCSPChecks` / `insecureMode`.

### Checklist


azure_storage_client.go
```diff
@@ -45,14 +45,15 @@ func (util *snowflakeAzureClient) createClient(info *execResponseStageInfo, _ bo
 	if err != nil {
 		return nil, err
 	}
+	transport := getTransport(util.cfg)
```
sfc-gh-pfus, line 48: Maybe we can inline this?

s3_storage_client.go
```diff
@@ -49,6 +49,7 @@ func (util *snowflakeS3Client) createClient(info *execResponseStageInfo, useAcce
 	stageCredentials := info.Creds
 	s3Logger := logging.LoggerFunc(s3LoggingFunc)
 	endPoint := getS3CustomEndpoint(info)
+	transport := getTransport(util.cfg)
```
sfc-gh-pfus, line 52: Maybe we can just call `createClientWithConfig(info, useAccelerateEndpoint, util.cfg)` instead of copying function body?

util.go
```diff
@@ -348,3 +349,16 @@ func findByPrefix(in []string, prefix string) int {
 	}
 	return -1
 }
+
+func getTransport(cfg *Config) *http.Transport {
+	if cfg == nil {
```
sfc-gh-pfus, line 354: Is it fine to hot have cfg provided here? If not, maybe it is worth logging on warn, not on debug?

util_test.go
```diff
@@ -420,3 +421,47 @@ func TestFindByPrefix(t *testing.T) {
 	assertEqualE(t, findByPrefix(nonEmpty, "dd"), -1)
 	assertEqualE(t, findByPrefix([]string{}, "dd"), -1)
 }
+
+type tcConfigWithTransport struct {
```
sfc-gh-pfus, line 425: Can we inline it?

azure_storage_client.go
sfc-gh-dszmolka: sure, inlined all occurrences

util.go
sfc-gh-dszmolka: originally the function was prepared for `nil` config to deal with the 'speciality' of `snowflakeFileTransferAgent' having the Config on a different level. But since then , that issue has been taken care of (by createClientWithConfig)

So you're right, now it's not expected. I would still keep it at DEBUG level tho, thinking about the uproar I caused last time (e.g. in `vault` which depends on `gosnowflake`) with excessive logging around the dbus daemon  :)

s3_storage_client.go
sfc-gh-dszmolka: yes, should have done that 🤦

util_test.go
sfc-gh-dszmolka: can you please expand on how you meant to do inlining this struct?

all other concerns have been addressed.

sfc-gh-dszmolka: as discussed ; has now been inlined

connection_test.go
```diff
@@ -826,3 +826,56 @@ func TestBeginCreatesTransaction(t *testing.T) {
 		}
 	})
 }
+
+type EmptyTransporter struct{}
+
+func (t EmptyTransporter) RoundTrip(req *http.Request) (*http.Response, error) {
+	return snowflakeInsecureTransport.RoundTrip(req)
```
sfc-gh-pfus, line 833: So it is a bit misleading. If it is an empty transporter, I suggest returning just `nil, nil`.

sfc-gh-pfus (approved):

connection_test.go
sfc-gh-dszmolka: that's fair ;) changed it

sfc-gh-dszmolka:
in addition to the automated tests; manual tests were performed issuing
* `PUT` / `GET` (with 1mb CSV)
* `LIST`
* `RM` 

commands on stages, on accounts placed in

* AWS US East 1
* Azure West Europe
* GCP US Central 1

with and without `disableOCSPChecks` / `insecureMode` ,

connection.go
```diff
@@ -856,3 +856,21 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
+		return SnowflakeTransport
+	}
+	// if user configured a custom Transporter, prioritize that
+	if cfg.Transporter != nil {
+		logger.Debug("getTransport: using Transporter configured by the user")
+		return cfg.Transporter
+	}
+	if cfg.DisableOCSPChecks || cfg.InsecureMode {
+		logger.Debug("getTransport: skipping OCSP validation for cloud storage")
+		return snowflakeInsecureTransport
```
sfc-gh-pbulawa, line 872: I think that we should name it differently, e.g. `snowflakeOcspDisabledTransport` to avoid any association that OCSP disabled = insecure

sfc-gh-dszmolka: currently we use OCSP for certificate validation. if OCSP is disabled, we cannot guarantee there isn't a MITM who presents a fake certificate.

plot twist of course on summer 2024 where CAs started to move away from OCSP and started to really consider OCSP Responder endpoint in their certs as optional. E.g. we now have Google certs who absolutely don't have OCSP Responder endpoints for certain CA certs (_not_ all certs in the chain!) so for these particular certs we cannot perform _full_ OCSP validation anyways - but we _can_ perform OCSP validation on the other certs in the chain who _do_ have OCSP Responder endpoint populated.

if one decides to deliberately and manually  disable OCSP verification for all certs in the chain, that doesn't make the whole process *more* secure :)

sfc-gh-dszmolka: but renamed , thank you for the suggestion

sfc-gh-dheyman (approved):

PR #1293, opened 2025-01-28
deps: upgrade Arrow to v18

zeroshade:
### Description

Upgrading Apache Arrow dependency to v18.1.0, with the shift to it's own repository out of the arrow monorepo, there will be fewer major version changes allowing downstream consumers to be able to freely upgrade the Arrow dependency in the future after this update.

There are also significant benefits, including performance and bug fixes, to shifting from v16 -> v18. 

It will also result in shrinking the size of the Snowflake ADBC driver as it won't need to include both v16 and v18 in the resulting library.

### Checklist


sfc-gh-pfus:
Hello again @zeroshade ! Thanks for your suggestion. However, we still have to support Go 1.21 until Go 1.24 is released due to our internal policy. Let's put this PR on hold for the next couple of months.

zeroshade:
@sfc-gh-pfus Looks like [Go 1.24](https://tip.golang.org/doc/go1.24) has been released! can we take another look at this?

sfc-gh-pfus:
Sure, I'll make sure it is merged before the next release!

sfc-gh-pfus:
Hi @zeroshade ! I checked, that you bumped Arrow to v18.1, which requires Go 1.22.7. We have to cover 1.22, so the best we can do now is to upgrade to v18.0. Is it ok for you? Do you want to do this in your PR or you want me to make it?

zeroshade:
@sfc-gh-pfus I've updated this to be v18.0.0 that's fine on my end as it would allow downstream consumers to still use v18.1.0 if they want without requiring multiple versions of different major Arrow versions.

go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/snowflakedb/gosnowflake
 
-go 1.22
+go 1.22.7
```
sfc-gh-pfus, line 3: Can we use 1.22 here?

zeroshade: Probably yea, `go mod tidy` was the one that added the `.7`, i'll try changing this back to just `go 1.22`

zeroshade:
@sfc-gh-pfus how's this?

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

sfc-gh-pfus:
@zeroshade merged, thanks for your input once again!

PR #1297, opened 2025-01-31
SNOW-1313648 GO - Verify value bindings for all field types while exceeding CLIENT_STAGE_ARRAY_BINDING_THRESHOLD

sfc-gh-ext-simba-jy:
### Description

SNOW-313648 Please explain the changes you made here.
- [x] Fixed the time difference when the timestamp value is before 1600
- [x] Add testing to validate that inserting data with the stage binding, array binding, and interface array binding are the same.

### Checklist


bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
 	})
 }
 
+func TestSNOW1313648(t *testing.T) {
```
sfc-gh-pfus, line 877: Can we have a better name?

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+	arrayInsertTable := "Snow1313648Insert"
```
sfc-gh-pfus, line 878: Here the same, this is just a regular feature that we can name instead of using Jira ID.

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+	stageBindingTable := "Snow1313648stageBinding"
+	interfaceArrayTable := "Snow1313648stageInterface"
+
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", arrayInsertTable))
```
sfc-gh-pfus, line 883: Also boolean and double?

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", stageBindingTable))
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", interfaceArrayTable))
+
+		someTime := time.Date(1, time.January, 1, 12, 34, 56, 123456789, time.UTC)
+		someDate := time.Date(2024, time.March, 18, 0, 0, 0, 0, time.UTC)
+		testingDate := time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)
+		someBinary := []byte{0x01, 0x02, 0x03}
```
sfc-gh-pfus, line 890: You can use just `'a', 'b'`, but your version is also fine :)

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+			strArr[i] = "test" + strconv.Itoa(i)
+			ltzArr[i] = testingDate
+			tzArr[i] = testingDate.Add(time.Hour).UTC()
+			ntzArr[i] = testingDate.Add(2 * time.Hour)
+			dateArr[i] = someDate
+			timeArr[i] = someTime
+			binArr[i] = someBinary
+		}
+		dbt.mustExec(fmt.Sprintf("insert into %v values (?, ?, ?, ?, ?, ?, ?, ?)", arrayInsertTable), Array(&intArr), Array(&strArr), Array(&ltzArr, TimestampLTZType), Array(&tzArr, TimestampTZType), Array(&ntzArr, TimestampNTZType), Array(&dateArr, DateType), Array(&timeArr, TimeType), Array(&binArr))
+		dbt.mustExec("ALTER SESSION SET CLIENT_STAGE_ARRAY_BINDING_THRESHOLD = 1")
```
sfc-gh-pfus, line 934: Does it work? I thought this was an internal parameter.

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
 
+func TestSNOW1313648(t *testing.T) {
+	arrayInsertTable := "Snow1313648Insert"
+	stageBindingTable := "Snow1313648stageBinding"
+	interfaceArrayTable := "Snow1313648stageInterface"
+
+	runDBTest(t, func(dbt *DBTest) {
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", arrayInsertTable))
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", stageBindingTable))
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", interfaceArrayTable))
```
sfc-gh-pfus, line 885: Please, check if there is any similar existing test and if yes - decide if duplication is needed (merge or update/remove).

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+
+		someTime := time.Date(1, time.January, 1, 12, 34, 56, 123456789, time.UTC)
+		someDate := time.Date(2024, time.March, 18, 0, 0, 0, 0, time.UTC)
+		testingDate := time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)
+		someBinary := []byte{0x01, 0x02, 0x03}
+
+		numRows := 5
+		intArrInterface := make([]any, numRows)
```
sfc-gh-pfus, line 893: Why do we need such a case?

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+			tzArrInterface[i] = testingDate.Add(time.Hour).UTC()
+			ntzArrInterface[i] = testingDate.Add(2 * time.Hour)
+			dateArrInterface[i] = someDate
+			timeArrInterface[i] = someTime
+			binArrInterface[i] = someBinary
+		}
+
+		dbt.mustExec(fmt.Sprintf("insert into %v values (?, ?, ?, ?, ?, ?, ?, ?)", arrayInsertTable), Array(&intArrInterface), Array(&strArrInterface), Array(&ltzArrInterface, TimestampLTZType), Array(&tzArrInterface, TimestampTZType), Array(&ntzArrInterface, TimestampNTZType), Array(&dateArrInterface, DateType), Array(&timeArrInterface, TimeType), Array(&binArrInterface))
+
+		numRows = 5
```
sfc-gh-pfus, line 914: Needed?

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+			assertBytesEqualE(t, b, ib)
+
+			assertEqualE(t, "test"+strconv.Itoa(cnt), s)
+			assertEqualE(t, ltz.UTC(), testingDate.UTC())
+			assertEqualE(t, tz.UTC(), testingDate.Add(time.Hour).UTC())
+			assertEqualE(t, ntz.UTC(), testingDate.Add(2*time.Hour).UTC())
+			assertEqualE(t, date, someDate)
+			assertEqualE(t, tt, someTime)
+			assertBytesEqualE(t, b, someBinary)
+			cnt++
```
sfc-gh-pfus, line 998: Do we need this? You wanted to perform assertion on rows number?

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+
+			if err := bindingRows.Scan(&bi, &bs, &bltz, &btz, &bntz, &bDate, &btt, &bb); err != nil {
+				t.Fatal(err)
+			}
+
+			if err := interfaceRows.Scan(&ii, &is, &iltz, &itz, &intz, &iDate, &itt, &ib); err != nil {
+				t.Fatal(err)
+			}
+
+			assertEqualE(t, i, bi)
```
sfc-gh-pfus, line 967: If all values are empty/zeros this comparison succeeds. Maybe it would be better to compare with given values?

bindings_test.go
sfc-gh-pfus, line 883: It is a good practice (I know that missing in many tests) to defer dropping the table.

bindings_test.go
```diff
@@ -874,6 +874,132 @@ func TestBulkArrayBinding(t *testing.T) {
...
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", arrayInsertTable))
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", stageBindingTable))
+		dbt.mustExec(fmt.Sprintf("create or replace table %v (c1 integer, c2 string, c3 timestamp_ltz, c4 timestamp_tz, c5 timestamp_ntz, c6 date, c7 time, c8 binary)", interfaceArrayTable))
+
+		someTime := time.Date(1, time.January, 1, 12, 34, 56, 123456789, time.UTC)
+		someDate := time.Date(2024, time.March, 18, 0, 0, 0, 0, time.UTC)
+		testingDate := time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC)
+		someBinary := []byte{0x01, 0x02, 0x03}
+
+		numRows := 5
```
sfc-gh-pfus, line 892: I'm wondering if such a small number is good enough. There may be a lot of problems with greater number (like 100k) that do not exist with 5 rows. Let's try to increase it and decide if it is not too slow.

converter.go
```diff
@@ -3219,6 +3200,25 @@ func convertTzTypeToSnowflakeType(tzType timezoneType) snowflakeType {
...
+func getTimestampBindValue(x time.Time, stream bool, t snowflakeType) string {
+	if stream {
+		return x.Format(format)
+	}
+	return convertTimeToTimeStamp(x, t)
+}
+
+func convertTimeToTimeStamp(x time.Time, t snowflakeType) string {
+	unixTime, _ := new(big.Int).SetString(fmt.Sprintf("%d", x.Unix()), 10)
+	m, _ := new(big.Int).SetString(strconv.FormatInt(1e9, 10), 10)
```
sfc-gh-pfus, line 3212: Maybe we should propagate the errors here?

converter.go
```diff
@@ -3219,6 +3200,25 @@ func convertTzTypeToSnowflakeType(tzType timezoneType) snowflakeType {
...
+	unixTime.Mul(unixTime, m)
+	tmNanos, _ := new(big.Int).SetString(fmt.Sprintf("%d", x.Nanosecond()), 10)
+	if t == timestampTzType {
+		_, offset := x.Zone()
+		return fmt.Sprintf("%v %v", unixTime.Add(unixTime, tmNanos), offset/60+1440)
```
sfc-gh-pfus, line 3217: Previously it operated on original time (`x` in this context), is it intentional to use `unixTime` here? Your change seems rational to me, but wanted to double check.

bindings_test.go
sfc-gh-ext-simba-jy: I changed it.

bindings_test.go
sfc-gh-ext-simba-jy: I changed them.

bindings_test.go
sfc-gh-ext-simba-jy: Added.

bindings_test.go
sfc-gh-ext-simba-jy: I removed the binary because the same testing exists.

bindings_test.go
sfc-gh-ext-simba-jy: It only works on GitHub testing, but not on the local. Would it be better to make this testing runnable on the local?

bindings_test.go
sfc-gh-ext-simba-jy: As far as I know, we need to validate that all inserted values are the same. Unlike other drivers, Go has a unique binding with the interface type, so I think we need to validate this as well. WDYT? https://github.com/snowflakedb/gosnowflake/blob/master/converter.go#L2681

bindings_test.go
sfc-gh-ext-simba-jy: Nope!

bindings_test.go
sfc-gh-ext-simba-jy: I removed it.

bindings_test.go
sfc-gh-ext-simba-jy: Actually, I need this value as an ID. Since the other values are generated randomly, I need to organize them in order to check if any values have been altered.

bindings_test.go
sfc-gh-ext-simba-jy: Done.

bindings_test.go
sfc-gh-ext-simba-jy: I increased this value to 1000, but I'm not sure if it's necessary, given that we can execute stage binding using the ALTER SESSION command. If num is over 100k, the driver will execute the stage binding only.

converter.go
sfc-gh-ext-simba-jy: Yes, this is intentional due to Go's time handling. If we use the previous method, times older than 1700 years (or at some point) cause shifts in the tz column. To avoid this, the value should be converted to a big integer, like NTZ or LTZ, and the offset should be applied accordingly.

converter.go
sfc-gh-ext-simba-jy: MO, error handling is not necessary here because the FormatInt function cannot fail—it operates on a constant (essentially hardcoded). This means the m value is always 1,000,000,000, which is why we do not handle the error in this case.
https://github.com/snowflakedb/gosnowflake/pull/1297/files#diff-ecad48054a05ca5aecf42b278530331772747ed7211a193aacd94214cb53aad0L906

bindings_test.go
sfc-gh-pfus: It is good enough then. I didn't know we had this permission on GH.

bindings_test.go
```diff
@@ -874,6 +874,130 @@ func TestBulkArrayBinding(t *testing.T) {
...
+			assertNilF(t, interfaceRows.Close())
+		}()
+		var i, bi, ii int
+		var s, bs, is string
+		var ltz, bltz, iltz, itz, btz, tz, intz, ntz, bntz, iDate, date, bDate, itt, tt, btt time.Time
+		var b, bb, ib bool
+		var d, bd, id float64
+
+		for k := 0; k < numRows; k++ {
+			insertRows.Next()
```
sfc-gh-pfus, line 948: We should verify there are rows using `assertTrueF(t, insertRows.Next())`.

bindings_test.go
```diff
@@ -874,6 +874,130 @@ func TestBulkArrayBinding(t *testing.T) {
...
+			if err := insertRows.Scan(&i, &s, &ltz, &tz, &ntz, &date, &tt, &b, &d); err != nil {
+				t.Fatal(err)
```
sfc-gh-pfus, line 950: `assertNilF(t, err)`?

bindings_test.go
sfc-gh-pfus: Ok, so my comment should have been put on the line below :) What I mean is - let's suppose that from any reason (newly introduced bug?) binding strings doesn't work and we always insert empty string. In that case, if you select data from two tables and compare them, it works, because they were both incorrectly bound. It would be better to compare not two tables to each other, but each table to `"test" + strconv.Itoa(i)`, wouldn't it? Same goes for all types. Floats do not have to be random, they can be also related to index.

converter.go
sfc-gh-pfus: That is true for today. But you never know when someone changes the other behavior that may influence this one, right? It is safer to always handle errors, even if today's code never fails.

bindings_test.go
sfc-gh-pfus: Ok, fair enough!

bindings_test.go
sfc-gh-pfus: Ok, so in that case we can stay with 3 or 5, there is no difference between 3 and 1000 IMHO.

bindings_test.go
sfc-gh-ext-simba-jy: I replaced it

converter.go
sfc-gh-ext-simba-jy: Make sense. I added error handling.

bindings_test.go
sfc-gh-ext-simba-jy: From line 984, the test validates whether the inserted binding values match the values in the randomly generated array. As a result, it ensures that different bindings hold the same values and that those values match the array. For example, according to lines 960, 961, and 984, s == bs == is == "test" + strconv.Itoa(i), confirming that all three values are identical. WDYT?

converter.go
sfc-gh-pfus: Yes, I know, this is a problem with Go coverage :) In other languages this is just exception, no extra lines for propagating error up.

bindings_test.go
sfc-gh-pfus: That is fine, but less readable :) Why can't we have something like:

```
assertEqualE(t, "test"+strconv.Itoa(k), s)
assertEqualE(t, "test"+strconv.Itoa(k), is)
assertEqualE(t, "test"+strconv.Itoa(k), bs)
```

sfc-gh-ext-simba-jy: Sure! No problem then. I just want to double-check it.

sfc-gh-pfus (approved):

bindings_test.go
```diff
@@ -1423,3 +1547,11 @@ func testInsertLOBData(t *testing.T, useArrowFormat bool, isLiteral bool) {
 		dbt.mustExec(unsetFeatureMaxLOBSize)
 	})
 }
+
+func getRandomDate() time.Time {
+	return time.Date(rand.Intn(1700)+1, time.January, rand.Intn(40), rand.Intn(40), rand.Intn(40), rand.Intn(40), rand.Intn(40), time.UTC)
```
sfc-gh-dprzybysz, line 1552: why date is > 1700? we should test also values below 1582 (julian to gregorian calendar switch)

sfc-gh-ext-simba-jy: When I reproduced the error, the time shift occurred in the year before 1700/1/1 as far as I remember. We can use 1582. I  replaced the value. Thank you!

PR #1300, opened 2025-02-05
SNOW-921006 Introduce wiremock

sfc-gh-pfus:
### Description

SNOW-921006 Introduce wiremock

### Checklist

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -67,6 +67,7 @@ jobs:
                 CLOUD_PROVIDER: ${{ matrix.cloud }}
                 GORACE: history_size=7
                 GO_TEST_PARAMS: ${{ inputs.goTestParams }}
+                WIREMOCK_PORT: 14335
               run: ./ci/test.sh
             - name: Upload coverage to Codecov
               uses: codecov/codecov-action@v5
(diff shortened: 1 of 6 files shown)


.github/workflows/build-test.yml
```diff
@@ -26,8 +26,6 @@ concurrency:
 jobs:
     lint:
         runs-on: ubuntu-latest
-        strategy:
-          fail-fast: false
```
sfc-gh-mkubik, line 30: why do we remove this line?

sfc-gh-mkubik (approved):

wiremock_test.go
sfc-gh-dheyman, line 1: The name of this file suggests that we are testing wiremock, but I assume that this is a base utility class for tests that use wiremock, right?

wiremock_test.go
```diff
@@ -0,0 +1,86 @@
...
+type wiremockClient struct {
+	host   string
+	port   int
+	client http.Client
+}
+
+func newWiremock() *wiremockClient {
+	return &wiremockClient{
+		host: "127.0.0.1",
+		port: 14355,
```
sfc-gh-dheyman, line 23: perhaps it would be better to pass this as an ENV VAR in test scripsts?

sfc-gh-dheyman (approved):

wiremock_test.go
sfc-gh-pfus: Yes. I have to add this suffix, because this is the only way how to exclude files from going into production.

wiremock_test.go
sfc-gh-pfus: Good idea, done.

PR #1301, opened 2025-02-07
SNOW-1825790: Secure OAuth token cache on linux

sfc-gh-jszczerbinski:
### Description

### Checklist

### DIFF
--- a/secure_storage_manager.go
+++ b/secure_storage_manager.go
@@ -109,18 +109,68 @@ func (ssm *fileBasedSecureStorageManager) createCacheDir(credCacheDir string) er
 	return err
 }
 
+func lookupCacheDir(envVar string, pathSegments ...string) (string, error) {
+	envVal := os.Getenv(envVar)
+	if envVal == "" {
+		return "", fmt.Errorf("environment variable %s not set", envVar)
+	}
+
+	fileInfo, err := os.Stat(envVal)
+	if err != nil {
+		return "", fmt.Errorf("failed to stat %s=%s, due to %w", envVar, envVal, err)
+	}
+
+	if !fileInfo.IsDir() {
+		return "", fmt.Errorf("environment variable %s=%s is not a directory", envVar, envVal)
+	}
+
+	cacheDir := envVal
+
+	if len(pathSegments) > 0 {
+		for _, pathSegment := range pathSegments {
+			err := os.Mkdir(pathSegment, os.ModePerm)
+			if err != nil {
+				return "", fmt.Errorf("failed to create cache directory. %v, err: %w", pathSegment, err)
+			}
+			cacheDir = filepath.Join(cacheDir, pathSegment)
+		}
+		fileInfo, err = os.Stat(cacheDir)
+		if err != nil {
+			return "", fmt.Errorf("failed to stat %s=%s, due to %w", envVar, cacheDir, err)
+		}
+	}
+
+	if fileInfo.Mode()&os.ModePerm != 0o700 {
+		err := os.Chmod(cacheDir, 0o700)
+		if err != nil {
+			return "", fmt.Errorf("failed to chmod cache directory. %v, err: %w", cacheDir, err)
+		}
+	}
+
+	return cacheDir, nil
+}
+
 func (ssm *fileBasedSecureStorageManager) buildCredCacheDirPath() string {
-	credCacheDir := os.Getenv(credCacheDirEnv)
-	if credCacheDir != "" {
-		return credCacheDir
+	type cacheDirConf struct {
+		envVar       string
+		pathSegments []string
 	}
-	home := os.Getenv("HOME")
-	if home == "" {
-		logger.Info("HOME is blank")
-		return ""
+	confs := []cacheDirConf{
+		{envVar: credCacheDirEnv, pathSegments: []string{}},
+		{envVar: "XDG_CACHE_DIR", pathSegments: []string{"snowflake"}},
+		{envVar: "HOME", pathSegments: []string{".cache", "snowflake"}},
 	}
-	credCacheDir = filepath.Join(home, ".cache", "snowflake")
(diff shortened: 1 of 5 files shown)


secure_storage_manager.go
```diff
@@ -23,11 +24,21 @@ const (
...
+	credCacheFileName = "credential_cache_v1.json"
 )
 
+type cacheDirConf struct {
+	envVar       string
+	pathSegments []string
+}
+
+var defaultLinuxCacheDirConf = []cacheDirConf{
+	{envVar: credCacheDirEnv, pathSegments: []string{}},
```
sfc-gh-pfus, line 37: Design doc also adds TEMPORARY_CREDENTIAL_CACHE_DIR parameter, should we add it as well?

secure_storage_manager.go
```diff
@@ -109,141 +115,259 @@ func (ssm *fileBasedSecureStorageManager) createCacheDir(credCacheDir string) er
...
-	ssm.credCacheLock.Lock()
-	defer ssm.credCacheLock.Unlock()
-	localCredCache := ssm.readTemporaryCacheFile()
-	cred := localCredCache[credentialsKey]
-	if cred != "" {
-		logger.Debug("Successfully read token. Returning as string")
-	} else {
-		logger.Debug("Returned credential is empty")
+	err := ssm.lockFile()
+	if err != nil {
```
sfc-gh-pfus, line 246: Is there a chance, that err means something else than creating a dir? If yes, we can create a dir (meaning locks for other processes), but we never unlocks the file.

sfc-gh-pfus (approved):

secure_storage_manager.go
sfc-gh-jszczerbinski: This is optional, I don't think we should unless you see a benefit of adding it?

secure_storage_manager.go
sfc-gh-jszczerbinski: I don't think so. The contract here is that lockFile is successful iff we created a lock/directory. If it returns an error it means we don't have to unlock it because we did not created a lock/directory.

secure_storage_manager.go
sfc-gh-pfus: I don't. Changing via env var is good enough.

secure_storage_manager.go
```diff
@@ -73,177 +84,279 @@ func newSecureStorageManager() secureStorageManager {
...
+		return "", fmt.Errorf("failed to stat %s=%s, due to %v", envVar, envVal, err)
+	}
+
+	if !fileInfo.IsDir() {
+		return "", fmt.Errorf("environment variable %s=%s is not a directory", envVar, envVal)
+	}
+
+	cacheDir := filepath.Join(envVal, filepath.Join(pathSegments...))
+
+	if err = os.MkdirAll(cacheDir, os.FileMode(0o755)); err != nil {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Because it is `MkdirAll` - we create all path segments that didn't exist before, but only the last one has to be 755.

secure_storage_manager.go
```diff
@@ -23,11 +24,21 @@ const (
...
+
+var defaultLinuxCacheDirConf = []cacheDirConf{
+	{envVar: credCacheDirEnv, pathSegments: []string{}},
+	{envVar: "XDG_CACHE_DIR", pathSegments: []string{"snowflake"}},
+	{envVar: "HOME", pathSegments: []string{".cache", "snowflake"}},
+}
+
 type secureTokenSpec struct {
 	host, user string
 	tokenType  tokenType
```
sfc-gh-jszczerbinski, line 44: Do we want to add OAUTH_ACCESS_TOKEN and OAUTH_REFRESH_TOKEN token types already?

secure_storage_manager_test.go
```diff
@@ -3,9 +3,98 @@
...
+	skipOnWindows(t, "file system permission is different")
+	credCacheDir, err := os.MkdirTemp("", "")
+	assertNilF(t, err)
+	assertNilF(t, os.MkdirAll(credCacheDir, 0777))
+	credCacheDirEnvOverride := overrideEnv(credCacheDirEnv, credCacheDir)
+	defer credCacheDirEnvOverride.rollback()
+	ssm, err := newFileBasedSecureStorageManager()
+	assertNilF(t, err)
+
+	t.Run("success", func(t *testing.T) {
```
sfc-gh-jszczerbinski, line 69: Can we add similar test where we operate on two tokens at once? For example it would make sure that we don't remove/override existing tokens when adding a new one.

secure_storage_manager.go
```diff
@@ -73,177 +84,279 @@ func newSecureStorageManager() secureStorageManager {
...
-		}
-		if err = os.Mkdir(credCacheLockFileName, 0600); err != nil {
-			logger.Debugf("failed to delete lock file. file: %v, err: %v. ignored.\n", credCacheLockFileName, err)
-			return
+	stat, err := os.Stat(ssm.credFilePath())
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	if err == nil {
+		if stat.Mode().String() != "-rw-------" {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Done.

secure_storage_manager.go
sfc-gh-pfus: No, we'll add them when needed.

secure_storage_manager_test.go
sfc-gh-pfus: Definitely.

sfc-gh-pfus (approved):

sfc-gh-pmotacki (approved):

sfc-gh-pfus:
Work continued in https://github.com/snowflakedb/gosnowflake/pull/1327

PR #1305, opened 2025-02-11
SNOW-1825500 Handle OAuth2 parameters in DSN

sfc-gh-pfus:
### Description

SNOW-1825500 Added config option for OAuth2

### Checklist

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -54,6 +54,13 @@ type Config struct {
 	Role      string // Role
 	Region    string // Region
 
+	OauthClientID         string // Client id for OAuth2 external IdP
+	OauthClientSecret     string // Client secret for OAuth2 external IdP
+	OauthAuthorizationURL string // Authorization URL of Auth2 external IdP
+	OauthTokenRequestURL  string // Token request URL of Auth2 external IdP
+	OauthRedirectURI      string // Redirect URI registered in IdP. The default is http://127.0.0.1:<random port>/
+	OauthScopes           string // Comma separated list of scopes. If empty it is derived from role.
+
 	// ValidateDefaultParameters disable the validation checks for Database, Schema, Warehouse and Role
 	// at the time a connection is established
 	ValidateDefaultParameters ConfigBool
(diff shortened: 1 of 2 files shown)


dsn.go
```diff
@@ -54,6 +54,13 @@ type Config struct {
 	Role      string // Role
 	Region    string // Region
 
+	OauthClientID         string // Client id for OAuth2 external IdP
```
sfc-gh-mhofman, line 57: EnableExperimentalAuthentication flag will be a separate part of:
https://snowflakecomputing.atlassian.net/browse/SNOW-1924902?

dsn.go
```diff
@@ -54,6 +54,13 @@ type Config struct {
...
+	OauthClientSecret     string // Client secret for OAuth2 external IdP
+	OauthAuthorizationURL string // Authorization URL of Auth2 external IdP
+	OauthTokenRequestURL  string // Token request URL of Auth2 external IdP
+	OauthRedirectURI      string // Redirect URI registered in IdP. The default is http://127.0.0.1:<random port>/
+	OauthScopes           string // Comma separated list of scopes. If empty it is derived from role.
```
sfc-gh-mhofman, line 62: We named that in ODBC and JDBC without plural. Typically it will be just one scope. I'm wondering whether should we change python, jdbc and odbc impl or just remove ~s postfix here.
cc @sfc-gh-dheyman

dsn.go
```diff
@@ -54,6 +54,13 @@ type Config struct {
 	Role      string // Role
 	Region    string // Region
 
+	OauthClientID         string // Client id for OAuth2 external IdP
+	OauthClientSecret     string // Client secret for OAuth2 external IdP
+	OauthAuthorizationURL string // Authorization URL of Auth2 external IdP
```
sfc-gh-mhofman, line 59: @sfc-gh-pfus can you put those values into the Drivers OAuth/PAT parameters spreadsheet of @sfc-gh-dheyman so we gather naming in one place?

sfc-gh-mhofman, line 59: authorization URL should have a default value defined in line 21 for authenticator = “oauth_client_credentials” or “oauth_authorization_code and set I guess somewhere in fillMissingConfigParameters, right?

sfc-gh-dheyman: I am handling this when actually requesting for authz code, so I understand if @sfc-gh-pfus would like to address this in later PRs

dsn.go
```diff
@@ -54,6 +54,13 @@ type Config struct {
...
+	OauthTokenRequestURL  string // Token request URL of Auth2 external IdP
```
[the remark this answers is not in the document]
sfc-gh-dheyman: same as in case of authz request url

dsn.go
sfc-gh-dheyman: I think we should remove 's' from this param name and if someone want's to specify multiple scopes than it should be separated by whitespace. Just like we're sending it. Also probably good idea to point this out in the docs.

dsn.go
sfc-gh-pfus: Yes, it will be added when it is needed. I can't block properties based on that :)

dsn.go
sfc-gh-pfus: Sure, will do when the PR is merged.

sfc-gh-pfus: Yes, it will be the oauth client responsibility to build default URLs.

dsn.go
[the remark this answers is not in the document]
sfc-gh-pfus: same as above

dsn.go
sfc-gh-pfus: Ok, good catches!

sfc-gh-mhofman (approved):

sfc-gh-dheyman (approved):

PR #1309, opened 2025-02-18
NO-SNOW log JWT claims on debug loglevel

sfc-gh-dszmolka:
### Description

Being able to see what claims are used inside JWT is immensely useful for troubleshooting issues with such authentication type, and can even one-shot certain types of issues (wrong key used, wrong login_name used, etc.). 

It would be equally useful having in gosnowflake too. Even more useful, since it would also help Snowflake Terraform Provider issues which trace back to keypair authentication.

No auth logic was changed, all the change is doing is to add two additional loglines, one of which logs out the JWT claims for the token generated.

### Checklist

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -481,6 +481,7 @@ func prepareJWTToken(config *Config) (string, error) {
 	if config.PrivateKey == nil {
 		return "", errors.New("trying to use keypair authentication, but PrivateKey was not provided in the driver config")
 	}
+	logger.Debug("preparing JWT for keypair authentication")
 	pubBytes, err := x509.MarshalPKIXPublicKey(config.PrivateKey.Public())
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


auth.go
```diff
@@ -505,6 +506,7 @@ func prepareJWTToken(config *Config) (string, error) {
 		return "", err
 	}
 
+	logger.Debugf("successfully generated JWT. token: %v", tokenString)
```
sfc-gh-mhofman, line 509: don't

sfc-gh-mhofman, line 509: It's better just to just say that it was generated without logging the value.

sfc-gh-mhofman (changes requested):

auth.go
sfc-gh-dszmolka: modified to log not-so-sensitive information, which after reviewing, don't even need for troubleshooting, only the header w/claims

sfc-gh-pfus (approved):

sfc-gh-mhofman (approved):

PR #1324, opened 2025-03-05
SNOW-921006 Add HTTPS wiremock with custom CA

sfc-gh-pfus:
### Description

SNOW-921006 Added a test scenario for HTTPS on wiremock.

### Checklist


.github/workflows/build-test.yml
```diff
@@ -66,6 +66,7 @@ jobs:
                 GORACE: history_size=7
                 GO_TEST_PARAMS: ${{ inputs.goTestParams }}
                 WIREMOCK_PORT: 14335
+                WIREMOCK_HTTPS_PORT: 13567
```
sfc-gh-pmotacki, line 69: Do we must use static port?

sfc-gh-pmotacki (approved):

wiremock_test.go
```diff
@@ -1,22 +1,28 @@
...
+var wiremockHTTPS = newWiremockHTTPS()
 
 type wiremockClient struct {
-	protocol string
-	host     string
-	port     int
-	client   http.Client
+	protocol  string
+	host      string
+	port      int
```
sfc-gh-knozderko, line 23: isn't it that port is https port, and adminPort is http port?
What do you think of using httpsPort and httpPort instead of port and adminPort?

sfc-gh-pfus: I don't want it. With my solution `port` is a target port, regardless if this is HTTP or HTTPS. No one has to consider what port should we use.

ci/scripts/README.md
```diff
@@ -0,0 +1,7 @@
+# Refreshing wiremock test cert
+
+Password for CA is `password`.
+
+```bash
+openssl x509 -req -in wiremock.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out wiremock.crt -days 365 -sha256 -extfile wiremock.v3.ext
```
sfc-gh-knozderko, line 6: So once every 365 days this certificate will need to be replaced?

I think it would be good to describe how all this certificate related files were created.

sfc-gh-knozderko (approved):

PR #1327, opened 2025-03-06
SNOW-1825790 Implement safer file based token cache

sfc-gh-pfus:
### Description

SNOW-1825790 Implement safer file based token cache

### Checklist
- [x] Created tests which fail without the change (if possible)
- [x] Extended the README / documentation, if necessary

PR #1332, opened 2025-03-12
MINOR: Change stray Info logs to Debug

lidavidm:
### Description

We noticed in the ADBC Snowflake driver that there was some stray logging output that appeared to come from gosnowflake. I think these should be changed to Debug level as they're not useful to the user and may confuse them instead. Also, changing the logger level doesn't help with the "Credentials cache path" message since that gets triggered by a global, before any application code can run.

### Checklist


sfc-gh-pfus:
Hi, there are a lot of changes in token caching going on: https://github.com/snowflakedb/gosnowflake/pull/1327
I checked that there are no info logs left. I'm going to close your PR, because it won't be applicable for the next release. Thank you for your valid suggestion, though!

PR #1360, opened 2025-04-09
SNOW-2032053: Added logic to propagate trace context, if it exists, to GS

sfc-gh-mkarjikar:
### Description

SNOW-2032053 

Added support for trace propagation by using OpenTelemetry's trace context injector for Go. This change adds one additional dependency for the Go driver, which is OpenTelemetry v1.35.0. 

The only functional change is an extra couple headers are injected into the HTTPS request to GS, and GS is already equipped to extract it on the other side if it is present - i.e. the header "traceparent", where the value is in W3 Trace Context Format.

This change will allow trace context to be propagated from anywhere the Go driver is being used to GS. The primary use case is for a Go service running on SPCS. The user will then be able to see the entire line of execution in Snowsight in one view, from their SPCS Service to the SQL queries they ran using the connector. 

This is a feature request from Relational AI, an external customer.

The compilation time and size as well as the runtime were benchmarked before and after the change using the `cm/select1` program. These were tested (built and ran) 5 times each.

Compilation/Runtime for cmd/select1 before/after these changes:
- Before changes (times are averaged):
  - Compilation Time: `2.21 seconds`
  - Runtime: `0.8 seconds`
  - Compilation Size: `33.35MB`
 
- With changes (times are averaged):
  - Compilation Time: `2.25 seconds`
  - Runtime: `0.82 seconds`
  - Compilation Size: `33.43MB`

### Checklist


sfc-gh-mkarjikar:

connection.go
```diff
@@ -128,6 +130,9 @@ func (sc *snowflakeConn) exec(
 	if isFileTransfer(query) {
 		headers[httpHeaderAccept] = headerContentTypeApplicationJSON
 	}
+	// propagate traceID and spanID via traceparent header. this is a no-op if invalid IDs
+	propagator := propagation.TraceContext{}
```
sfc-gh-bnandibhatla, line 134: Can we add a unit test for this and verify that the trace is being propagated?

connection_test.go
```diff
@@ -271,6 +274,56 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+
+	postQueryMock := func(_ context.Context, _ *snowflakeRestful,
+		_ *url.Values, headers map[string]string, _ []byte, _ time.Duration,
+		_ UUID, _ *Config) (*execResponse, error) {
+
+		// ensure the traceID and spanID from the ctx passed in has been injected into the headers
+		// in W3 Trace Context format
+		if headers["traceparent"] != expectedTraceparent {
+			t.Fatalf("traceparent doesn't match. expected: %v, got: %v", expectedTraceparent, headers["traceparent"])
+		}
```
sfc-gh-bnandibhatla, line 300:
```suggestion
		require.Equal(t, expectedTraceparent, headers["traceparent"])
```

connection_test.go
```diff
@@ -271,6 +274,56 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+
+	sc := &snowflakeConn{
+		cfg:               &Config{Params: map[string]*string{}},
+		rest:              sr,
+		queryContextCache: (&queryContextCache{}).init(),
+	}
+	if _, err := sc.exec(ctx, "", false /* noResult */, false, /* isInternal */
+		false /* describeOnly */, nil); err != nil {
+		t.Fatalf("err: %v", err)
+	}
```
sfc-gh-bnandibhatla, line 323:
```suggestion
	_, err := sc.exec(ctx, "", /* noResult */ false,  /* isInternal */ false,
		/* describeOnly */ false, /* bindings */ nil)
	require.NoError(t, err)	
```

connection_test.go
```diff
@@ -271,6 +274,56 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+
```
sfc-gh-bnandibhatla, line 324: remove unnecessary newline.

connection_test.go
```diff
@@ -271,6 +274,56 @@ func TestExecWithSpecificRequestID(t *testing.T) {
 	}
 }
 
+func TestExecContextPropagationIntegrationTest(t *testing.T) {
+	tp := trace.NewTracerProvider()
+	otel.SetTracerProvider(tp)
```
sfc-gh-pfus, line 279: Is this some static/global setting? Should we revert the change at the end (`t.Cleanup`)?

connection_test.go
sfc-gh-pfus: We don't use `require`, but we have custom assertions - can you use `assertEqualE`?

connection_test.go
sfc-gh-pfus: `assertNilF`?

connection.go
sfc-gh-pfus, line 134: This change is applied for executing queries only (issues via Query/Exec/transactions). It is not applied to:
1. Login/logout requests.
2. Fetching results for long running queries (after 45s the query continues, but GS responds with "I'm continuing query, check for results in `/monitoring/queries`, take a look at https://github.com/snowflakedb/gosnowflake/blob/ae006c15b53408aeed184228369d72fb0e8a3c1e/restful.go#L273-L277)
3.  Checking query status (look at `GetQueryStatus`)
4. Heartbeat
5. Async queries (`WithAsyncMode`) - I'm not sure if your code is sufficient enough, maybe it is, maybe not :)
6. Calling OCSP cache server/retry proxy.

I believe that calling external resources (IdP, cloud providers, OCSP responders etc) is not relevant here. However, this is the rough list, you have to verify on your own where the trace should be propagated and where not.

Also bonus question - for retries, the spanId shouldn't change, right?

sfc-gh-mkarjikar: I intend for this change to support only queries. Asynchronous queries will be nice to support as well, but I intend to leave this for a future PR if it is not already supported. 

Yes, for retries, the spanId shouldn't change.

connection_test.go
sfc-gh-mkarjikar: Yes we should, good catch. I've reset the global tracer provider in `t.cleanup` in most recent commit.

connection.go
sfc-gh-mkarjikar: Added a test for async queries and update doc.go to include an example as well.

sfc-gh-pfus: I'm not sure if this is correct. Async context is checked before calling `exec` (see `ExecContext` or `QueryContext` functions). So if you test on `exec` function level, we don't need to check `WithAsync`. However, I'm still considering if we need to add this propagation context only to query requests - what about getting results for longer queries? If a query takes more than 45s, GS returns queryID and the code 333333 (sync queries) or 333334 (async queries) and driver is responsible for checking query status. Should the trace context also be included here? Take a look at `TestLongRunningAsyncQuery` (BTW - it should work in both sync and async, the test is insufficient).

connection_test.go
```diff
@@ -271,6 +274,65 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+		rest:              sr,
+		queryContextCache: (&queryContextCache{}).init(),
+	}
+
+	// Tests synchronous query
+	_, err1 := sc.exec(ctx, "", false /* noResult */, false, /* isInternal */
+		false /* describeOnly */, nil)
+	assertNilF(t, err1)
+
+	// Tests asynchronous query
```
sfc-gh-pfus, line 326: If we test on this level, we can get rid of this.

connection_test.go
```diff
@@ -271,6 +274,65 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+	_, err2 := sc.exec(WithAsyncMode(ctx), "", false /* noResult */, false, /* isInternal */
+		false /* describeOnly */, nil)
+	assertNilF(t, err2)
+
+	t.Cleanup(func() {
```
sfc-gh-pfus, line 331: It has to be registered earlier - right after first `SetTraceProvider` is called, otherwise we can never reach this cleanup (if any `assertNilF` fails above).

connection.go
sfc-gh-bnandibhatla: @sfc-gh-pfus we want to start with passing the context with just while executing the query and will later expand it to other cases like long running queries in future. For now, we basically want to mimic what we have in the python connector library. I will remove the `WithAsync` test.

sfc-gh-pfus: Who will add this support and when? I don't want to be left with partially implemented feature forever :) Why not add it now?

sfc-gh-bnandibhatla: We want to add tracing only for normal queries to help our customers with debugging issues. This is what we did in the Python Connector and our customers seem satisfied with it. Adding tracing in all the unnecessary places will make the traces noisy and not really help with debugging. This is why we want to keep this change simple and not add in all the places upfront.

sfc-gh-pfus: But long-running queries are normal queries. It's common for a query to take more than 45 seconds. I also don't want to have issues that our team will have to fix. If the whole change was implemented by our team, I would be fine with cutting the scope now and finish later, but if this is an external change, it would be best to cover everything by your team now.

PR #1361, opened 2025-04-09
SNOW-2017056 Validate secure storage lock dir owner

sfc-gh-pfus:
### Description

SNOW-2017056 Validate secure storage lock dir owner. Otherwise, if some attacker created a lock file dir and kept recreating it, the driver would always be hanging on locking secure storage.

### Checklist


secure_storage_manager.go
```diff
@@ -268,18 +268,40 @@ func (ssm *fileBasedSecureStorageManager) lockFile() error {
...
+
+		ownerUID, err := provideFileOwner(lockFile)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+		currentUser, err := user.Current()
+		if err != nil {
+			return err
+		}
+		if errors.Is(err, os.ErrNotExist) {
```
sfc-gh-dstempniak, line 290: Is it ok? The previous `if` returns `err` if `err != nil`.

secure_storage_manager.go
```diff
@@ -268,18 +268,40 @@ func (ssm *fileBasedSecureStorageManager) lockFile() error {
...
+			return nil
+		}
+		if strconv.Itoa(int(ownerUID)) != currentUser.Uid {
+			return errors.New("incorrect owner of " + lockFile.Name())
+		}
+
+		// removing stale lock
+		now := time.Now()
+		if !errors.Is(err, os.ErrNotExist) && fileInfo.ModTime().Add(time.Second).UnixNano() < now.UnixNano() {
```
sfc-gh-dstempniak, line 299: Same here. `err` is the result of `user.Current()`

secure_storage_manager.go
sfc-gh-pfus: Actually, this `if` doesn't make any sense - we know that the file exists from the check above.

secure_storage_manager.go
sfc-gh-pfus: Same.

secure_storage_manager.go
```diff
@@ -268,18 +268,37 @@ func (ssm *fileBasedSecureStorageManager) lockFile() error {
 	const retryInterval = 100 * time.Millisecond
 	lockPath := ssm.lockPath()
 
-	fileInfo, err := os.Stat(lockPath)
+	lockFile, err := os.Open(lockPath)
```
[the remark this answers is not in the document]
sfc-gh-pfus: Good catch too, fixed.

sfc-gh-dheyman (approved):

sfc-gh-dstempniak (approved):

PR #1376, opened 2025-04-17
fix

sfc-gh-ext-simba-jy:
### Description

### Checklist

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -114,7 +114,7 @@ type Config struct {
 
 	DisableSamlURLCheck ConfigBool // Indicates whether the SAML URL check should be disabled
 
-	GCSUseVirtualEndPoint ConfigBool // When true, the virtual endpoint url is used, see: https://cloud.google.com/storage/docs/request-endpoints#xml-api
+	GcsUseVirtualEndPoint ConfigBool // When true, the virtual endpoint url is used, see: https://cloud.google.com/storage/docs/request-endpoints#xml-api
 }
 
 // Validate enables testing if config is correct.
@@ -289,8 +289,8 @@ func DSN(cfg *Config) (dsn string, err error) {
 	if cfg.DisableSamlURLCheck != configBoolNotSet {
 		params.Add("disableSamlURLCheck", strconv.FormatBool(cfg.DisableSamlURLCheck != ConfigBoolFalse))
 	}
-	if cfg.GCSUseVirtualEndPoint != configBoolNotSet {
-		params.Add("gcsUseVirtualEndPoint", strconv.FormatBool(cfg.DisableSamlURLCheck != ConfigBoolFalse))
+	if cfg.GcsUseVirtualEndPoint != configBoolNotSet {
+		params.Add("gcsUseVirtualEndPoint", strconv.FormatBool(cfg.GcsUseVirtualEndPoint != ConfigBoolFalse))
 	}
 
 	dsn = fmt.Sprintf("%v:%v@%v:%v", url.QueryEscape(cfg.User), url.QueryEscape(cfg.Password), cfg.Host, cfg.Port)
@@ -884,6 +884,17 @@ func parseDSNParams(cfg *Config, params string) (err error) {
 				cfg.DisableSamlURLCheck = ConfigBoolTrue
 			} else {
 				cfg.DisableSamlURLCheck = ConfigBoolFalse
+			}
+		case "gcsUseVirtualEndPoint":
+			var vv bool
+			vv, err = strconv.ParseBool(value)
+			if err != nil {
+				return
+			}
+			if vv {
+				cfg.GcsUseVirtualEndPoint = ConfigBoolTrue
+			} else {
+				cfg.GcsUseVirtualEndPoint = ConfigBoolFalse
 			}
 		default:
 			if cfg.Params == nil {
--- a/dsn.go
+++ b/dsn.go
@@ -113,6 +113,8 @@ type Config struct {
 	DisableConsoleLogin ConfigBool // Indicates whether console login should be disabled
 
 	DisableSamlURLCheck ConfigBool // Indicates whether the SAML URL check should be disabled
+
+	GCSUseVirtualEndPoint ConfigBool // When true, the virtual endpoint url is used, see: https://cloud.google.com/storage/docs/request-endpoints#xml-api
 }
 
 // Validate enables testing if config is correct.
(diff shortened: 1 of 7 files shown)


dsn.go
```diff
@@ -113,6 +113,8 @@ type Config struct {
 	DisableConsoleLogin ConfigBool // Indicates whether console login should be disabled
 
 	DisableSamlURLCheck ConfigBool // Indicates whether the SAML URL check should be disabled
+
+	GcsUseVirtualEndPoint ConfigBool // When true, the virtual endpoint url is used, see: https://cloud.google.com/storage/docs/request-endpoints#xml-api
```
sfc-gh-pfus, line 117: `ConfigBool` is a good option when we want to have 'not set' option. This is useful in case of token cache, because not setting value behaves differently on windows/mac and linux. However, in that case we want just false as a default, right? In that case, it can be just a bool.

gcs_storage_client.go
```diff
@@ -393,10 +393,16 @@ func (util *snowflakeGcsClient) extractBucketNameAndPath(location string) *gcsLo
...
 func (util *snowflakeGcsClient) generateFileURL(stageInfo *execResponseStageInfo, filename string) (*url.URL, error) {
+	var URL *url.URL
+	var err error
 	gcsLoc := util.extractBucketNameAndPath(stageInfo.Location)
 	fullFilePath := gcsLoc.path + filename
-	endPoint := getGcsCustomEndpoint(stageInfo)
-	URL, err := url.Parse(endPoint + "/" + gcsLoc.bucketName + "/" + url.QueryEscape(fullFilePath))
+	endPoint := util.getGcsCustomEndpoint(stageInfo)
+	if util.cfg.GcsUseVirtualEndPoint == ConfigBoolTrue {
+		URL, err = url.Parse(endPoint + "/" + url.QueryEscape(fullFilePath))
```
sfc-gh-pfus, line 402: We can just `return url.Parse...` from both branches here and all lines below and vars declaration above would be unnecessary, right?

gcs_storage_client_test.go
```diff
@@ -118,6 +118,25 @@ func TestGenerateFileURL(t *testing.T) {
 			}
 		})
 	}
+	gcsUtil.cfg.GcsUseVirtualEndPoint = ConfigBoolTrue
+
+	for _, test := range testcases {
+		t.Run(test.location, func(t *testing.T) {
+			stageInfo := &execResponseStageInfo{}
+			stageInfo.Location = test.location
```
sfc-gh-pfus, line 126: So actually this location is a bucket name, right? If yes, can we rename it?

gcs_storage_client.go
```diff
@@ -408,18 +414,19 @@ func (util *snowflakeGcsClient) isTokenExpired(resp *http.Response) bool {
...
 
-func getGcsCustomEndpoint(info *execResponseStageInfo) string {
+func (util *snowflakeGcsClient) getGcsCustomEndpoint(info *execResponseStageInfo) string {
 	endpoint := "https://storage.googleapis.com"
 
 	// TODO: SNOW-1789759 hardcoded region will be replaced in the future
 	isRegionalURLEnabled := (strings.ToLower(info.Region) == gcsRegionMeCentral2) || info.UseRegionalURL
 	if info.EndPoint != "" {
 		endpoint = fmt.Sprintf("https://%s", info.EndPoint)
+	} else if util.cfg.GcsUseVirtualEndPoint == ConfigBoolTrue {
```
sfc-gh-pbulawa, line 427: The usage of virtual endpoint should be based on the `info` response, the same way as the `Region` below. Snowflake does not yet respond with this field, but we should be bound to it.

dsn.go
sfc-gh-pbulawa: The virtualUrls should be based on the response from the Snowflake, not the parameter. Please see https://github.com/snowflakedb/snowflake-connector-python/pull/2274

sfc-gh-pbulawa (changes requested):

gcs_storage_client.go
sfc-gh-ext-simba-jy: I used the old PR as a reference. I will update the my PR you provided. Thank you!

gcs_storage_client_test.go
```diff
@@ -118,6 +118,25 @@ func TestGenerateFileURL(t *testing.T) {
...
 		})
 	}
+
+	for _, test := range testcases {
+		t.Run(test.location, func(t *testing.T) {
+			stageInfo := &execResponseStageInfo{}
+			stageInfo.Location = test.location
+			stageInfo.UseVirtualURL = true
+			gcsURL, err := gcsUtil.generateFileURL(stageInfo, test.fname)
+			if err != nil {
```
sfc-gh-pfus, line 128: Can you use `assertNilF` for this?

gcs_storage_client_test.go
```diff
@@ -118,6 +118,25 @@ func TestGenerateFileURL(t *testing.T) {
...
+				t.Error(err)
```
sfc-gh-pfus, line 129: We should use `fatal` here. Error just marks the tests as failed, but continues execution and fatal stops execution immediately. Generally fatal is good for preconditions (like no error was returned), and errors are good for assertions (checking if values match expected).

gcs_storage_client_test.go
```diff
@@ -118,6 +118,25 @@ func TestGenerateFileURL(t *testing.T) {
...
+			}
+			expectedURL, err := url.Parse("https://sfc-eng-regression.storage.googleapis.com/" + url.QueryEscape(test.filepath))
+			if err != nil {
+				t.Error(err)
+			}
+			if gcsURL.String() != expectedURL.String() {
+				t.Fatalf("failed. expected: %v but got: %v", expectedURL.String(), gcsURL.String())
```
sfc-gh-pfus, line 136: `assertEqualE`?

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

gcs_storage_client_test.go
sfc-gh-ext-simba-jy: Done.

gcs_storage_client_test.go
sfc-gh-ext-simba-jy: Fixed.

gcs_storage_client_test.go
sfc-gh-ext-simba-jy: Replaced it!

PR #1386, opened 2025-04-28
SNOW-1887766 Implement logic to propagate trace context, if it exists, to GS

sfc-gh-bnandibhatla:
### Description

SNOW-1887766 

Added support for trace propagation by using OpenTelemetry's trace context injector for Go. This change adds one additional dependency for the Go driver, which is OpenTelemetry v1.35.0.

The only functional change is an extra couple headers are injected into the HTTPS request to GS, and GS is already equipped to extract it on the other side if it is present - i.e. the header "traceparent", where the value is in W3 Trace Context Format.

This change will allow trace context to be propagated from anywhere the Go driver is being used to GS. The primary use case is for a Go service running on SPCS. The user will then be able to see the entire line of execution in Snowsight in one view, from their SPCS Service to the SQL queries they ran using the connector.

This is a feature request from Relational AI, an external customer.

The compilation time and size as well as the runtime were benchmarked before and after the change using the cm/select1 program. These were tested (built and ran) 5 times each.

Compilation/Runtime for cmd/select1 before/after these changes:

* Before changes (times are averaged):

    * Compilation Time: 2.21 seconds
    * Runtime: 0.8 seconds
    * Compilation Size: 33.35MB
* With changes (times are averaged):

    * Compilation Time: 2.25 seconds
    * Runtime: 0.82 seconds
    * Compilation Size: 33.43MB

### Checklist


doc.go
```diff
@@ -315,6 +315,20 @@ query/command by Ctrl+C, add a os.Interrupt trap in context to execute methods t
...
+A context containing OpenTelemetry headers for distributed tracing can be
+created. Each query, both synchronous and asynchronous, run with this context
+will include the Trace ID and Span ID as metadata. If you are instrumenting your
+program with OpenTelemetry and exporting telemetry data to Snowflake, then
+queries run with this context will be properly nested under the appropriate
+parent span. This can be viewed in the Traces and Logs tab in Snowsight.
+
+For example:
+
+	ctx, parent_span := tracer.Start(context.Background(), "parent_span")
```
sfc-gh-pfus, line 329: Missing `defer parent_span.End()`?

connection_test.go
```diff
@@ -271,6 +274,58 @@ func TestExecWithSpecificRequestID(t *testing.T) {
...
+		FuncPostQuery: postQueryMock,
+	}
+
+	sc := &snowflakeConn{
+		cfg:               &Config{Params: map[string]*string{}},
+		rest:              sr,
+		queryContextCache: (&queryContextCache{}).init(),
+	}
+
+	_, err1 := sc.exec(ctx, "", false /* noResult */, false, /* isInternal */
```
sfc-gh-pfus, line 324: `err` without `1`?

connection_test.go
```diff
@@ -271,6 +274,58 @@ func TestExecWithSpecificRequestID(t *testing.T) {
 	}
 }
 
+func TestExecContextPropagationIntegrationTest(t *testing.T) {
+	originalTracerProvider := otel.GetTracerProvider()
+	t.Cleanup(func() {
```
sfc-gh-pfus, line 279: Actually it should be after `otel.SetTracerProvider(tp)` line, right?

doc.go
sfc-gh-bnandibhatla: Added.

sfc-gh-pfus (approved):

sfc-gh-pmotacki (approved):
LGTM

PR #1404, opened 2025-05-12
NO-SNOW: chunkDownloader: Wrap errors for better introspection

mattayes:
### Description

I'm experiencing a situation where chunk downloads are intermittently stalling: they hang for 10min+, and eventually fail. We've worked around the issue so far by adding a timeout to the query. Queries still hang, now returning `context.DeadlineExceeded` after the timeout.

We were able to track the error down to [chunk_downloader.go/downloadChunk](https://github.com/snowflakedb/gosnowflake/blob/2b9978d91fabd689f04ef5c2f341751841627e86/chunk_downloader.go#L366-L369). However, due to the lack of error wrapping, it's unclear where the stalling and error are coming from.

This PR:

* Wrap errors for better introspection.
* Check error equality using `errors.Is()`.
* Miscellaneous cleanup.

### Checklist
- [x] [N/A] Created tests which fail without the change (if possible)
- [x] [N/A] Extended the README / documentation, if necessary


mattayes:

mattayes:
recheck

chunk_downloader.go
```diff
@@ -621,8 +631,8 @@ func useStreamDownloader(ctx context.Context) bool {
 	if val == nil {
 		return false
 	}
-	s, ok := val.(bool)
-	return s && ok
+	s, _ := val.(bool)
```
sfc-gh-pfus, line 634: What is this change for?

chunk_downloader.go
```diff
@@ -691,42 +701,47 @@ func copyChunkStream(body io.Reader, rows chan<- []*string) error {
...
-	} else {
-		source = bufStream
 	}
 	r := io.MultiReader(strings.NewReader("["), source, strings.NewReader("]"))
 	dec := json.NewDecoder(r)
 	openToken := json.Delim('[')
 	closeToken := json.Delim(']')
 	for {
-		if t, err := dec.Token(); err == io.EOF {
+		t, err := dec.Token()
```
sfc-gh-pfus, line 721: I'm curious, why this syntax is better? Does it behave differently?

chunk_downloader.go
```diff
@@ -788,14 +803,13 @@ func usesArrowBatches(ctx context.Context) bool {
...
 	}
-	a, ok := val.(bool)
-	return a && ok
+	a, _ := val.(bool)
+	return a
 }
 
-func countArrowBatchRows(recs *[]arrow.Record) int {
-	var cnt int
+func countArrowBatchRows(recs *[]arrow.Record) (cnt int) {
```
sfc-gh-pfus, line 810: Same here - why this syntax is better?

sfc-gh-pfus (commented):
Thanks for this change! It makes our error handling easier to track. I left only a couple of minor comments.

chunk_downloader.go
mattayes: @sfc-gh-pfus It's the pattern used earlier in the file in `snowflakeChunkDownloader.totalUncompressedSize()`. Wanted to keep things consistent, and picked the first example.

chunk_downloader.go
mattayes: When a type assertion fails, it returns `<zero-value>, false`. For booleans, this becomes `false, false`. We can return the value by itself for the same effect as `s && ok`.

sfc-gh-pfus (approved):

mattayes:
@sfc-gh-pfus thanks for the review! It looks like there's an issue with Jenkins (or `master`): these three checks are failing for all open PRs. Otherwise we're good to go 😁.

sfc-gh-pbulawa (approved):

sfc-gh-pfus:
@mattayes I merged your PR, thank you!

PR #1405, opened 2025-05-12
NO-SNOW: encrypt_util: Fix bugs where errors are never returned

mattayes:
### Description

Fix bugs in [encryptStreamCBC()](https://github.com/snowflakedb/gosnowflake/blob/2b9978d91fabd689f04ef5c2f341751841627e86/encrypt_util.go#L63-L67) and [decryptStreamCBC()](https://github.com/snowflakedb/gosnowflake/blob/2b9978d91fabd689f04ef5c2f341751841627e86/encrypt_util.go#L272-L299).

`encryptStreamCBC()` ignores all read errors.

For `decryptStreamCBC()`, as far as I can tell, the original intent is for `0, err` to be returned when `src.Read()` returns an error. Unfortunately, `err` is rescoped inside the `for` loop when created using `:=`, and does not reuse the outer `err`. Since we checked if `err != nil` before the `for` loop, the check after the `for` loop is always true, and we never return the error.

This change returns `0, err` when `src.Read()` returns a non-`io.EOF` error. It keeps the original behavior otherwise.

### Checklist
- [x] [N/A] Extended the README / documentation, if necessary


mattayes:
recheck

encrypt_util.go
```diff
@@ -63,9 +63,14 @@ func encryptStreamCBC(
 		// read the stream buffer up to len(chunk) bytes into chunk
 		// note that all spaces in chunk may be used even if Read() returns n < len(chunk)
 		n, err := src.Read(chunk)
-		if n == 0 || err != nil {
+		if err != nil && err != io.EOF {
+			return nil, fmt.Errorf("reading: %w", err)
+		}
+		if n == 0 {
```
sfc-gh-pfus, line 69: When this situation may happen? Only when EOF is returned?

encrypt_util.go
```diff
@@ -279,28 +284,31 @@ func decryptStreamCBC(
...
-	}
+
 	if prevChunk != nil {
 		totalFileSize -= paddingOffset(prevChunk)
 	}
-	return totalFileSize, err
+	return totalFileSize, nil
 }
 
 func encryptGCM(iv []byte, plaintext []byte, encryptionKey []byte, aad []byte) ([]byte, error) {
```
sfc-gh-pfus, line 314: Can you ensure that GCM encryption is not flawed with the same? We are going go migrate to GCM encryption later this year.

mattayes: @sfc-gh-pfus Looked through the code, I don't see this issue in the GCM encryption/decryption code 👍🏻.

sfc-gh-pfus (approved):

encrypt_util_test.go
```diff
@@ -118,6 +120,47 @@ func TestEncryptDecryptLargeFileCBC(t *testing.T) {
...
+
+	wantErr := errors.New("test error")
+	r := iotest.ErrReader(wantErr)
+
+	n, err := encryptStreamCBC(&sfe, r, nil, 0)
+	assertTrueF(t, errors.Is(err, wantErr), fmt.Sprintf("expected error: %v, got: %v", wantErr, err))
+	assertEqualE(t, n, 0, "expected 0 bytes written")
+}
+
+func TestDecryptStreamCBDReadError(t *testing.T) {
```
sfc-gh-pfus, line 138: CBC?

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

sfc-gh-pfus:
@mattayes merged! Thank you for your input!

Issue #1406, opened 2025-05-13
Nil pointer error while uploading files


nearsyh:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of GO driver are you using?

1.13.3
   
2. What operating system and processor architecture are you using?

Linux AMD
   
3. What version of GO are you using?
run `go version` in your console

1.22.4

4.Server version:* E.g. 1.90.1
You may get the server version by running a query:
```
9.12.1
```
5. What did you do?

Use `PUT` to upload files to stage. I can't reliability reproduce it.

6. What did you expect to see?

   I expect to upload the file successfully, but got this error

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x188 pc=0x11b34ad]
goroutine 71357 [running]:
github.com/snowflakedb/gosnowflake.withCloudStorageTimeout[...](0x2?, 0xc197cd5604?)
	/go/src/github.com/snowflakedb/gosnowflake/file_transfer_agent.go:633 +0x2d
github.com/snowflakedb/gosnowflake.(*snowflakeS3Client).getFileHeader(0xc001b7ee38, 0xc197d28000, {0xc197cd5604?, 0xc000451b28?})
	/go/src/github.com/snowflakedb/gosnowflake/s3_storage_client.go:120 +0xf5
github.com/snowflakedb/gosnowflake.(*remoteStorageUtil).uploadOneFile(0x3b9aca00?, 0xc197d28000)
	/go/src/github.com/snowflakedb/gosnowflake/storage_client.go:68 +0x1c2
github.com/snowflakedb/gosnowflake.(*remoteStorageUtil).uploadOneFileWithRetry(0xc19851c008, 0xc197d28000)
	/go/src/github.com/snowflakedb/gosnowflake/storage_client.go:119 +0x188
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).uploadOneFile(0xc1882e8f20, 0xc197d28000)
	/go/src/github.com/snowflakedb/gosnowflake/file_transfer_agent.go:885 +0x203
github.com/snowflakedb/gosnowflake.(*snowflakeFileTransferAgent).uploadFilesSequential(0xc1882e8f20, {0xc1870ed0b8, 0x1, 0xc1b26a0008?})
[...]

PR #1410, opened 2025-05-15
SNOW-2102260 don't delete config in cleanup

sfc-gh-dszmolka:
### Description

SNOW-2102260 ,  issue #1406 

It appears that under certain conditions, a race can be introduced between closing a connection (especially if the context has been created with timeout or cancel) and creating/using the cloud storage client e.g. for a PUT operation.
This can result in a panic due to the cloud storage client created without a Config which was previously deleted from the connection in the cleanup phase; when it was cancelled or timed out - in such a case; the cloud storage client is initialized with `nil` config and is reproducibly panicked. 

The change tries to address the situation by not removing the Config upon cleanup, so any dependent checks can proceed.

Change was manually tested on the reproduction which consistently reproduces the panic mentioned in #1406 ; with the patch, the panic doesn't happen anymore. 

I'd like to thank my colleague @sfc-gh-pfus for his valuable insights and guidance while debugging / addressing this !

### Checklist

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -278,7 +278,6 @@ func (sc *snowflakeConn) cleanup() {
 		sc.rest.Client.CloseIdleConnections()
 	}
 	sc.rest = nil
-	sc.cfg = nil
 }
 
 func (sc *snowflakeConn) Close() (err error) {
--- a/connection.go
+++ b/connection.go
@@ -277,7 +277,6 @@ func (sc *snowflakeConn) cleanup() {
 	if sc.rest != nil && sc.rest.Client != nil {
 		sc.rest.Client.CloseIdleConnections()
 	}
-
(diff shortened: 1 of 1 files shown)


connection.go
```diff
@@ -278,7 +278,6 @@ func (sc *snowflakeConn) cleanup() {
 		sc.rest.Client.CloseIdleConnections()
 	}
 	sc.rest = nil
```
sfc-gh-pfus, line 280: I'd remove this one as well. This time the processing failed on missing `cfg`, but missing `rest` may cause the same problem (if cleanup is invoked just before some REST call).

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

PR #1418, opened 2025-05-21
SNOW-2085893 Address OCSP check vulnerability in GoSnowflake - falsely invalidated host certificate chain containing a trusted certificate in the middle but not at the tail

sfc-gh-ext-simba-jy:
### Description

SNOW-2085893

### Checklist

### DIFF
--- a/ocsp_test.go
+++ b/ocsp_test.go
@@ -669,9 +669,7 @@ func TestVerifyPeerCertificateChainOrdering(t *testing.T) {
 	ln, err := tls.Listen("tcp", "127.0.0.1:0", &tls.Config{
 		Certificates: []tls.Certificate{serverCert},
 	})
-	if err != nil {
-		t.Fatalf("failed to start server: %v", err)
-	}
+	assertNilF(t, err, "failed to start server")
 	defer ln.Close()
 
 	go func() {
@@ -680,12 +678,13 @@ func TestVerifyPeerCertificateChainOrdering(t *testing.T) {
 			if err != nil {
 				return
 			}
-			go func(c net.Conn) {
-				defer c.Close()
-				buf := make([]byte, 1)
-				c.SetReadDeadline(time.Now().Add(5 * time.Second))
-				c.Read(buf)
-			}(conn)
+
+			tlsConn := conn.(*tls.Conn)
+			defer tlsConn.Close()
+			assertNilF(t, tlsConn.Handshake(), "failed to set deadline on TLS connection")
+
+			time.Sleep(2 * time.Second)
+			tlsConn.Close()
 		}
 	}()
 
@@ -725,14 +724,10 @@ func TestVerifyPeerCertificateChainOrdering(t *testing.T) {
 
 	// Client dials server
 	conn, err := tls.Dial("tcp", ln.Addr().String(), conf)
-	if err != nil {
-		t.Fatalf("tls.Dial failed: %v", err)
-	}
-	conn.Close()
+	defer conn.Close()
 
-	if !called {
-		t.Fatalf("VerifyPeerCertificate callback was not called")
-	}
+	assertNilF(t, err, "failed to dial server")
+	assertTrueF(t, called, "VerifyPeerCertificate callback was not called")
 }
 
 func syncUpdateOcspResponseCache(f func()) {


ocsp.go
```diff
@@ -687,15 +687,30 @@ func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certifi
...
+		if !isCA || !isIssuer {
 			// Check if the last Non Root Cert is also a CA or is self signed.
 			// if the last certificate is not, add it to the list
-			rca := caRoot[string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer)]
 			if rca == nil {
 				return fmt.Errorf("failed to find root CA. pkix.name: %v", verifiedChains[i][numberOfNoneRootCerts].Issuer)
 			}
 			verifiedChains[i] = append(verifiedChains[i], rca)
 			numberOfNoneRootCerts++
+		} else if isCA && isIssuer {
```
sfc-gh-pfus, line 702: A couple of thoughts:
1. You check only the last but one certificate, but the trusted root may also be at any position, right? In our case it was the third in the chain out of five.
2. I'm not sure it has to be self issuer. In our case the cert was cross signed, and the version presented by AWS was still the one signed by Starfield, not self signed.

Maybe I'm missing something magical here. Have you had a chance to run tests described in https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/1267 ?

ocsp.go
```diff
@@ -687,15 +687,30 @@ func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certifi
 		numberOfNoneRootCerts := len(verifiedChains[i]) - 1
 		logger.Tracef("checking cert, %v, %v, isCa: %v, rawIssuer: %v, rawSubject: %v", i, numberOfNoneRootCerts, verifiedChains[i][numberOfNoneRootCerts].IsCA, string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer), string(verifiedChains[i][numberOfNoneRootCerts].RawSubject))
 		logger.Tracef("checking cert, base64, rawIssuer: %v, rawSubject: %v", base64.StdEncoding.EncodeToString(verifiedChains[i][numberOfNoneRootCerts].RawIssuer), base64.StdEncoding.EncodeToString(verifiedChains[i][numberOfNoneRootCerts].RawSubject))
-		if !verifiedChains[i][numberOfNoneRootCerts].IsCA || string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer) != string(verifiedChains[i][numberOfNoneRootCerts].RawSubject) {
+		isCA := verifiedChains[i][numberOfNoneRootCerts].IsCA
+		isIssuer := string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer) == string(verifiedChains[i][numberOfNoneRootCerts].RawSubject)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: You are right. I replaced the variable name with isSelfSigned.

ocsp.go
sfc-gh-ext-simba-jy: Hmmm..  Actually, I followed the same implementation as in the Python and JDBC drivers. I did not take a look at this issue in detail. (Python)[https://github.com/snowflakedb/snowflake-connector-python/pull/2299/files], and (JDBC)[https://github.com/snowflakedb/snowflake-jdbc/pull/2189/files]. If you do not mind, could you please have a look at these PRs?

sfc-gh-ext-simba-jy: Hello @sfc-gh-pfus, you are right. Unlike other drivers that iterate through the entire certificate chain, the Go driver only validates the last certificate.

To address this, my updated approach first checks if the last certificate is the Root CA. If not, it traverses the chain from the start and reorders it by placing the Root CA at the end. Please review my update. Thank you.

ocsp.go
```diff
@@ -687,15 +687,27 @@ func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certifi
...
+		isCA := verifiedChains[i][numberOfNoneRootCerts].IsCA
+		isSelfSigned := string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer) == string(verifiedChains[i][numberOfNoneRootCerts].RawSubject)
+
+		if !isCA || !isSelfSigned {
+			// The last cert is not the Root CA certificate. Search for the Root CA certificate in the chain.
+			logger.Debugf("Certificate chain is not signed by Root CA. Searching for Root CA in the chain: %v", verifiedChains[i][numberOfNoneRootCerts].Issuer)
+			// Search for the Root CA certificate in the chain.
+			for j := 0; j < numberOfNoneRootCerts; j++ {
+				cert := verifiedChains[i][j]
+				isRootCA := cert.IsCA && string(cert.RawIssuer) == string(cert.RawSubject)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: You are correct. I totally misunderstood this issue. I have fixed the code and let you know if I am ready. Thank you for your feedback.

ocsp.go
```diff
@@ -687,15 +687,30 @@ func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certifi
 		numberOfNoneRootCerts := len(verifiedChains[i]) - 1
 		logger.Tracef("checking cert, %v, %v, isCa: %v, rawIssuer: %v, rawSubject: %v", i, numberOfNoneRootCerts, verifiedChains[i][numberOfNoneRootCerts].IsCA, string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer), string(verifiedChains[i][numberOfNoneRootCerts].RawSubject))
 		logger.Tracef("checking cert, base64, rawIssuer: %v, rawSubject: %v", base64.StdEncoding.EncodeToString(verifiedChains[i][numberOfNoneRootCerts].RawIssuer), base64.StdEncoding.EncodeToString(verifiedChains[i][numberOfNoneRootCerts].RawSubject))
-		if !verifiedChains[i][numberOfNoneRootCerts].IsCA || string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer) != string(verifiedChains[i][numberOfNoneRootCerts].RawSubject) {
+		isCA := verifiedChains[i][numberOfNoneRootCerts].IsCA
+		isSelfSigned := string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer) == string(verifiedChains[i][numberOfNoneRootCerts].RawSubject)
+
+		if !isCA || !isSelfSigned {
```
sfc-gh-pfus: Why do we need this condition? My suggestion is:
1. Remove this condition and the condition from line 697.
2. Run the loop for all certificates but the last one (as it is now).

Apart from this, your change seems to be fine. Have you had a chance to battle-test it with the scenario described in the issue?

As a side note - don't need to fix it, rather FYI, but:
1. Instead of using `string(cert.RawIssuer)` you can use `cert.Issuer.String()`, but both produces different strings, so we would have to chance `caRoot` key too (and all other usages of `caRoot`).
2. Golang's `x509.Verify` should already present correct chain to `verifyPeerCertificate`. Could you test your code before your change and after (according to test scenario from the issue)? I have an impression, that it would work in both cases, because Golang should build the same chain here (can be verified in debugger).

sfc-gh-ext-simba-jy: Thank you for your kind feedback. I researched point number 2 you mentioned, and you are correct. Since the Go driver does not control InsecureSkipVerify (which is false by default), I believe that x509.Verify is always executed (which is why we can get the chain values from verifiedChains). In this case, the root certificate always appears at the end of the chain because of the Verify function. Therefore, I believe this change is unnecessary, as the root CA cannot exist in the middle of the chain. Do you think we can write up this summary on the ticket and close this PR? @sfc-gh-pfus

sfc-gh-pfus: I'm not sure. Did you find it empirically or in Go's spec/doc? If only empirically, then we can't depend on this behaviour, if it is consistent and won't change in the future (as a reference - python gives the full chain in the same situation).

sfc-gh-ext-simba-jy: It is from the go doc. [here](https://pkg.go.dev/crypto/x509#Certificate.Verify)

PR #1420, opened 2025-05-21
SNOW-1029631: [gosnowflake][GH#536] optimize PUT memory usage / PUT operations read full input data into memory

sfc-gh-ext-simba-jy:
### Description

### Checklist


file_transfer_agent.go
```diff
@@ -354,15 +355,39 @@ func (sfa *snowflakeFileTransferAgent) initFileMetadata() error {
...
-				srcStream:         sfa.sourceStream,
-				srcFileSize:       srcFileSize,
-				stageLocationType: sfa.stageLocationType,
-				stageInfo:         sfa.stageInfo,
-			})
+			fileInfo, err := os.Stat(fileName)
+			//This means the file is not a local file, but a stream. Ex) Bulk Insert.
+			if err != nil {
+				buf := new(bytes.Buffer)
+				_, err := buf.ReadFrom(sfa.sourceStream)
```
sfc-gh-pfus, line 366: This operation still needs to load the full stream into memory, right?

file_util.go
```diff
@@ -154,6 +155,7 @@ type fileMetadata struct {
 	options            *SnowflakeFileTransferOptions
 
 	/* streaming PUT */
+	fileStream    io.Reader
```
sfc-gh-pfus, line 158: Do we need both fileStream and srcStream?

file_transfer_agent.go
sfc-gh-ext-simba-jy: Yes, this logic is originally intended for bulk inserts, where an INSERT query is converted to a PUT query if the number of bindings exceeds a certain threshold. However, unlike other drivers, the Go driver also uses this workflow for PUT streaming. As a result, it doesn't load the entire dataset into memory when executing a PUT query.

file_util.go
sfc-gh-ext-simba-jy: Yes, this is intended to retain the I/O object passed by the customer. The srcStream holds the data read from the file stream. Currently, we need both of them. In my opinion, we may rename 'srcStream'.

file_transfer_agent.go
sfc-gh-pfus: > As a result, it doesn't load the entire dataset into memory when executing a PUT query.
So does or doesn't it load the full stream into memory? My concern is that it does, so if we get PUT with a file that is 10GB large, it will explode, right?

sfc-gh-ext-simba-jy: Yes. As I mentioned, this code is designed for bulk inserts. According to the logic, if sourceStream exists, the driver attempts to read the file's stat. If the file exists on the local machine (indicating a PUT command), the driver does not load the entire data into memory. Instead, it saves the I/O stream(see line 382). If the file does not exist, it means the file is not local, and the original query was an INSERT statement that was internally converted to a PUT command for bulk insertion. As a result, if a user executes the PUT command with 10GB of data, the driver will not load the whole data into the memory (I ran the put testing [here](https://github.com/snowflakedb/gosnowflake/pull/1420/files/b1aa0e3d6d309dd2ec51086eff12769035558c6b) with a 1GB file locally and confirmed the driver does not use the high volume of memory anymore.

sfc-gh-pfus: Ok, makes sense now, thank you! Can you add a short comment to the code?

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

file_transfer_agent.go
sfc-gh-ext-simba-jy: I will. Thank you!

PR #1423, opened 2025-05-23
SNOW-2118577 fix nil pointer dereference

obs-gh-felixbowing:
Problem: In rare cases err can be context.DeadlineExceeded but ctx.Err()==nil still holds. (E.g. when some
downstream operation had a tighter timeout and returns an error, but the passed ctx here is still not done). In this case this code would return nil for both
return values. This can cause nil pointer dereferences in the caller, as the expectation is that
exactly one is not nil.

Solution: Always return what
FuncPostQueryHelper returns, but try to cancel
the query if the returned error was a cancellation or timeout. In case cancellation fails,
show both errors to the caller.

### DIFF
--- a/restful.go
+++ b/restful.go
@@ -202,15 +202,15 @@ func postRestfulQuery(
 
 	data, err = sr.FuncPostQueryHelper(ctx, sr, params, headers, body, timeout, requestID, cfg)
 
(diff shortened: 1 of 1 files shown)


obs-gh-felixbowing:
recheck

sfc-gh-pfus (approved):

PR #1424, opened 2025-05-26
SNOW-1902248: Add support for GCP & OIDC Workload Identity authentication

sfc-gh-dheyman:
### Description
SNOW-1902248

### Checklist

### DIFF
--- a/auth_wif.go
+++ b/auth_wif.go
@@ -1,0 +1,250 @@
+package gosnowflake
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"net/http"
+	"strings"
+	"time"
+)
+
+const (
+	awsWif   wifProviderType = "AWS"
+	gcpWif   wifProviderType = "GCP"
+	azureWif wifProviderType = "AZURE"
+	oidcWif  wifProviderType = "OIDC"
+)
+
+type wifProviderType string
+
+type wifAttestation struct {
+	ProviderType string            `json:"providerType"`
+	Credential   string            `json:"credential"`
+	Metadata     map[string]string `json:"metadata"`
+}
+
+type wifAttestationCreator interface {
+	createAttestation(ctx context.Context) (*wifAttestation, error)
+}
+
+type wifAttestationProvider struct {
+	context      context.Context
+	awsCreator   wifAttestationCreator
+	gcpCreator   wifAttestationCreator
+	azureCreator wifAttestationCreator
+	oidcCreator  wifAttestationCreator
+}
+
+func createWifAttestationProvider(ctx context.Context) *wifAttestationProvider {
+	return &wifAttestationProvider{
+		context:      ctx,
+		awsCreator:   &awsIdentityAttestationCreator{attestationService: createDefaultAwsAttestationService(ctx)},
+		gcpCreator:   nil,
+		azureCreator: nil,
+		oidcCreator:  nil,
+	}
+}
+
+func (p *wifAttestationProvider) getAttestation(identityProvider string) (*wifAttestation, error) {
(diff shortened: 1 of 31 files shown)


auth_wif.go
```diff
@@ -248,3 +268,103 @@ func (creator *awsIdentityAttestationCreator) createBase64EncodedRequestCredenti
...
+	}
+	return string(body)
+}
+
+func extractSubIssWithoutVerifyingSignature(token string) (subject string, issuer string, err error) {
+	claims, err := extractClaimsMap(token)
+	if err != nil {
+		return "", "", err
+	}
+	issuer, ok := claims["iss"].(string)
```
sfc-gh-pfus, line 329: Is it fine? I had an impression it may fail only if issuer is not a string, not if it is missing?

auth_wif.go
```diff
@@ -248,3 +268,103 @@ func (creator *awsIdentityAttestationCreator) createBase64EncodedRequestCredenti
...
+	subject, ok = claims["sub"].(string)
+	if !ok {
+		return "", "", errors.New("missing sub claim in JWT token")
+	}
+	return subject, issuer, nil
+}
+
+// extractClaimsMap parses a JWT token and returns its claims as a map.
+// It does not verify the token signature.
+func extractClaimsMap(token string) (map[string]interface{}, error) {
```
sfc-gh-pfus, line 342: Is it a general function or specific for GCP?

auth_wif.go
sfc-gh-pfus: But there is also a cast here, and in your code there is not :)

sfc-gh-dheyman:
@sfc-gh-pfus Added also suggestions/tests from your AWS review.

auth_wif.go
sfc-gh-dheyman: True. Fixed by separating get and cast.

auth_wif.go
```diff
@@ -248,3 +267,103 @@ func (creator *awsIdentityAttestationCreator) createBase64EncodedRequestCredenti
...
+	}
+	issuerClaim, ok := claims["iss"]
+	if !ok {
+		return "", "", errors.New("missing issuer claim in JWT token")
+	}
+	subjectClaim, ok := claims["sub"]
+	if !ok {
+		return "", "", errors.New("missing sub claim in JWT token")
+	}
+	return issuerClaim.(string), subjectClaim.(string), nil
```
sfc-gh-pbulawa, line 336: Different order than in the function definition.

sfc-gh-pbulawa (changes requested):

auth_wif.go
sfc-gh-dheyman: fixed

auth_wif.go
```diff
@@ -132,84 +152,84 @@ func createDefaultAwsAttestationService(ctx context.Context) *defaultAwsAttestat
...
 		return nil, err
 	}
 
 	return &wifAttestation{
-		ProviderType: "AWS",
+		ProviderType: string(awsWif),
 		Credential:   credential,
 		Metadata:     map[string]string{"arn": arn},
 	}, nil
 }
```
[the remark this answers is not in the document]
sfc-gh-dheyman: i think it's a stretch, I would leave it as is.

auth_wif.go
```diff
@@ -248,3 +267,103 @@ func (creator *awsIdentityAttestationCreator) createBase64EncodedRequestCredenti
...
+	}
+	issuerClaim, ok := claims["iss"]
+	if !ok {
+		return "", "", errors.New("missing issuer claim in JWT token")
+	}
+	subjectClaim, ok := claims["sub"]
+	if !ok {
+		return "", "", errors.New("missing sub claim in JWT token")
+	}
+	return subjectClaim.(string), issuerClaim.(string), nil
```
sfc-gh-pfus, line 336: If you use naked return it would increase safety - why not leverage it in that case (returning two strings)?

dsn_test.go
```diff
@@ -97,6 +97,23 @@ func TestParseDSN(t *testing.T) {
 			ocspMode: ocspModeFailOpen,
 			err:      nil,
 		},
+		{
+			dsn: "u:p@/db?account=ac&workloadIdentityEntraResource=https%3A%2F%2Fexample.com%2F.default&workloadIdentityProvider=azure",
+			config: &Config{
```
sfc-gh-pfus, line 102: What about the second test (`TestDSN`)?

PR #1429, opened 2025-05-29
SNOW-2117152 Implement CRL algorithm

sfc-gh-pfus:
### Description

### Checklist


sfc-gh-jkasten (commented):
Made a quick pass. Thanks @sfc-gh-pfus!

crl.go
```diff
@@ -0,0 +1,119 @@
...
+			logger.Debugf("finished CRL validation for %v", subject)
+		}
+	}
+	return nil
+}
+
+func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x509.Certificate) error {
+	for _, distributionPoint := range cert.CRLDistributionPoints {
+		logger.Debugf("validating %v against %v", string(cert.RawSubject), distributionPoint)
+		crlBytes, err := cv.downloadCrl(distributionPoint)
```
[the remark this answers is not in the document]
sfc-gh-jkasten: Piotr's initial Slack message indicated that caching was not complete. Sorry for the noise.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+			if isSelfSigned {
+				logger.Debugf("certificate %v is self signed, assuming root CA", subject)
+				break
+			}
+			if certPos == len(verifiedChain)-1 {
+				// Is it correct to assume that the last certificate in the chain is the self signed?
+				return fmt.Errorf("expected last certificate to be self signed, but it's not. subject: %v, issuer: %v", subject, issuer)
+			}
+			issuerCert := verifiedChain[certPos+1]
+			if err := cv.verifyCertificate(cert, issuerCert); err != nil {
```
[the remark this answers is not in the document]
sfc-gh-jkasten: Thinking a bit more... there is some risk here. A CA can decide that it doesn't want to maintain a root or intermediate CA and drop official support for it. 
For example, maybe the root is expiring and they don't want to pay for the audit. All of their active chains have cross signs. 
Maybe the client hasn't updated their root store and has both roots. The server may still be sending the cross sign for maximum compatibility. If the CA doesn't continue provisioning the CRL endpoint, this code would fail.

Most cross-sign transitions have a few client casualties where this type of thing pops up.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+	return &crlValidator{
+		httpClient: httpClient,
+	}
+}
+
+// function to be set as custom TLS verification in the http client
+func (cv *crlValidator) verifyPeerCertificates(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	for _, verifiedChain := range verifiedChains {
+		for certPos, cert := range verifiedChain {
+			subject := string(cert.RawSubject)
```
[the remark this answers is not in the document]
sfc-gh-pfus: Good one! I took it from the existing code, but very good comments.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+			issuer := string(cert.RawIssuer)
+			logger.Debugf("started CRL validation for %v", subject)
+			isSelfSigned := subject == issuer
+			if isSelfSigned {
+				logger.Debugf("certificate %v is self signed, assuming root CA", subject)
+				break
+			}
+			if certPos == len(verifiedChain)-1 {
```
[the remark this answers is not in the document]
sfc-gh-pfus: I believe this is the difference between first parameter `rawCerts` is for - it contains raw cert response from TLS handshake. `verifiedChains` contains parsed certs, and from my observations - root issuer is added, even if retrieved cert chain does not have it.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+				// Is it correct to assume that the last certificate in the chain is the self signed?
```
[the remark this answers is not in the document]
sfc-gh-pfus: Ok, so I assume it is fine then.

crl.go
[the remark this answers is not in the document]
sfc-gh-pfus: Good catch! I fixed it that any valid chain causes the whole process to succeed.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+	crlBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	logger.Debugf("downloaded %v bytes for CRL %v", len(crlBytes), url)
+	return crlBytes, err
+}
+
+func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, cert *x509.Certificate, distributionPoint string) error {
+	for _, ext := range append(crl.Extensions, crl.ExtraExtensions...) {
```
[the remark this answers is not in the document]
sfc-gh-pfus: I prefer leaving it just to be on the safe side, if you don't mind.

crl.go
```diff
@@ -0,0 +1,119 @@
...
+		if ext.Id.Equal(idpOID) {
+			var idp issuingDistributionPoint
+			_, err := asn1.Unmarshal(ext.Value, &idp)
+			if err != nil {
+				return fmt.Errorf("failed to unmarshal IDP extension: %w", err)
+			}
+			if string(idp.DistributionPoint.FullName[0].Bytes) != distributionPoint {
```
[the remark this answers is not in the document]
sfc-gh-pfus: Ok, I fixed - I hope I understood it properly.

crl.go
```diff
@@ -0,0 +1,134 @@
...
+func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, cert *x509.Certificate, distributionPoint string) error {
+	for _, ext := range append(crl.Extensions, crl.ExtraExtensions...) {
+		if ext.Id.Equal(idpOID) {
+			var idp issuingDistributionPoint
+			_, err := asn1.Unmarshal(ext.Value, &idp)
+			if err != nil {
+				return fmt.Errorf("failed to unmarshal IDP extension: %w", err)
+			}
+			for _, dp := range idp.DistributionPoint.FullName {
+				if string(dp.Bytes) == distributionPoint {
```
sfc-gh-jkasten, line 122: I am worried that this conversion might not work.

In particular, if I read the spec correctly, this `fullName` is of type `GeneralNames`. `GeneralNames` is a sequence of `GeneralName` (which you are iterating over) and a `GeneralName` can be a number of types with a CHOICE [[ref](https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.5:~:text=Profile%20%20%20%20%20%20%20%20%20%20%20%20May%202008-,GeneralName%20%3A%3A%3D%20CHOICE%20%7B,-otherName%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%5B0%5D%20%20AnotherName)]

The only [type allowed for CRL Distribution points](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712112-crl-distribution-points) is the "URI" type, which means that it is guaranteed to be an IA5String (essentially ASCII), but I believe the tag [6]/header is/are still encoded in the bytes.

The tests indicate that you are currently only creating your own version here. It might be helpful to test a real CRL with the IDP. 
http://c.pki.goog/we2/yK5nPhtHKQs.crl is as small as I could quickly find.
You can `curl http://c.pki.goog/we2/yK5nPhtHKQs.crl | base64` and copy it over to [asn1parse](https://lapo.it/asn1js/) to get a slightly better look. (Note it is still too big for the UI to work extremely well on the website)

(`curl http://c.pki.goog/we2/yK5nPhtHKQs.crl | openssl asn1parse -inform der` wouldn't parse the extension for me)

I think either you need to do a bit more parsing to convert that GeneralName into the raw IA5String's value or convert the distributionPoint into a GeneralName. (Once you get the format, it is also possible to append the right prefix to the distribution point name if need be if you are missing the library calls to make it work cleanly.)

I know this is a pain. This [code](https://github.com/letsencrypt/boulder/blob/0d7ea60b2cb6fa3553920d4ad3d630dbea28e66e/linter/lints/cpcps/lint_crl_has_idp.go#L157) might help as a reference.

sfc-gh-pfus: Hmm, based on what you sent, it seems that I parse these bytes correctly. GeneralName is to be the URI type. That's what I do - I parse all bytes into string. I added the test with the CRL you mentioned and it is parsed correctly.
Btw - this ASN1 parser - nice one, very helpful!

sfc-gh-jkasten: Thanks for taking a look. There is some [magic](https://github.com/golang/go/issues/13999) in the asn1 golang package, so it may be just ignoring the tag? Anyway, we are guaranteed an IA5String so if it does work on the real world example, it will work consistently. Thanks for adding a test.

crl.go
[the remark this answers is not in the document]
sfc-gh-jkasten: A bit more info on the edge case here after thinking about crazy clients on the internet...

I know some clients will to pin to a particular intermediate against all guidance to the contrary. One way that you would could probably do this on the client is by replacing the trust store with just the intermediates you care about.

If that is the case, this code would fail. In some ways, you are helping the client from breaking upon an local update rather than waiting for the server to break them, :) but you might also get blamed for taking down their working system.

I might label this as overstepping the bounds of this library. x509.Verify() is responsible for giving you cryptographically verified chains to a trust anchor... what that trust anchor is outside the role of your CRL verification here? 

I would personally remove it. I suspect that "system roots" (the `nil` value) is being configured and it is probably impossible to know what the user has done within their environment. Non-zero risk and no expected return value?

crl.go
```diff
@@ -0,0 +1,134 @@
...
+			logger.Debugf("finished CRL validation for %v", cert.Subject)
+		}
+		if isValidChain {
+			return nil
+		}
+	}
+	return fmt.Errorf("no valid certificate chain found after CRL validation. errors: %w", allErrors)
+}
+
+func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x509.Certificate) error {
```
sfc-gh-jkasten, line 66: Note: This code does not verify that the CRL is fresh. In essence, the attacker can MiTM the outgoing HTTP request and replace it with a year old version of the CRL (e..g before the cert was even issued) and this code would succeed.

I originally thought this problem was going to be solved by checking that the returned value was strictly greater than the "cached" value and mostly YOLOing the initial cert fetch, but if you are truly trying to provide a "fail closed" mode and a cache isn't being built yet... it probably makes sense to return a failure if time.Now() is greater than the CRL's `nextUpdate` field.

sfc-gh-pfus: I'll add checking freshness and sequence numbers in the commit that implements disk caching.

sfc-gh-pbulawa (approved):

crl.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func newCrlValidator(httpClient http.Client) *crlValidator {
+	return &crlValidator{
+		httpClient: httpClient,
+	}
+}
+
+// function to be set as custom TLS verification in the http client
+func (cv *crlValidator) verifyPeerCertificates(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
sfc-gh-dheyman, line 34: do we need this `rawCerts` argument if it doesn't get used anywhere? This is interface implementation?

crl.go
```diff
@@ -0,0 +1,134 @@
...
+	var allErrors error = errors.New("")
```
sfc-gh-dheyman: no need to declare "error" type

crl.go
```diff
@@ -0,0 +1,134 @@
...
+			isSelfSigned := bytes.Equal(cert.RawSubject, cert.RawIssuer)
+			if isSelfSigned {
+				logger.Debugf("certificate %v is self signed, assuming root CA", cert.Subject)
+				break
+			}
+			if certPos == len(verifiedChain)-1 {
+				// Is it correct to assume that the last certificate in the chain is the self signed?
+				err := fmt.Errorf("expected last certificate to be self signed, but it's not. subject: %v, issuer: %v", cert.Subject, cert.Issuer)
+				logger.Debug(err)
+				allErrors = fmt.Errorf("%w%w", allErrors, err)
```
sfc-gh-dheyman: `%w\n%w`?

crl.go
sfc-gh-jkasten: OK - just note that as of now this is vulnerable. This defense is in the same line as why the Issuing Distribution Point extension is checked. Make sure an attacker isn't swapping out another CRL with a valid signature.

sfc-gh-jkasten (approved):

crl.go
```diff
@@ -0,0 +1,134 @@
...
+				isValidChain = false
+				break
+			}
+			logger.Debugf("finished CRL validation for %v", cert.Subject)
+		}
+		if isValidChain {
+			return nil
+		}
+	}
+	return fmt.Errorf("no valid certificate chain found after CRL validation. errors: %w", allErrors)
```
sfc-gh-dheyman, line 63: we'll add support for failOpen in the later PRs?

crl.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		logger.Debugf("parsed CRL for %v, number of revoked certificates: %v", distributionPoint, len(crl.RevokedCertificateEntries))
+		if err = crl.CheckSignatureFrom(issuerCert); err != nil {
+			return fmt.Errorf("signature verification error for CRL %v: %w", distributionPoint, err)
+		}
+		if !bytes.Equal(crl.RawIssuer, cert.RawIssuer) {
+			logger.Debugf("failed to verify CRL issuer, got: %v, expected: %v", crl.Issuer, cert.Issuer)
+			return fmt.Errorf("failed to verify CRL issuer")
+		}
+		if err = cv.verifyAgainstIdpExtension(crl, distributionPoint); err != nil {
```
sfc-gh-dheyman, line 85: what is the purpose of this check?

sfc-gh-jkasten: An attacker can substitute out a CRL partition that the certificate isn't actually included under.
e.g.
The issuer "Intermediate CA - RSA" issues certs that are contained in the following CRL partitions ["crl.pki/1", and "crl.pki/2"], but both CRLs are signed by the same key and have the same issuer. CAs do this to limit the size of their CRLs.

An attacker who has access to a revoked certificate and is intercepting a connection to "example.com" with a cert containing the crl.pki/1 CRL distribution point can just MITM the connection for crl.pki/1 and replace it with crl.pki/2. 

If this check doesn't exist then the code will see that it is a valid CRL issued by "Intermediate CA - RSA" and that the serial number of the cert is not in that CRL. It will mistakenly think the certificate is not revoked.

crl.go
sfc-gh-pfus: As of now - it isn't, because in this piece of code we don't store CRL, we download a new one for every verification :) But yes, it will be important while implementing disk caching.

crl.go
sfc-gh-pfus: Yes, it is - it is see https://pkg.go.dev/crypto/tls and search for `verifyPeerCertificate`.

crl.go
sfc-gh-pfus: It's more common to have all errors joined in one line.

crl.go
sfc-gh-pfus: Yes, I even have it ready, let's merge this PR ASAP.

crl.go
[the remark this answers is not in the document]
sfc-gh-pfus: Ok, your explanation convinces me.

PR #1431, opened 2025-05-30
SNOW-1902248: Add Azure WIF authentication support

sfc-gh-dheyman:
### Description

SNOW-1902248

### Checklist

### DIFF
--- a/auth_wif.go
+++ b/auth_wif.go
@@ -15,6 +15,7 @@ import (
 	"github.com/golang-jwt/jwt/v5"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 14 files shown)


auth_wif.go
```diff
@@ -281,10 +291,12 @@ func (c *gcpIdentityAttestationCreator) createAttestation() (*wifAttestation, er
 	}
 	sub, iss, err := extractSubIssWithoutVerifyingSignature(token)
 	if err != nil {
-		return nil, fmt.Errorf("could not extract claims from token: %v", err.Error())
+		logger.Errorf("could not extract claims from token: %v", err.Error())
+		return nil, nil
```
sfc-gh-pfus, line 295: I'm still feeling unease about returning nil in case of errors. Is it again because of autodetect? Maybe we can pass a flag during struct construction if it is autodetect or not, and in case of normal mode return proper error?

auth_wif.go
```diff
@@ -375,3 +387,162 @@ func (c *oidcIdentityAttestationCreator) createAttestation() (*wifAttestation, e
...
+			a.azureAttestationMetadataProvider.clientID(),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create Azure Functions identity request: %v", err)
+		}
+	}
+
+	tokenJSON := fetchTokenFromMetadataService(request, a.cfg)
+	if tokenJSON == "" {
+		logger.Debug("Could not fetch Azure token.")
```
sfc-gh-pfus, line 455: Is it a debug only? No error describing what happened?

auth_wif.go
```diff
@@ -375,3 +387,162 @@ func (c *oidcIdentityAttestationCreator) createAttestation() (*wifAttestation, e
...
+	}
+
+	return &wifAttestation{
+		ProviderType: string(azureWif),
+		Credential:   token,
+		Metadata:     map[string]string{"sub": sub, "iss": iss},
+	}, nil
+}
+
+func determineEntraResource(config *Config) string {
```
sfc-gh-pfus, line 499: Is it azure specific? Can we embed it into azure struct?

sfc-gh-pbulawa (approved):

auth_wif.go
sfc-gh-dheyman: this is used for the purpose of azure struct constructor so we can't really. We can embed it in Config, but I'd rather leave WIF logic in auth_wif.go

auth_wif.go
sfc-gh-dheyman: Yes, imho no point in handling errors here since we won't do anything beside ending the flow (which might be expected in autodetect). The error does get printed in the called method though.

auth_wif.go
sfc-gh-pfus: So this is common in go to have "factory methods" like `new<typeName>`. Maybe we can have sth like:
```
func newX(cfg *Config) {
  x := &x{
    // set remaining fields
  }
  x.entraResource = x.DetermineEntraResource(cfg)
  return x
}
```
WDYT?

auth_wif.go
sfc-gh-dheyman: I don't think there is much sense in it. The error will get printed eventually in the upper layers if autodetect/specified provider fails. If there is a neccessity to check the details debug mode will help. I know it's not ideal, but introduction of another boolean flag just for the sake of logs is imo not a great solution either.

auth_wif.go
sfc-gh-dheyman: If I'm being honest, I don't see any value in this. I'd rather have a constructor that ensures struct coherence

sfc-gh-pfus: A matter of preference. My constructor would ensure that the object is properly created.

PR #1433, opened 2025-06-04
SNOW-625282: Make logrus dependency optional or drop it all together

rickr-sigma:
### Description

Resolved version of https://github.com/snowflakedb/gosnowflake/pull/1286 which was accidentally closed.

The following changes decouple logrs from SFLogger, with the eventual goal to allow per context logging control (in a followup PR) by a user provided SFLogger.

One thing to be aware of is that certain interfaces are no longer supported in particular the functional variants of logging, and the WithTime, WithFields, WithError, Panic and Fatal which were all ONLY being exercised in unit tests.

### Checklist


rickr-sigma:
@sfc-gh-pfus,  @frbvianna sorry accidentally closed https://github.com/snowflakedb/gosnowflake/pull/1286.  Here's the same PR sync'd against latest.

rickr-sigma:
ping @sfc-gh-pfus , @sfc-gh-dheyman ... making sure this is still on your radar, thanks.

sfc-gh-pfus:
It is, sorry to keep you waiting!

ctx_test.go
```diff
@@ -36,15 +38,28 @@ func TestCtxVal(t *testing.T) {
...
-	fs2 := context2Fields(ctx2)
-	l1 := log.WithFields(*fs1)
-	l2 := log.WithFields(*fs2)
-	l1.Info("Hello 1")
-	l2.Warning("Hello 2")
+	var b bytes.Buffer
+	log.SetOutput(&b)
+	err := log.SetLogLevel("trace")
+	if err != nil {
+		t.Error("could not set log level")
```
sfc-gh-pfus: Can you use our assertions (i.e. `assertNilF` in this case)? I assume it should be fatal, not error, there's no point in continuing the test if log level is not set?

ctx_test.go
```diff
@@ -36,15 +38,28 @@ func TestCtxVal(t *testing.T) {
...
 
-	fs1 := context2Fields(ctx1)
-	fs2 := context2Fields(ctx2)
-	l1 := log.WithFields(*fs1)
-	l2 := log.WithFields(*fs2)
-	l1.Info("Hello 1")
-	l2.Warning("Hello 2")
+	var b bytes.Buffer
+	log.SetOutput(&b)
+	err := log.SetLogLevel("trace")
```
sfc-gh-pfus: Correct me if I'm wrong, but we are using a default `logger`, so we modify the properties of a global variable. I don't think this is the best idea, because all tests start spamming with trace to... bytes.Buffer ;) I think it may be better to create a dedicated instance for this test.

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
 // logger.WithContext is used
 var LogKeys = [...]contextKey{SFSessionIDKey, SFSessionUserKey}
 
-// SFLogger Snowflake logger interface to expose FieldLogger defined in logrus
-// Deprecated: will be reorganized in the future releases.
+// Fields
+type Fields map[string]any
+
+// ConvertibleEntry returns the underlying logrus Entry struct.
```
sfc-gh-pfus: Let's not expose anything logrus related. I'd like to get rid of it.

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+type ConvertibleEntry interface {
+	ToEntry() *rlog.Entry
+}
+
+// LogEntry allows for logging using a snapshot of field values, similar to logrus.Entry.
```
sfc-gh-pfus: Let's not mention logrus :)

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+// No references to logrus or other implementat specific logging should be placed into this interface.
```
sfc-gh-pfus: implementat - typo

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	Debug(args ...interface{})
+	Info(args ...interface{})
+	Print(args ...interface{})
+	Warn(args ...interface{})
+	Warning(args ...interface{})
+	Error(args ...interface{})
+	Fatal(args ...interface{})
+	Panic(args ...interface{})
+
+	Traceln(args ...interface{})
```
sfc-gh-pfus, line 68: Do we need this? What is the purpose of this vs `Trace`?

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
-	rlog.Ext1FieldLogger
+	LogEntry
+	WithField(key string, value interface{}) LogEntry
+	WithFields(fields Fields) LogEntry
+	WithError(err error) LogEntry
+
 	SetLogLevel(level string) error
 	GetLogLevel() string
-	WithContext(ctx context.Context) *rlog.Entry
+	WithContext(ctx ...context.Context) LogEntry
```
sfc-gh-pfus: Why do you think we need multiple contexts here?

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+
+// LogEntry allows for logging using a snapshot of field values, similar to logrus.Entry.
+// No references to logrus or other implementat specific logging should be placed into this interface.
+type LogEntry interface {
+	Tracef(format string, args ...interface{})
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Printf(format string, args ...interface{})
+	Warnf(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
```
sfc-gh-pfus, line 53: I think we could get rid of this duplicate.

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	ToEntry() *rlog.Entry
+}
+
+// LogEntry allows for logging using a snapshot of field values, similar to logrus.Entry.
+// No references to logrus or other implementat specific logging should be placed into this interface.
+type LogEntry interface {
+	Tracef(format string, args ...interface{})
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Printf(format string, args ...interface{})
```
sfc-gh-pfus, line 51: This one is used only in tests.

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	Warnf(format string, args ...interface{})
+	Warningf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+	Panicf(format string, args ...interface{})
```
sfc-gh-pfus, line 56: I'm 50/50 on this. Do you see a value in it?

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	Panicln(args ...interface{})
+}
+
+// SFLogger Snowflake logger interface which abstracts away the underlying logging mechanism.
+// No references to logrus or other implementation specific logging should be placed into this interface.
 type SFLogger interface {
-	rlog.Ext1FieldLogger
+	LogEntry
+	WithField(key string, value interface{}) LogEntry
+	WithFields(fields Fields) LogEntry
```
sfc-gh-pfus, line 84: I think that having a dedicated type for `Fields` is like shooting a sparrow with a cannon :)

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+	WithError(err error) LogEntry
```
sfc-gh-pfus: Do we need this? Isn't it a logrus leftover?

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
...
+
 	SetLogLevel(level string) error
 	GetLogLevel() string
-	WithContext(ctx context.Context) *rlog.Entry
+	WithContext(ctx ...context.Context) LogEntry
 	SetOutput(output io.Writer)
 	CloseFileOnLoggerReplace(file *os.File) error
```
sfc-gh-pfus: I think we can unexport this and `Replace`, right?

log.go
```diff
@@ -57,6 +103,8 @@ func SFCallerPrettyfier(frame *runtime.Frame) (string, string) {
 	return path.Base(frame.Function), fmt.Sprintf("%s:%d", path.Base(frame.File), frame.Line)
 }
 
+var _ SFLogger = &defaultLogger{} // ensure defaultLogger isa SFLogger.
```
sfc-gh-pfus: Typo.

log.go
```diff
@@ -39,13 +39,59 @@ func RegisterLogContextHook(contextKey string, ctxExtractor ClientLogContextHook
 // logger.WithContext is used
 var LogKeys = [...]contextKey{SFSessionIDKey, SFSessionUserKey}
```
sfc-gh-pfus, line 40: It would be great to incorporate it to some interface (maybe not `SFLogger`, but something else) and get rid of the global variable. Same for `clientLogContextHooks`.

log.go
rickr-sigma: WithError and WithTime removed along with tests.

log.go
rickr-sigma: CloseFileOnLoggerReplace() and Replace() no longer on interface.

log.go
rickr-sigma: so far I've resisted touching any logging callsites, thinking we would do that cleanup (if at all) once we put an interface in place.

log.go
rickr-sigma: fixed.

log.go
rickr-sigma: +1, again I'd like to clean this up after this PR.

log.go
rickr-sigma: +1, again I'd like to clean this up after this PR.

log.go
rickr-sigma: +1, again I'd like to clean this up after this PR.  I'd prefer a single variant of all these calls, less burden on the SFLogger implementor and cleaner interface.  Its pretty straightforward to get rid of the ln() variants since they have less callers.  

To consolidate further will be slightly harder as there are lots of callers to say Info() and Infof().  I could see an advantage to keeping Infof() since an implementation can avoid the formatting cost if the level is high enough (arguments would still be eval'd though).

The non-stringf versions can then just be removed, since they are emitting a string w/o args (at least I think that is the intention and what most callers are doing).  That said, I believe some callsites are actually broken...although the contract is not clear from the interface.

log.go
rickr-sigma: I think we can change this once the interface lands.  I'd like to see WithField(s) removed from the interface then we can add a ```WithContextKeys() SFLogger``` that replaces ```clientLogContextHooks``` logic but we would no longer support the callback.   Instead the client would look something like this ; 

```
logger = CreateDefaultLogger().WithContextKeys(key1, key2);
...
logger.WithContext(ctx) 
```

connection_util.go
```diff
@@ -34,7 +33,7 @@ func (sc *snowflakeConn) getClientSessionKeepAliveHeartbeatFrequency() (time.Dur
 
 	num, err := strconv.Atoi(*v)
 	if err != nil {
-		logger.WithError(err).Warnf("Failed to parse client session keepalive heartbeat frequency. Falling back to default.")
+		logger.WithField("error", err).Warnf("Failed to parse client session keepalive heartbeat frequency. Falling back to default.")
```
sfc-gh-pfus: I don't see a value in this field, maybe it's worth just logging it?

easy_logging.go
```diff
@@ -96,11 +96,14 @@ func reconfigureEasyLogging(logLevel string, logPath string) error {
...
 	}
 	newLogger.SetOutput(output)
-	err = newLogger.CloseFileOnLoggerReplace(file)
+	dl := newLogger.(*defaultLogger)
+	err = dl.closeFileOnLoggerReplace(file)
 	if err != nil {
 		logger.Errorf("%s", err)
 	}
-	logger.Replace(&newLogger)
+	if currentLogger, ok := logger.(*defaultLogger); ok {
```
sfc-gh-pfus, line 108: Maybe we should check this condition at the beginning of the function? Otherwise, there's no point in the whole creating new logger, closing file and so on.

connection_util.go
rickr-sigma: fixed.

easy_logging.go
rickr-sigma: Added a bail-early condition at the beginning but chose to keep the same silent-do-nothing behavior if reconfigure is not possible.  A follow-up PR can change this semantic if needed.

connection_util.go
```diff
@@ -34,7 +33,7 @@ func (sc *snowflakeConn) getClientSessionKeepAliveHeartbeatFrequency() (time.Dur
 
 	num, err := strconv.Atoi(*v)
 	if err != nil {
-		logger.WithError(err).Warnf("Failed to parse client session keepalive heartbeat frequency. Falling back to default.")
+		logger.Warnf("Failed to parse client session keepalive heartbeat frequency: %v. Falling back to default.", err)
```
sfc-gh-boler, line 37: Any particular reason for this change?

easy_logging.go
```diff
@@ -84,6 +84,9 @@ func easyLoggingInitError(err error) error {
 }
 
 func reconfigureEasyLogging(logLevel string, logPath string) error {
+	if _, ok := logger.(*defaultLogger); !ok {
+		return nil // cannot replace custom logger
```
sfc-gh-boler, line 89: Shouldn't this return an error?

connection_util.go
sfc-gh-pfus: It was my ask - we don't need `WithError` anymore, it was a leftover from logrus.

easy_logging.go
sfc-gh-pfus: No, it's fine, we shouldn't configure a new logger automatically from easy logging, if customized logger is used.

sfc-gh-boler: Got it. Can we add a comment explaining why replacing the logger is not allowed?

rickr-sigma:
> Thanks for the contribution. In general this is OK, but mind that we will be releasing a new go driver in the next few quarters and the logging infra will change there once again.
> 
Not a problem.  The logging infra can change but what will be important is to create (and maintain) a compatible interface and means of injecting a logger that clients can rely on.   

Our particular use-case is to be able to dynamically control the logger / logging level  from the client code.  For example, my code may be servicing a request for which I want debug log level and once it's complete I want to switch to warn level.   If all the driver code goes through the interface then this can be more easily accomplished.  

> Please update the PR description -- currently it says that the PR is incomplete.

Done,

PR #1435, opened 2025-06-05
SNOW-2117152 Add fail open mode to CRL implementation

sfc-gh-pfus:
### Description

### Checklist

### DIFF
--- a/crl.go
+++ b/crl.go
@@ -21,11 +21,13 @@ type issuingDistributionPoint struct {
 }
 
 type crlValidator struct {
+	failClosed bool
 	httpClient http.Client
 }
 
-func newCrlValidator(httpClient http.Client) *crlValidator {
+func newCrlValidator(failClosed bool, httpClient http.Client) *crlValidator {
 	return &crlValidator{
+		failClosed: failClosed,
 		httpClient: httpClient,
 	}
 }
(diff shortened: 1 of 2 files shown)


crl.go
```diff
@@ -64,26 +61,51 @@ func (cv *crlValidator) verifyPeerCertificates(rawCerts [][]byte, verifiedChains
 }
 
 func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x509.Certificate) error {
+	if len(cert.CRLDistributionPoints) == 0 {
+		if cv.failClosed {
```
sfc-gh-dheyman, line 65: can't we just return errors and in the function that call this one perform failClosed/failOpen check once?

crl.go
```diff
@@ -37,14 +39,9 @@ func (cv *crlValidator) verifyPeerCertificates(rawCerts [][]byte, verifiedChains
 		isValidChain := true
 		for certPos, cert := range verifiedChain {
 			logger.Debugf("started CRL validation for %v", cert.Subject)
-			isSelfSigned := bytes.Equal(cert.RawSubject, cert.RawIssuer)
```
sfc-gh-dheyman, line 40: why do we remove this? is this assumption wrong?

sfc-gh-pfus: Yes. A customer may add any certificate to their truststore, it doesn't have to be self-signed.

crl.go
sfc-gh-pfus: Some errors should break failOpen, some shouldn't. But I have an idea how to it.

sfc-gh-pfus: No, unfortunately my idea was wrong. We can't as in many cases we don't want to return immediately in fail open, but continue the function.

PR #1436, opened 2025-06-05
Fix unprotected concurrent access to slice

obs-gh-felixbowing:
Access to the slices length must be protected by the mutex, but isn't. Running a test with the race detector would show the problem clearly.

Background: We are facing a large amount of memory usage that stems from the error telemetry framework of gosnowflake. It could be a leak, but it is unclear what it exactly is. While digging through the code, we found this potential issue. Most likely not
the cause of the memory growth, but good to fix in any case.

### DIFF
--- a/telemetry.go
+++ b/telemetry.go
@@ -53,8 +53,9 @@ func (st *snowflakeTelemetry) addLog(data *telemetryData) error {
 	}
 	st.mutex.Lock()
 	st.logs = append(st.logs, data)
+	shouldFlush := len(st.logs) >= st.flushSize
 	st.mutex.Unlock()
-	if len(st.logs) >= st.flushSize {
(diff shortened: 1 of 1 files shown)


sfc-gh-pfus (approved):

sfc-gh-pfus:
@obs-gh-felixbowing thanks for the PR! You are obviously right with the fix.
Side note: we are running [tests](https://github.com/snowflakedb/gosnowflake/blob/bea1b0ce5807c79b3384b43be6abe3e0847f7f2c/ci/scripts/execute_tests.sh#L18-L20) with race detection, but we didn't find this. Maybe our tests are too focused on functional correctness and not on real-life usage. To be considered, thanks for pointing this out!

PR #1437, opened 2025-06-06
SNOW-2117152 Add CRL disk caching

sfc-gh-pfus:
### Description

### Checklist

### DIFF
--- a/crl.go
+++ b/crl.go
@@ -81,10 +81,11 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
 	for _, distributionPoint := range cert.CRLDistributionPoints {
 		crl, err := cv.getCrlForDistributionPoint(cert, issuerCert, distributionPoint)
 		if err != nil {
+			err = fmt.Errorf("failed to get CRL for %v: %w", distributionPoint, err)
 			if cv.failClosed {
-				return fmt.Errorf("failed to get CRL for %v: %w", distributionPoint, err)
+				return err
 			}
-			logger.Warnf("failed to get CRL for %v: %v", distributionPoint, err)
+			logger.Warn(err)
 			continue
 		}
 		for _, rce := range crl.RevokedCertificateEntries {
@@ -229,12 +230,13 @@ func (cv *crlValidator) saveCrlToDisk(distributionPoint string, crl *x509.Revoca
 func (cv *crlValidator) getCrlFromDisk(distributionPoint string) ([]byte, *time.Time, error) {
 	file, err := os.OpenFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), os.O_RDONLY, 0644)
 	if errors.Is(err, os.ErrNotExist) {
-		logger.Debugf("CRL cache file %v does not exist, will download it", distributionPoint)
-		return nil, nil, nil // CRL not found in cache, will download it
+		logger.Debugf("CRL cache file %v does not exist", distributionPoint)
+		return nil, nil, nil
 	}
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to open CRL cache file %v: %w", distributionPoint, err)
 	}
+	defer file.Close()
 	crlBytes, err := io.ReadAll(file)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to read CRL cache file %v: %w", distributionPoint, err)


crl.go
```diff
@@ -69,52 +79,105 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
 		return nil
 	}
 	for _, distributionPoint := range cert.CRLDistributionPoints {
-		logger.Debugf("validating %v against %v", cert.Subject, distributionPoint)
-		crlBytes, err := cv.downloadCrl(distributionPoint)
+		crl, err := cv.getCrlForDistributionPoint(cert, issuerCert, distributionPoint)
 		if err != nil {
 			if cv.failClosed {
-				return fmt.Errorf("failed to download CRL from %v: %w", distributionPoint, err)
+				return fmt.Errorf("failed to get CRL for %v: %w", distributionPoint, err)
```
sfc-gh-dheyman: these interpolated string can be created once and reused. Same in other places.

crl.go
```diff
@@ -69,52 +79,105 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+		if err != nil {
+			logger.Debugf("failed to parse CRL for %v from disk cache: %v", distributionPoint, err)
 		}
-		if !bytes.Equal(crl.RawIssuer, cert.RawIssuer) {
-			logger.Debugf("failed to verify CRL issuer, got: %v, expected: %v", crl.Issuer, cert.Issuer)
-			if cv.failClosed {
-				return fmt.Errorf("failed to verify CRL issuer")
+	}
+	// If the CRL is not found in the disk cache or is outdated, download it
+	if downloadTime == nil || downloadTime.Add(cv.diskCacheTtl).Before(time.Now()) || (crl != nil && crl.NextUpdate.Before(time.Now())) {
```
sfc-gh-dheyman: can't we extract the caching logic in separate functions? This function got quite big and I think it can be made easier to understand/maintain by doing this.

crl.go
```diff
@@ -69,52 +79,105 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+		if err != nil {
+			logger.Debugf("failed to parse CRL for %v from disk cache: %v", distributionPoint, err)
 		}
-		if !bytes.Equal(crl.RawIssuer, cert.RawIssuer) {
-			logger.Debugf("failed to verify CRL issuer, got: %v, expected: %v", crl.Issuer, cert.Issuer)
-			if cv.failClosed {
-				return fmt.Errorf("failed to verify CRL issuer")
+	}
+	// If the CRL is not found in the disk cache or is outdated, download it
+	if downloadTime == nil || downloadTime.Add(cv.diskCacheTTL).Before(time.Now()) || (crl != nil && crl.NextUpdate.Before(time.Now())) {
```
sfc-gh-dheyman, line 116: don't we want to add some delta to NextUpdate?

crl.go
```diff
@@ -154,3 +217,34 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
 	}
 	return nil
 }
+
+func (cv *crlValidator) cacheFileName(distributionPoint string) string {
```
sfc-gh-dheyman, line 222: what would be that string? We would have to make this compatibile with other drivers?

crl.go
```diff
@@ -154,3 +217,34 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
...
+}
+
+func (cv *crlValidator) saveCrlToDisk(distributionPoint string, crl *x509.RevocationList) error {
+	return os.WriteFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), crl.Raw, 0644)
+}
+
+func (cv *crlValidator) getCrlFromDisk(distributionPoint string) ([]byte, *time.Time, error) {
+	file, err := os.OpenFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), os.O_RDONLY, 0644)
+	if errors.Is(err, os.ErrNotExist) {
+		logger.Debugf("CRL cache file %v does not exist, will download it", distributionPoint)
```
sfc-gh-dheyman: I'd avoid logs that say that something "will" be done. "does not exists" is fine, but I'd add according log where we actually download.

crl.go
```diff
@@ -154,3 +217,34 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
...
+		return nil, nil, nil // CRL not found in cache, will download it
+	}
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to open CRL cache file %v: %w", distributionPoint, err)
+	}
+	crlBytes, err := io.ReadAll(file)
```
sfc-gh-dheyman, line 240: don't we have to `defer.close()`?

crl.go
```diff
@@ -154,3 +217,34 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
...
+func (cv *crlValidator) cacheFileName(distributionPoint string) string {
+	return base64.URLEncoding.EncodeToString([]byte(distributionPoint))
+}
+
+func (cv *crlValidator) saveCrlToDisk(distributionPoint string, crl *x509.RevocationList) error {
+	return os.WriteFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), crl.Raw, 0644)
+}
+
+func (cv *crlValidator) getCrlFromDisk(distributionPoint string) ([]byte, *time.Time, error) {
+	file, err := os.OpenFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), os.O_RDONLY, 0644)
```
sfc-gh-dheyman, line 231: do we verify file permission when the file already exists?

crl.go
sfc-gh-pfus: This was the only one that I fixed - in other cases debug log is reacher version of the error.

crl.go
sfc-gh-pfus: Why do you think it's needed?

crl.go
sfc-gh-pfus: According to the doc this is the distribution point - in plain English, the URL where the CRL is to be downloaded from, it should be compatible with other drivers, but this is not critical.

crl.go
sfc-gh-pfus: This is not a critical file - it is publicly accessible. There's no point in verifying permissions.

crl.go
sfc-gh-pfus: Good one. It is generated by copilot ;)

crl.go
sfc-gh-pfus: Good catch! Added.

sfc-gh-mhofman:
Can you put some description on how it works and what customer can configure as a location for the CRL cache? @sfc-gh-pfus

crl.go
sfc-gh-mhofman: According to the TM when others are able to tamper the file we need to verify the file, its signature etc? 
cc @sfc-gh-fochnik

crl.go
```diff
@@ -154,3 +218,35 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
 	}
 	return nil
 }
+
+func (cv *crlValidator) cacheFileName(distributionPoint string) string {
+	return base64.URLEncoding.EncodeToString([]byte(distributionPoint))
+}
+
+func (cv *crlValidator) saveCrlToDisk(distributionPoint string, crl *x509.RevocationList) error {
+	return os.WriteFile(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)), crl.Raw, 0644)
```
sfc-gh-mhofman, line 227: @sfc-gh-fochnik is 644 enough ?

crl.go
```diff
@@ -117,6 +98,89 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+				}
+			}
+		}
+	}
+	if crl == nil {
+		return nil, fmt.Errorf("failed to obtain CRL for %v", distributionPoint)
+	}
+
+	logger.Debugf("parsed CRL for %v, number of revoked certificates: %v", distributionPoint, len(crl.RevokedCertificateEntries))
+	if err = crl.CheckSignatureFrom(issuerCert); err != nil {
```
sfc-gh-fochnik, line 165: This should be done:
1. Before doing anything else with the CRL. 
2. Regardless if the CRL comes from the CA or local cache.

crl.go
sfc-gh-mhofman: The CA Security Council defines a CRL as “a digitally-signed file containing a list of certificates that have been revoked and have not yet expired (so I was wrong with previous comment)

crl.go
sfc-gh-fochnik: Yes.

crl.go
sfc-gh-fochnik: Don't we have an existing function that already verifies permissions?
Btw, isn't this equivalent to 
```
os.Open(filepath.Join(cv.cacheDir, cv.cacheFileName(distributionPoint)))
```
?

crl.go
```diff
@@ -117,6 +98,89 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+	if err != nil {
+		logger.Debugf("failed to read CRL %v from disk: %v", distributionPoint, err)
+	}
+	freshDownload := false
+	var crl *x509.RevocationList
+	if crlBytes != nil {
+		crl, err = x509.ParseRevocationList(crlBytes)
+		if err != nil {
+			logger.Debugf("failed to parse CRL for %v from disk cache: %v", distributionPoint, err)
+		}
```
sfc-gh-fochnik, line 113: You should check other condition that make the CRL invalid here. Signature, etc.

crl.go
```diff
@@ -117,6 +98,89 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+	}
+	// If the CRL is not found in the disk cache or is outdated, download it
+	if downloadTime == nil || downloadTime.Add(cv.diskCacheTTL).Before(time.Now()) || (crl != nil && crl.NextUpdate.Before(time.Now())) {
+		logger.Debugf("CRL for %v is outdated by disk TTL or missing, downloading", distributionPoint)
+		if crlBytes, err = cv.downloadCrl(distributionPoint); err != nil {
+			err = fmt.Errorf("failed to download CRL from %v: %w", distributionPoint, err)
+			if cv.failClosed { // if fail open, we can continue without fresh CRL
```
sfc-gh-fochnik, line 120: I wonder if we are making our lives more complicated than necessary by pushing this option all the way down.
Wouldn't it be clearer and less error-prone to do make the validator failure-mode-agnostic? Validation either passed, failed or errored out. With that result and failureMode you can then decide what to do. 
This behavior would be also easier to document.

crl.go
```diff
@@ -117,6 +98,89 @@ func (cv *crlValidator) verifyCertificate(cert *x509.Certificate, issuerCert *x5
...
+	if freshDownload {
+		newCrl, err := x509.ParseRevocationList(crlBytes)
+		if err != nil {
+			err = fmt.Errorf("failed to parse CRL for %v: %w", distributionPoint, err)
+			if cv.failClosed { // if fail open, we can continue without fresh CRL
+				return nil, err
+			}
+			logger.Debug(err)
+		} else {
+			if newCrl.NextUpdate.Before(time.Now()) {
```
sfc-gh-fochnik, line 137: You should check the signature before doing anything else with the CRL.

sfc-gh-fochnik (changes requested):

PR #1442, opened 2025-06-10
connectivity diagnosis - initial commit

sfc-gh-dszmolka:
### Description

Snowflake Python and JDBC drivers already have this capability; perform connectivity diagnostics based on allowlist file generated from `SELECT SYSTEM$ALLOWLIST` or `ALLOWLIST_PRIVATELINK`. 

This is an attempt to implement something similar to the Go driver as well.

#### Relevant configuration
* `ConnectionDiagnosticsEnabled` (default: false) - main flag to enable the diagnostics
* `ConnectionDiagnosticsAllowlistFile` - specify `/path/to/allowlist.json` to use a specific allowlist file which the driver should parse. If not specified, tne driver tries to open `allowlist.json` from the current directory. 
The `ConnectionDiagnosticsAllowlistFile` is only taken into consideration when `ConnectionDiagnosticsEnabled=true`
* `ConnectionDiagnosticsDownloadCRL` (default: false) - when enabled, the driver will try to attempt to download the CRL(s) it encounters during parsing the remote certificate chain; to verify connectivity to CRL endpoints as well.
The `ConnectionDiagnosticsDownloadCRL` is only taken into consideration when `ConnectionDiagnosticsEnabled=true`

#### Flow of operation when `ConnectionDiagnosticsEnabled=true`
* upon initial startup, driver opens and reads the `allowlist.json` to determine which hosts it needs to connect to, and then for each entry in the allowlist
* perform a DNS resolution test to see if the hostname is resolveable
  * driver logs an Error, when encountering a _public_ IP address for a host which looks to be a _private_ link hostname
* checks if proxy is used in the connection
* sets up a connection; for which we use the same transport which is driven by the driver's config (custom transport, or when OCSP disabled then OCSP-less transport, or by default, the OCSP-enabled transport)
* for HTTP endpoints, issues a HTTP GET request and see if it connects
* for HTTPS endpoints, the same , plus
  * verifies if HTTPS connectivity is set up correctly
  * parses the certificate chain and logs information on each certificate (on DEBUG loglevel, dump the whole cert)
  * if configured, also tries to connect to the CRL endpoints

### Checklist

### DIFF
--- a/connectivity_diagnosis.go
+++ b/connectivity_diagnosis.go
@@ -1,0 +1,347 @@
+package gosnowflake
+
+import (
+	"context"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+)
+
+type AllowlistEntry struct {
+	Host string `json:"host"`
+	Port int    `json:"port"`
+	Type string `json:"type"`
+}
+
+type Allowlist struct {
+	Entries []AllowlistEntry
+}
+
+var (
+	diagnosticTransport *http.Transport
+	client              *http.Client
+	// acceptable HTTP status codes for connectivity diagnosis
+	// for the sake of connectivity, e.g. 403 is perfectly fine
+	acceptableStatusCodes = []int{http.StatusOK, http.StatusForbidden}
+)
+
+// create a diagnostic client with the appropriate transport for the given config
+// by default SnowflakeTransport, if OCSP is disabled then snowflakeNoOcspTransport, if custom then that one
+func createDiagnosticClient(cfg *Config) *http.Client {
+	transport := createDiagnosticTransport(cfg)
+
+	clientTimeout := cfg.ClientTimeout
+	if clientTimeout == 0 {
+		clientTimeout = defaultClientTimeout // 900 seconds
+	}
+
+	return &http.Client{
+		Timeout:   clientTimeout,
+		Transport: transport,
+	}
+}
+
+// necessary to be able to log the IP address of the remote host to which we actually connected
+// might be even different from the result of DNS resolution
+func createDiagnosticDialContext() func(ctx context.Context, network, addr string) (net.Conn, error) {
+	dialer := &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+
+	return func(ctx context.Context, network, addr string) (net.Conn, error) {
+		conn, err := dialer.DialContext(ctx, network, addr)
+		if err != nil {
+			return nil, err
+		}
+
+		if remoteAddr := conn.RemoteAddr(); remoteAddr != nil {
+			remoteIPStr := remoteAddr.String()
+			// parse out just the IP (maybe port is present)
+			if host, _, err := net.SplitHostPort(remoteIPStr); err == nil {
+				remoteIPStr = host
+			}
+
+			// get hostname
+			hostname, _, _ := net.SplitHostPort(addr)
+			if hostname == "" {
+				hostname = addr
+			}
+
+			logger.Infof("[createDiagnosticDialContext] Connected to %s (remote IP: %s)", hostname, remoteIPStr)
+		}
+
+		return conn, nil
+	}
+}
+
+// enhance the transport with IP logging
+func createDiagnosticTransport(cfg *Config) *http.Transport {
+	// get transport based on OCSP / custom transport configuration
+	// default; SnowflakeTransport with OCSP validation
+	baseTransport := getTransport(cfg)
+
+	var httpTransport *http.Transport
+	if t, ok := baseTransport.(*http.Transport); ok {
+		httpTransport = t
+	} else {
+		httpTransport = SnowflakeTransport
+	}
+
+	// return a new transport enhanced with remote IP logging
+	// for SnowflakeNoOcspTransport, TLSClientConfig is nil
+	return &http.Transport{
+		TLSClientConfig: httpTransport.TLSClientConfig,
+		MaxIdleConns:    httpTransport.MaxIdleConns,
+		IdleConnTimeout: httpTransport.IdleConnTimeout,
+		Proxy:           httpTransport.Proxy,
+		DialContext:     createDiagnosticDialContext(),
+	}
+}
+
+func openAndReadAllowlistJSON(filePath string) (fileContent []byte, err error) {
+	if filePath == "" {
+		logger.Info("[openAndReadAllowlistJSON] allowlist.json location not specified, trying to load from current directory.")
+		filePath = "allowlist.json"
+	}
+	logger.Infof("[openAndReadAllowlistJSON] reading allowlist from %s.\n", filePath)
+	fileContent, err = os.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+	return fileContent, nil
+}
+
+func parseAllowlistJSON(fileContent []byte) (allowlist Allowlist, err error) {
+	logger.Debug("[parseAllowlistJSON] parsing allowlist.json")
+	err = json.Unmarshal(fileContent, &allowlist.Entries)
+	if err != nil {
+		return allowlist, err
+	}
+	return allowlist, nil
+}
+
+// look up the host, using the local resolver
+func resolveHostname(hostname string) {
+	ips, err := net.LookupIP(hostname)
+	if err != nil {
+		logger.Errorf("[resolveHostname] error resolving hostname %s: %s\n", hostname, err)
+		return
+	}
+	for _, ip := range ips {
+		logger.Infof("[resolveHostname] resolved hostname %s to %s\n", hostname, ip.String())
+		if isPrivateLinkHost(hostname) && !isPrivateIP(ip) {
+			logger.Errorf("[resolveHostname] this hostname should resolve to a private IP, but %s is public IP. Please, check your DNS configuration.\n", ip.String())
+		}
+	}
+}
+
+func isPrivateLinkHost(hostname string) bool {
+	return strings.HasSuffix(hostname, ".privatelink.snowflakecomputing.com")
+}
+
+func isPrivateIP(ip net.IP) bool {
+	return ip.IsPrivate()
+}
+
+func isAcceptableStatusCode(statusCode int, acceptableCodes []int) bool {
+	for _, code := range acceptableCodes {
+		if statusCode == code {
+			return true
+		}
+	}
+	return false
+}
+
+func fetchCRL(uri string) error {
+	logger.Infof("[fetchCRL] fetching  %s\n", uri)
+	resp, err := http.Get(uri)
+	if err != nil {
+		return fmt.Errorf("[fetchCRL] HTTP GET to %s endpoint failed: %w", uri, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("[fetchCRL] HTTP response status from endpoint: %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("[fetchCRL] failed to read response body: %w", err)
+	}
+	// TODO: do we want to persist CRL to disk ? (then cleanup, deal with read-only fs, etc)
+	logger.Infof("[fetchCRL] %s retrieved successfully (%d bytes), not persisting to disk.\n", uri, len(body))
+	return nil
+}
+
+func doHTTP(request *http.Request) error {
+	if request.URL.Host == "ocsp.snowflakecomputing.com" {
+		fullOCSPCacheURI := request.URL.String() + "/ocsp_response_cache.json"
+		newUrl, _ := url.Parse(fullOCSPCacheURI)
+		request.URL = newUrl
+	}
+	logger.Infof("[doHTTP] testing HTTP connection to %s\n", request.URL.String())
+	//resp, err := http.DefaultClient.Do(request)
+	resp, err := client.Do(request)
+	if err != nil {
+		return fmt.Errorf("HTTP GET to %s endpoint failed: %w", request.URL.String(), err)
+	}
+	defer resp.Body.Close()
+
+	if !isAcceptableStatusCode(resp.StatusCode, acceptableStatusCodes) {
+		return fmt.Errorf("HTTP response status from %s endpoint: %s", request.URL.String(), resp.Status)
+	}
+	logger.Infof("[doHTTP] Successfully connected to %s, HTTP response status: %s", request.URL.String(), resp.Status)
+	return nil
+}
+
+func doHTTPSGetCerts(request *http.Request, downloadCRLs bool) error {
+	logger.Infof("[doHTTPSGetCerts] connecting to %s\n", request.URL.String())
+	resp, err := client.Do(request)
+	if err != nil {
+		return fmt.Errorf("failed to connect: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if !isAcceptableStatusCode(resp.StatusCode, acceptableStatusCodes) {
+		return fmt.Errorf("HTTP response status from %s endpoint: %s", request.URL.String(), resp.Status)
+	}
+	logger.Infof("[doHTTPSGetCerts] Successfully connected to %s, HTTP response status: %s", request.URL.String(), resp.Status)
+
+	// get TLS connection state
+	logger.Debug("[doHTTPSGetCerts] getting TLS connection state")
+	tlsState := resp.TLS
+	if tlsState == nil {
+		return fmt.Errorf("no TLS connection state available")
+	}
+
+	// get cert chain
+	logger.Debug("[doHTTPSGetCerts] getting certificate chain")
+	certs := tlsState.PeerCertificates
+	logger.Infof("[doHTTPSGetCerts] Retrieved %d certificate(s).\n", len(certs))
+
+	// log individual cert details
+	for i, cert := range certs {
+		logger.Infof("[doHTTPSGetCerts] Certificate %d, serial number: %x\n", i+1, cert.SerialNumber)
+		logger.Infof("[doHTTPSGetCerts]   Subject: %s\n", cert.Subject)
+		logger.Infof("[doHTTPSGetCerts]   Issuer:  %s\n", cert.Issuer)
+		logger.Infof("[doHTTPSGetCerts]   Valid:   %s to %s\n", cert.NotBefore, cert.NotAfter)
+		logger.Infof("[doHTTPSGetCerts]   For further details, check https://crt.sh/?serial=%x (non-Snowflake site)\n", cert.SerialNumber)
+
+		// if cert has CRL endpoint, log them too
+		if len(cert.CRLDistributionPoints) > 0 {
+			logger.Infof("[doHTTPSGetCerts]   CRL Distribution Points:")
+			for _, dp := range cert.CRLDistributionPoints {
+				logger.Infof("[doHTTPSGetCerts]    - %s\n", dp)
+				// only try to download the actual CRL if configured to do so
+				if downloadCRLs {
+					if err := fetchCRL(dp); err != nil {
+						logger.Errorf("[doHTTPSGetCerts]      Failed to fetch CRL: %v\n", err)
+					}
+				}
+			}
+		} else {
+			logger.Infof("[doHTTPSGetCerts]   CRL Distribution Points not available")
+		}
+
+		// dump the full PEM data too on DEBUG loglevel
+		pemData := pem.EncodeToMemory(&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: cert.Raw,
+		})
+		logger.Debugf("[doHTTPSGetCerts]   certificate PEM:\n%s\n", string(pemData))
+	}
+	return nil
+}
+
+func createRequest(uri string) (*http.Request, error) {
+	logger.Infof("[createRequest] creating GET request to %s\n", uri)
+	req, err := http.NewRequest("GET", uri, nil)
+	if err != nil {
+		return nil, err
+	}
+	return req, nil
+}
+
+func checkProxy(req *http.Request) {
+	if diagnosticTransport == nil {
+		logger.Errorf("[checkProxy] diagnosticTransport is nil")
+		return
+	}
+	if diagnosticTransport.Proxy == nil {
+		// no proxy configured, nothing to log
+		return
+	}
+	p, err := diagnosticTransport.Proxy(req)
+	if p != nil {
+		logger.Infof("[checkProxy] PROXY detected in the connection: %v\n", p)
+	}
+	if err != nil {
+		logger.Errorf("[checkProxy] problem determining PROXY: %v\n", err)
+	}
+}
+
+func performDiagnosis(cfg *Config) {
+	allowlistFile := cfg.ConnectionDiagnosticsAllowlistFile
+	downloadCRLs := cfg.ConnectionDiagnosticsDownloadCRL
+
+	logger.Info("[performDiagnosis] starting connectivity diagnosis based on allowlist file.")
+	if downloadCRLs {
+		logger.Info("[performDiagnosis] CRLs will be attempted to be downloaded during https tests.")
+	}
+
+	// diagnostic client - its transport is based on the Config
+	// default: SnowflakeTransport with OCSP
+	client = createDiagnosticClient(cfg)
+	diagnosticTransport = client.Transport.(*http.Transport)
+
+	allowlistContent, err := openAndReadAllowlistJSON(allowlistFile)
+	if err != nil {
+		logger.Errorf("[performDiagnosis] error opening allowlist file: %v\n", err)
+		return
+	}
+	allowlist, err := parseAllowlistJSON(allowlistContent)
+	if err != nil {
+		logger.Errorf("[performDiagnosis] error parsing allowlist: %v\n", err)
+		return
+	}
+	for _, entry := range allowlist.Entries {
+		host := entry.Host
+		port := entry.Port
+		entryType := entry.Type
+		logger.Infof("[performDiagnosis] DNS check - resolving %s hostname %s\n", entryType, host)
+		resolveHostname(host)
+		if port == 80 {
+			logger.Infof("[performDiagnosis] HTTP check for %s %s\n", entryType, host)
+			req, err := createRequest(fmt.Sprintf("http://%s", host))
+			if err != nil {
+				logger.Errorf("[performDiagnosis] error creating request: %v\n", err)
+			}
+			checkProxy(req)
+			err = doHTTP(req)
+			if err != nil {
+				logger.Errorf("[performDiagnosis] error performing HTTP check: %v\n", err)
+				continue
+			}
+		} else if port == 443 {
+			logger.Infof("[performDiagnosis] HTTPS check - testing HTTPS connection to %s %s\n", entryType, host)
+			req, err := createRequest(fmt.Sprintf("https://%s", host))
+			if err != nil {
+				logger.Errorf("[performDiagnosis] error creating request: %v\n", err)
+			}
+			checkProxy(req)
+			err = doHTTPSGetCerts(req, downloadCRLs)
+			if err != nil {
+				logger.Errorf("[performDiagnosis] error performing HTTPS check: %v\n", err)
+				continue
+			}
+		}
+	}
+}
--- a/connectivity_diagnosis_test.go
+++ b/connectivity_diagnosis_test.go
@@ -1,0 +1,862 @@
+package gosnowflake
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+/*
+ * for the tests, we need to capture log output and assert on their content
+ * this is done by creating a fresh logger to log into a buffer and look at that buffer
+ * but we also need to preserve the original global logger to not modify that with tests
+ * and restore original logger after the tests
+ */
+func setupTestLogger() (buffer *bytes.Buffer, cleanup func()) {
+	originalLogger := logger
+	testLogger := CreateDefaultLogger() // from log.go
+	buffer = &bytes.Buffer{}
+	testLogger.SetOutput(buffer)
+	_ = testLogger.SetLogLevel("INFO")
+	logger = testLogger
+
+	cleanup = func() {
+		logger = originalLogger
+	}
+
+	return buffer, cleanup
+}
+
+func TestSetupTestLogger(t *testing.T) {
+	// save  original global logger
+	originalLogger := logger
+	// and restore it after test
+	defer func() { logger = originalLogger }()
+
+	t.Run("Setup Test Logger - buffer capture and cleanup", func(t *testing.T) {
+		buffer, cleanup := setupTestLogger()
+
+		assertNotNilE(t, buffer, "buffer should not be nil")
+		assertNotNilE(t, cleanup, "cleanup function should not be nil")
+
+		// the test message should be in the buffer
+		testMessage := "test log message for setupTestLogger"
+		logger.Info(testMessage)
+		logOutput := buffer.String()
+		assertStringContainsE(t, logOutput, testMessage, "buffer should capture log output")
+
+		// now cleanup
+		cleanup()
+		assertEqualE(t, logger, originalLogger, "cleanup should restore original logger")
+
+		// clear the buffer, log a new message into it
+		// logs should not go to the test logger anymore
+		buffer.Reset()
+		logger.Info("this should not appear in test buffer")
+		assertEqualE(t, buffer.String(), "", "buffer should be empty after cleanup")
+	})
+}
+
+// test case types
+type tcAllowlistEntry struct {
+	host      string
+	port      int
+	entryType string
+}
+
+type tcDiagnosticClient struct {
+	name            string
+	config          *Config
+	expectedTimeout time.Duration
+}
+
+type tcOpenAllowlistJSON struct {
+	name        string
+	setup       func() (string, func())
+	shouldError bool
+}
+
+type tcParseAllowlistJSON struct {
+	name           string
+	content        string
+	shouldError    bool
+	expectedLength int
+}
+
+type tcPrivateLinkHost struct {
+	hostname      string
+	isPrivateLink bool
+}
+
+type tcPrivateIP struct {
+	ip          string
+	isPrivateIP bool
+}
+
+type tcAcceptableStatusCode struct {
+	statusCode   int
+	isAcceptable bool
+}
+
+type tcFetchCRL struct {
+	name          string
+	setupServer   func() *httptest.Server
+	shouldError   bool
+	errorContains string
+}
+
+type tcCreateRequest struct {
+	name        string
+	uri         string
+	shouldError bool
+}
+
+type tcDoHTTP struct {
+	name          string
+	setupServer   func() *httptest.Server
+	setupRequest  func(serverURL string) *http.Request
+	shouldError   bool
+	errorContains string
+}
+
+type tcDoHTTPSGetCerts struct {
+	name          string
+	setupServer   func() *httptest.Server
+	downloadCRLs  bool
+	shouldError   bool
+	errorContains string
+}
+
+type tcResolveHostname struct {
+	name     string
+	hostname string
+}
+
+func TestAllowlistEntry(t *testing.T) {
+	testcases := []tcAllowlistEntry{
+		{"myaccount.snowflakecomputing.com", 443, "SNOWFLAKE_DEPLOYMENT"},
+		{"ocsp.snowflakecomputing.com", 80, "OCSP_CACHE"},
+	}
+
+	for _, tc := range testcases {
+		t.Run(fmt.Sprintf("%s_%d_%s", tc.host, tc.port, tc.entryType), func(t *testing.T) {
+			entry := AllowlistEntry{
+				Host: tc.host,
+				Port: tc.port,
+				Type: tc.entryType,
+			}
+
+			assertEqualE(t, entry.Host, tc.host, "host did not match")
+			assertEqualE(t, entry.Port, tc.port, "port did not match")
+			assertEqualE(t, entry.Type, tc.entryType, "type did not match")
+		})
+	}
+}
+
+func TestCreateDiagnosticClient(t *testing.T) {
+	testcases := []tcDiagnosticClient{
+		{
+			name: "Diagnostic Client with default timeout",
+			config: &Config{
+				ClientTimeout: 0,
+			},
+			expectedTimeout: defaultClientTimeout,
+		},
+		{
+			name: "Diagnostic Client with custom timeout",
+			config: &Config{
+				ClientTimeout: 60 * time.Second,
+			},
+			expectedTimeout: 60 * time.Second,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			client := createDiagnosticClient(tc.config)
+
+			assertNotNilE(t, client, "client should not be nil")
+			assertEqualE(t, client.Timeout, tc.expectedTimeout, "timeout did not match")
+			assertNotNilE(t, client.Transport, "transport should not be nil")
+		})
+	}
+}
+
+func TestCreateDiagnosticDialContext(t *testing.T) {
+	t.Run("Diagnostic DialContext creation", func(t *testing.T) {
+		dialContext := createDiagnosticDialContext()
+
+		assertNotNilE(t, dialContext, "dialContext should not be nil")
+
+		// new simple server to test basic connectivity
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		}))
+		defer server.Close()
+
+		u, _ := url.Parse(server.URL)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		conn, err := dialContext(ctx, "tcp", u.Host)
+		assertNilE(t, err, "error should be nil")
+		defer conn.Close()
+
+	})
+}
+
+func TestCreateDiagnosticTransport(t *testing.T) {
+	t.Run("Diagnostic Transport creation", func(t *testing.T) {
+		config := &Config{}
+		transport := createDiagnosticTransport(config)
+
+		assertNotNilE(t, transport, "transport should not be nil")
+		assertNotNilE(t, transport.DialContext, "dialContext should not be nil")
+
+		// by default we should use the SnowflakeTransport
+		assertEqualE(t, transport.TLSClientConfig, SnowflakeTransport.TLSClientConfig, "TLSClientConfig did not match with SnowflakeTransport default")
+		assertEqualE(t, transport.MaxIdleConns, SnowflakeTransport.MaxIdleConns, "MaxIdleConns did not match with SnowflakeTransport default")
+		assertEqualE(t, transport.IdleConnTimeout, SnowflakeTransport.IdleConnTimeout, "IdleConnTimeout did not match with SnowflakeTransport default")
+		if (transport.Proxy == nil) != (SnowflakeTransport.Proxy == nil) {
+			t.Errorf("Proxy function presence should match SnowflakeTransport default")
+		}
+	})
+}
+
+func TestOpenAndReadAllowlistJSON(t *testing.T) {
+	testcases := []tcOpenAllowlistJSON{
+		{
+			name: "Open and Read Allowlist - valid file path",
+			// create a temp allowlist file and then delete it
+			setup: func() (filePath string, cleanup func()) {
+				content := `[{"host":"myaccount.snowflakecomputing.com","port":443,"type":"SNOWFLAKE_DEPLOYMENT"}]`
+				tmpFile, _ := os.CreateTemp("", "allowlist_*.json")
+				_, _ = tmpFile.WriteString(content)
+				tmpFile.Close()
+
+				return tmpFile.Name(), func() { os.Remove(tmpFile.Name()) }
+			},
+			shouldError: false,
+		},
+		{
+			name: "Open and Read Allowlist - empty file path",
+			setup: func() (filePath string, cleanup func()) {
+				content := `[{"host":"myaccount.snowflakecomputing.com","port":443,"type":"SNOWFLAKE_DEPLOYMENT"}]`
+				_ = os.WriteFile("allowlist.json", []byte(content), 0644)
+
+				return "", func() { os.Remove("allowlist.json") }
+			},
+			shouldError: false,
+		},
+		{
+			name: "Open and Read Allowlist - non existent file",
+			setup: func() (filePath string, cleanup func()) {
+				return "/non/existent/file.json", func() {}
+			},
+			shouldError: true,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			filePath, cleanup := tc.setup()
+			defer cleanup()
+
+			content, err := openAndReadAllowlistJSON(filePath)
+
+			if tc.shouldError {
+				assertNotNilE(t, err, "error should not be nil")
+			} else {
+				assertNilE(t, err, "error should be nil")
+				assertNotNilE(t, content, "file content should not be nil")
+			}
+		})
+	}
+}
+
+func TestParseAllowlistJSON(t *testing.T) {
+	testcases := []tcParseAllowlistJSON{
+		{
+			name:           "Parse Allowlist - valid JSON",
+			content:        `[{"host":"myaccount.snowflakecomputing.com","port":443,"type":"SNOWFLAKE_DEPLOYMENT"}]`,
+			shouldError:    false,
+			expectedLength: 1,
+		},
+		{
+			name:           "Parse Allowlist - multiple entries",
+			content:        `[{"host":"myaccount.snowflakecomputing.com","port":443,"type":"SNOWFLAKE_DEPLOYMENT"},{"host":"ocsp.snowflakecomputing.com","port":80,"type":"OCSP_CACHE"}]`,
+			shouldError:    false,
+			expectedLength: 2,
+		},
+		{
+			name:           "Parse Allowlist - invalid JSON",
+			content:        `{a fully invalid json}`,
+			shouldError:    true,
+			expectedLength: 0,
+		},
+		{
+			name:           "Parse Allowlist - empty array",
+			content:        `[]`,
+			shouldError:    false,
+			expectedLength: 0,
+		},
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			allowlist, err := parseAllowlistJSON([]byte(tc.content))
+
+			if tc.shouldError {
+				assertNotNilE(t, err, "error should not be nil")
+			} else {
+				assertNilE(t, err, "error should be nil")
+			}
+
+			assertEqualE(t, len(allowlist.Entries), tc.expectedLength, "allowlist length did not match")
+		})
+	}
+}
+
+func TestIsPrivateLinkHost(t *testing.T) {
+	testcases := []tcPrivateLinkHost{
+		{"myaccount.eu-west-1.privatelink.snowflakecomputing.com", true},
+		{"myorg-myaccount.privatelink.snowflakecomputing.com", true},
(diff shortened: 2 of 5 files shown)


connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+type Allowlist struct {
+	Entries []AllowlistEntry
+}
+
+var (
+	diagnosticTransport *http.Transport
+	client              *http.Client
+	// acceptable HTTP status codes for connectivity diagnosis
+	// for the sake of connectivity, e.g. 403 is perfectly fine
+	acceptableStatusCodes = []int{http.StatusOK, http.StatusForbidden}
```
sfc-gh-dheyman: Would it be worth to broaden this range by checking for 2xx and 4xx? Or it might backfire...?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+}
+
+func openAndReadAllowlistJSON(filePath string) (fileContent []byte, err error) {
+	if filePath == "" {
+		logger.Info("[openAndReadAllowlistJSON] allowlist.json location not specified, trying to load from current directory.")
+		filePath = "allowlist.json"
+	}
+	logger.Infof("[openAndReadAllowlistJSON] reading allowlist from %s.\n", filePath)
+	fileContent, err = os.ReadFile(filePath)
```
sfc-gh-dheyman: is defer close file not needed?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+func resolveHostname(hostname string) {
+	ips, err := net.LookupIP(hostname)
+	if err != nil {
+		logger.Errorf("[resolveHostname] error resolving hostname %s: %s\n", hostname, err)
+		return
+	}
+	for _, ip := range ips {
+		logger.Infof("[resolveHostname] resolved hostname %s to %s\n", hostname, ip.String())
+		if isPrivateLinkHost(hostname) && !isPrivateIP(ip) {
+			logger.Errorf("[resolveHostname] this hostname should resolve to a private IP, but %s is public IP. Please, check your DNS configuration.\n", ip.String())
```
sfc-gh-dheyman: I think it would be useful if we could print the hostname in this log as well. I know it's in the other logs but just for the sake of visibility

connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("[fetchCRL] HTTP response status from endpoint: %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("[fetchCRL] failed to read response body: %w", err)
+	}
+	// TODO: do we want to persist CRL to disk ? (then cleanup, deal with read-only fs, etc)
```
sfc-gh-dheyman: imho not needed for the sake of diagnostics. I guess when cached, it could be used for CRL checks during actual, non-diagnostics workflows, but probs better to not complicate this code

sfc-gh-dheyman: We could potentially leverage the code written for the sake of CRLs here (still WIP though)

connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+		logger.Errorf("[performDiagnosis] error parsing allowlist: %v\n", err)
+		return
+	}
+	for _, entry := range allowlist.Entries {
+		host := entry.Host
+		port := entry.Port
+		entryType := entry.Type
+		logger.Infof("[performDiagnosis] DNS check - resolving %s hostname %s\n", entryType, host)
+		resolveHostname(host)
+		if port == 80 {
```
sfc-gh-dheyman: very similar logic inside this if`s branches. Perhaps extract to a separate function?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,349 @@
...
+	// get TLS connection state
+	logger.Debug("[doHTTPSGetCerts] getting TLS connection state")
+	tlsState := resp.TLS
+	if tlsState == nil {
+		return fmt.Errorf("no TLS connection state available")
+	}
+
+	// get cert chain
+	logger.Debug("[doHTTPSGetCerts] getting certificate chain")
+	certs := tlsState.PeerCertificates
```
sfc-gh-dheyman, line 257: this would be an already verified cert chain, correct?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,862 @@
...
+	// because it will be modified during test
+	originalClient := client
+	// after the test, restore the original value
+	defer func() { client = originalClient }()
+
+	testcases := []tcDoHTTPSGetCerts{
+		// simple disposable server with TLS to test basic connectivity
+		{
+			name: "Do HTTPS - successful https request",
+			setupServer: func() *httptest.Server {
```
sfc-gh-dheyman, line 489: we could've leverage wiremock server with proper mappings which we already use in the project, but this is also fine :)

connectivity_diagnosis.go
sfc-gh-dszmolka: the idea was behind using specific response codes; is that I know we can expect these for sure :) and they are normal. 
i'm concerned if we just consider every 2xx and especially every 4xx as acceptable by default, and they are not present on the Snowflake(related) infra today but might appear in the future accompanying an issue, then we might hide that

connectivity_diagnosis.go
sfc-gh-dszmolka: to my understanding it implicitly happens . as a file is not Open-ed before so we don't have a `File` pointer to think about, but `os.ReadFile` returns `[]byte` which doesn't seem to have a Close method.

connectivity_diagnosis.go
sfc-gh-dszmolka: good point!
edit: added logging

connectivity_diagnosis.go
sfc-gh-dszmolka: as it is WIP, i would leave this as-is for the time being

connectivity_diagnosis.go
sfc-gh-dszmolka: oops, very valid point. moved it to a separate function `performConnectivityCheck`

connectivity_diagnosis.go
sfc-gh-dszmolka: yes, as we're already successfully connected to the server at this point.  This behaviour (prior cert verification) still can be overridden, in the case when someone uses their own custom transport - but that's even without this change.

connectivity_diagnosis_test.go
sfc-gh-dszmolka: okay - for now, leaving it as-is, as it was simple and fulfilling the purpose of unit test

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+)
+
+// AllowlistEntry is a structure of allowlist entries
+type AllowlistEntry struct {
```
sfc-gh-pfus: Do we need this to be exported?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	Host string `json:"host"`
+	Port int    `json:"port"`
+	Type string `json:"type"`
+}
+
+// Allowlist is a list of allowlist entries; the allowlist itself
+type Allowlist struct {
```
sfc-gh-pfus: Same.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	Entries []AllowlistEntry
+}
+
+var (
+	diagnosticTransport *http.Transport
+	client              *http.Client
```
sfc-gh-pfus: What is the difference between diagnostic transport and client? Client should just use a transport. Do we need the transport itself?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	// acceptable HTTP status codes for connectivity diagnosis
+	// for the sake of connectivity, e.g. 403 is perfectly fine
+	acceptableStatusCodes = []int{http.StatusOK, http.StatusForbidden}
+)
+
+// create a diagnostic client with the appropriate transport for the given config
+// by default SnowflakeTransport, if OCSP is disabled then snowflakeNoOcspTransport, if custom then that one
```
sfc-gh-pfus: I wouldn't duplicate this sentence here, as for this function this logic is transparent. If we ever change the logic, it's very easy to forget to update this doc. And yes, we will change it while moving to CRLs.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+func createDiagnosticClient(cfg *Config) *http.Client {
+	transport := createDiagnosticTransport(cfg)
+
+	clientTimeout := cfg.ClientTimeout
+	if clientTimeout == 0 {
+		clientTimeout = defaultClientTimeout // 900 seconds
```
sfc-gh-pfus: Same for this comment.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+			logger.Infof("[createDiagnosticDialContext] Connected to %s (remote IP: %s)", hostname, remoteIPStr)
+		}
+
+		return conn, nil
+	}
+}
+
+// enhance the transport with IP logging
+func createDiagnosticTransport(cfg *Config) *http.Transport {
+	// get transport based on OCSP / custom transport configuration
```
sfc-gh-pfus: Is this comment needed?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	// default; SnowflakeTransport with OCSP validation
+	baseTransport := getTransport(cfg)
+
+	var httpTransport *http.Transport
+	if t, ok := baseTransport.(*http.Transport); ok {
```
sfc-gh-pfus, line 98: Why do we need such a check?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	return &http.Client{
+		Timeout:   clientTimeout,
+		Transport: transport,
+	}
+}
+
+// necessary to be able to log the IP address of the remote host to which we actually connected
+// might be even different from the result of DNS resolution
+func createDiagnosticDialContext() func(ctx context.Context, network, addr string) (net.Conn, error) {
+	dialer := &net.Dialer{
```
sfc-gh-pfus, line 59: Good that you repeated the same dialer that is used by a regular client, but maybe it is worth to have a function that creates it and use in all places? Or even one instance, as this is thread safe function? Just to not miss changes in one place only.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	if filePath == "" {
+		logger.Info("[openAndReadAllowlistJSON] allowlist.json location not specified, trying to load from current directory.")
+		filePath = "allowlist.json"
+	}
+	logger.Infof("[openAndReadAllowlistJSON] reading allowlist from %s.\n", filePath)
+	fileContent, err = os.ReadFile(filePath)
+	if err != nil {
+		return nil, err
+	}
+	return fileContent, nil
```
sfc-gh-pfus: Actually these last few lines can be replaced with just `return os.ReadFile(filePath)`.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+func parseAllowlistJSON(fileContent []byte) (allowlist Allowlist, err error) {
+	logger.Debug("[parseAllowlistJSON] parsing allowlist.json")
+	err = json.Unmarshal(fileContent, &allowlist.Entries)
```
sfc-gh-pfus, line 127: Same here. But why splitting into two functions? Maybe reading and parsing can be done together.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	return &http.Transport{
+		TLSClientConfig: httpTransport.TLSClientConfig,
+		MaxIdleConns:    httpTransport.MaxIdleConns,
+		IdleConnTimeout: httpTransport.IdleConnTimeout,
+		Proxy:           httpTransport.Proxy,
+		DialContext:     createDiagnosticDialContext(),
+	}
+}
+
+func openAndReadAllowlistJSON(filePath string) (fileContent []byte, err error) {
```
sfc-gh-pfus: It's not so popular in Go to use naked returns, especially that you don't use it - you still mentions vars during return.

connectivity_diagnosis.go
sfc-gh-pfus, line 127: If we unmarshal directly into an internal list, why do we need the wrapping struct at all? (Tricky question!)

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	err = json.Unmarshal(fileContent, &allowlist.Entries)
+	if err != nil {
+		return allowlist, err
+	}
+	return allowlist, nil
+}
+
+// look up the host, using the local resolver
+func resolveHostname(hostname string) {
+	ips, err := net.LookupIP(hostname)
```
sfc-gh-pfus, line 133: Is this the same that the driver does (or what is done under the hood)?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	for _, ip := range ips {
+		logger.Infof("[resolveHostname] resolved hostname %s to %s\n", hostname, ip.String())
+		if isPrivateLinkHost(hostname) && !isPrivateIP(ip) {
+			logger.Errorf("[resolveHostname] this hostname %s should resolve to a private IP, but %s is public IP. Please, check your DNS configuration.\n", hostname, ip.String())
+		}
+	}
+}
+
+func isPrivateLinkHost(hostname string) bool {
+	return strings.HasSuffix(hostname, ".privatelink.snowflakecomputing.com")
```
sfc-gh-pfus: It won't work for china deployments, right? It would be better to reuse driver function `isPrivateLink` (again - do the same behaviour that driver does).

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+func isPrivateIP(ip net.IP) bool {
```
sfc-gh-pfus: Is it needed? :)

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	Host string `json:"host"`
+	Port int    `json:"port"`
+	Type string `json:"type"`
+}
+
+// Allowlist is a list of allowlist entries; the allowlist itself
+type Allowlist struct {
+	Entries []AllowlistEntry
+}
+
```
sfc-gh-pfus, line 32: General notice - it would be better, to create a struct like `connectivityDiagnoser` os sth and move all functions inside. Otherwise, you create a function that seems very generic (like `resolveHostname` or scary `isAcceptableStatusCode`) that is specific for this tool only but that is available in the whole module.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	if err != nil {
+		return fmt.Errorf("[fetchCRL] failed to read response body: %w", err)
+	}
+	// TODO: do we want to persist CRL to disk ? (then cleanup, deal with read-only fs, etc)
+	logger.Infof("[fetchCRL] %s retrieved successfully (%d bytes), not persisting to disk.\n", uri, len(body))
+	return nil
+}
+
+func doHTTP(request *http.Request) error {
+	if request.URL.Host == "ocsp.snowflakecomputing.com" {
```
sfc-gh-pfus: 1. Will it work for China?
2. Will it work for privatelinks?
But generally speaking - we are planing to remove OCSP in a month or two, so probably before this tool is widely used. Maybe it's not worth investing in OCSP checks here.

connectivity_diagnosis.go
sfc-gh-pfus: I agree with Dawid - it may be worth having this function work even if HTTPS fails. Maybe for downloading HTTPS we should use `insecureMode` of TLS config?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	certs := tlsState.PeerCertificates
+	logger.Infof("[doHTTPSGetCerts] Retrieved %d certificate(s).\n", len(certs))
+
+	// log individual cert details
+	for i, cert := range certs {
+		logger.Infof("[doHTTPSGetCerts] Certificate %d, serial number: %x\n", i+1, cert.SerialNumber)
+		logger.Infof("[doHTTPSGetCerts]   Subject: %s\n", cert.Subject)
+		logger.Infof("[doHTTPSGetCerts]   Issuer:  %s\n", cert.Issuer)
+		logger.Infof("[doHTTPSGetCerts]   Valid:   %s to %s\n", cert.NotBefore, cert.NotAfter)
+		logger.Infof("[doHTTPSGetCerts]   For further details, check https://crt.sh/?serial=%x (non-Snowflake site)\n", cert.SerialNumber)
```
sfc-gh-pfus: Is this serial number going to pretty print correctly?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+				logger.Infof("[doHTTPSGetCerts]    - %s\n", dp)
+				// only try to download the actual CRL if configured to do so
+				if downloadCRLs {
+					if err := fetchCRL(dp); err != nil {
+						logger.Errorf("[doHTTPSGetCerts]      Failed to fetch CRL: %v\n", err)
+					}
+				}
+			}
+		} else {
+			logger.Infof("[doHTTPSGetCerts]   CRL Distribution Points not available")
```
sfc-gh-pfus: `DPs not available` sounds scary - maybe `not present/included/<use better word> in the certificate`?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+		request.URL = newURL
+	}
+	logger.Infof("[doHTTP] testing HTTP connection to %s\n", request.URL.String())
+	resp, err := client.Do(request)
+	if err != nil {
+		return fmt.Errorf("HTTP GET to %s endpoint failed: %w", request.URL.String(), err)
+	}
+	defer resp.Body.Close()
+
+	if !isAcceptableStatusCode(resp.StatusCode, acceptableStatusCodes) {
```
sfc-gh-pfus: Should we care for this? I think that we make a very dummy request, so it is fine if we get any answer including bad request, unauthorized or even 500.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	if diagnosticTransport == nil {
+		logger.Errorf("[checkProxy] diagnosticTransport is nil")
+		return
+	}
+	if diagnosticTransport.Proxy == nil {
+		// no proxy configured, nothing to log
+		return
+	}
+	p, err := diagnosticTransport.Proxy(req)
+	if p != nil {
```
sfc-gh-pfus: After checking err?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+func performConnectivityCheck(entryType, host string, port int, downloadCRLs bool) (err error) {
+	var protocol string
+	var req *http.Request
+
+	if port == 80 {
+		protocol = "http"
+	} else if port == 443 {
+		protocol = "https"
+	} else {
+		// we should never arrive here
```
sfc-gh-pfus, line 332: Why?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+	logger.Infof("[performConnectivityCheck] %s check for %s %s\n", strings.ToUpper(protocol), entryType, host)
+	req, err = createRequest(fmt.Sprintf("%s://%s", protocol, host))
+	if err != nil {
+		logger.Errorf("[performConnectivityCheck] error creating request: %v\n", err)
+		return err
+	}
+
+	checkProxy(req)
+
+	if port == 80 {
```
sfc-gh-pfus: Why not depend on `protocol`?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,366 @@
...
+		logger.Errorf("[performConnectivityCheck] error performing %s check: %v\n", strings.ToUpper(protocol), err)
+		return err
+	}
+
+	return nil
+}
+
+func performDiagnosis(cfg *Config) {
+	allowlistFile := cfg.ConnectionDiagnosticsAllowlistFile
+	downloadCRLs := cfg.ConnectionDiagnosticsDownloadCRL
```
sfc-gh-pfus: Maybe we could depend on a fact that CRL is enabled in cfg?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+	return buffer, cleanup
+}
+
+func TestSetupTestLogger(t *testing.T) {
+	// save  original global logger
+	originalLogger := logger
+	// and restore it after test
+	defer func() { logger = originalLogger }()
+
+	t.Run("Setup Test Logger - buffer capture and cleanup", func(t *testing.T) {
```
sfc-gh-pfus: Why do we need internal run?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+		// clear the buffer, log a new message into it
+		// logs should not go to the test logger anymore
+		buffer.Reset()
+		logger.Info("this should not appear in test buffer")
+		assertEqualE(t, buffer.String(), "", "buffer should be empty after cleanup")
+	})
+}
+
+// test case types
+type tcAllowlistEntry struct {
```
sfc-gh-pfus: Is it the same as the original one?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+	for _, tc := range testcases {
+		t.Run(fmt.Sprintf("%s_%d_%s", tc.host, tc.port, tc.entryType), func(t *testing.T) {
+			entry := AllowlistEntry{
+				Host: tc.host,
+				Port: tc.port,
+				Type: tc.entryType,
+			}
+
+			assertEqualE(t, entry.Host, tc.host, "host did not match")
```
sfc-gh-pfus: What is the point of this test?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+			assertNotNilE(t, client, "client should not be nil")
+			assertEqualE(t, client.Timeout, tc.expectedTimeout, "timeout did not match")
+			assertNotNilE(t, client.Transport, "transport should not be nil")
+		})
+	}
+}
+
+func TestCreateDiagnosticDialContext(t *testing.T) {
+	t.Run("Diagnostic DialContext creation", func(t *testing.T) {
```
sfc-gh-pfus: Why `t.Run`?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+		// new simple server to test basic connectivity
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		}))
+		defer server.Close()
+
+		u, _ := url.Parse(server.URL)
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
sfc-gh-pfus: We can even lower it to like 100ms, it is a local connection, right?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+		defer cancel()
+
+		conn, err := dialContext(ctx, "tcp", u.Host)
+		assertNilE(t, err, "error should be nil")
+		defer conn.Close()
```
sfc-gh-pfus: Do we need this defer as this is the last line of the function?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+		transport := createDiagnosticTransport(config)
+
+		assertNotNilE(t, transport, "transport should not be nil")
+		assertNotNilE(t, transport.DialContext, "dialContext should not be nil")
+
+		// by default we should use the SnowflakeTransport
+		assertEqualE(t, transport.TLSClientConfig, SnowflakeTransport.TLSClientConfig, "TLSClientConfig did not match with SnowflakeTransport default")
+		assertEqualE(t, transport.MaxIdleConns, SnowflakeTransport.MaxIdleConns, "MaxIdleConns did not match with SnowflakeTransport default")
+		assertEqualE(t, transport.IdleConnTimeout, SnowflakeTransport.IdleConnTimeout, "IdleConnTimeout did not match with SnowflakeTransport default")
+		if (transport.Proxy == nil) != (SnowflakeTransport.Proxy == nil) {
```
sfc-gh-pfus: `assertEqualE`?

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+		conn, err := dialContext(ctx, "tcp", u.Host)
+		assertNilE(t, err, "error should be nil")
+		defer conn.Close()
+
+	})
+}
+
+func TestCreateDiagnosticTransport(t *testing.T) {
+	t.Run("Diagnostic Transport creation", func(t *testing.T) {
```
sfc-gh-pfus: Same.

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+func TestOpenAndReadAllowlistJSON(t *testing.T) {
+	testcases := []tcOpenAllowlistJSON{
+		{
+			name: "Open and Read Allowlist - valid file path",
+			// create a temp allowlist file and then delete it
+			setup: func() (filePath string, cleanup func()) {
+				content := `[{"host":"myaccount.snowflakecomputing.com","port":443,"type":"SNOWFLAKE_DEPLOYMENT"}]`
+				tmpFile, _ := os.CreateTemp("", "allowlist_*.json")
+				_, _ = tmpFile.WriteString(content)
```
sfc-gh-pfus: Should we assertNilF on err here? Otherwise we silently ignore failed test setup. Same a line above.

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+	for _, tc := range testcases {
+		t.Run(tc.hostname, func(t *testing.T) {
+			result := isPrivateLinkHost(tc.hostname)
+			assertEqualE(t, result, tc.isPrivateLink, "could not determine if host is private link")
+		})
+	}
+}
+
+func TestIsPrivateIP(t *testing.T) {
```
sfc-gh-pfus: Do we need this? We actually test golang library, not our code.

connectivity_diagnosis.go
sfc-gh-pfus: This is a global variable inside the whole module, but with a very common name - not good. But supposing you lean towards agreeing on "general notice" it won't be relevant.

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+				assertNotNilE(t, req, "request should not be nil")
+				assertEqualE(t, req.Method, "GET", "method should be GET")
+				assertEqualE(t, req.URL.String(), tc.uri, "url should match")
+			}
+		})
+	}
+}
+
+func TestDoHTTP(t *testing.T) {
+	// 'client' is the global diagnostic client, lets preserve the original value
```
sfc-gh-pfus: If you need to leave such a comment - it means something is not right :) But if you change according to "general notice", it should magically disappear :)

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+		// set up transport with proxy
+		proxyURL, _ := url.Parse("http://my.pro.xy:8080")
+		diagnosticTransport = &http.Transport{
+			Proxy: func(req *http.Request) (*url.URL, error) {
+				return proxyURL, nil
+			},
+		}
+
+		// this should generate a log output which indicates we use a proxy
```
sfc-gh-pfus, line 569: Maybe this function could return proxy? I think that testing based on the expected return values is better than to expect logs, but this whole tool is about logs, so it's not that terrible.

connectivity_diagnosis_test.go
```diff
@@ -0,0 +1,930 @@
...
+		assertStringContainsE(t, logOutput, "[doHTTPSGetCerts] Retrieved", "should contain certificate retrieval log")
+		assertStringContainsE(t, logOutput, "certificate(s)", "should contain certificate count information")
+
+		// diagnostic dial context
+		assertStringContainsE(t, logOutput, "[createDiagnosticDialContext] Connected to", "should contain dial context connection logs")
+		assertStringContainsE(t, logOutput, "remote IP:", "should contain remote IP information")
+
+		// CRL download
+		// if certificate has CRLDistributionPoints this message is logged
+		if strings.Contains(logOutput, "CRL Distribution Points:") {
```
sfc-gh-pfus, line 814: We have `assertStringContainsE`.

dsn.go
```diff
@@ -321,6 +325,15 @@ func DSN(cfg *Config) (dsn string, err error) {
 	if cfg.DisableSamlURLCheck != configBoolNotSet {
 		params.Add("disableSamlURLCheck", strconv.FormatBool(cfg.DisableSamlURLCheck != ConfigBoolFalse))
 	}
+	if cfg.ConnectionDiagnosticsEnabled {
+		params.Add("connectionDiagnosticsEnabled", strconv.FormatBool(cfg.ConnectionDiagnosticsEnabled))
```
sfc-gh-pfus, line 369: Can we add tests for this in dsn_test.go and connection_configuration_test.go?

connectivity_diagnosis.go
sfc-gh-dszmolka: no, you're right

connectivity_diagnosis.go
sfc-gh-dszmolka: same you're right :)

connectivity_diagnosis.go
sfc-gh-dszmolka: you're also right; i'm using the Transport for the proxy check, derived from the diagnosticClient itself. so for now preferred to keep it (after changing the 'general approach'), but pls let me know if you have a strong preference of removing

connectivity_diagnosis.go
sfc-gh-dszmolka: ok!

connectivity_diagnosis.go
sfc-gh-dszmolka: ok!

connectivity_diagnosis.go
sfc-gh-dszmolka: removed

connectivity_diagnosis.go
sfc-gh-dszmolka: mostly thinking about the custom transport which the user can specify and i wanted to make sure we always have _some_  valid transport at the end of the check. If you recommend to remove it, i can remove it

connectivity_diagnosis.go
sfc-gh-dszmolka: thank you ! the code has been removed with combining the read + parse parts for the allowlist json

connectivity_diagnosis.go
sfc-gh-dszmolka: you're right, they have been combined

connectivity_diagnosis.go
sfc-gh-dszmolka: noted - thank you!

connectivity_diagnosis.go
sfc-gh-dszmolka: i wanted to have control over what is and what is not accepted as a content, which can be even (hopefully) modified or extended in the future, should the SYSTEM$ALLOWLIST output change, hoping it needs to be done only in the structure

if this is unwanted or incorrect, let me know please and i'll look into alternatives

connectivity_diagnosis.go
sfc-gh-dszmolka: yes, intent was to use the same resolver as the driver does, in hope of catching DNS resolution errors/weirdness

connectivity_diagnosis.go
sfc-gh-dszmolka: you're right again, and i missed we already have this function. modified.

connectivity_diagnosis.go
sfc-gh-dszmolka: well not _that_ much :) removed

connectivity_diagnosis.go
sfc-gh-dszmolka: i heeded this general direction and refactored with using struct for the whole functionality. Thank you !

connectivity_diagnosis.go
sfc-gh-dszmolka: 1. good question :) modified to also catch the default .cn domain
2. no, and it doesn't have to; as privatelink 'OCSP' doesn't go on port 80
3. per latest-latest developments on OCSP removal vs CRL, i would still like to leave legacy OCSP part in the checker, looks like it's still necessary

connectivity_diagnosis.go
sfc-gh-dszmolka: did a couple of manual tests and it did print correctly

sfc-gh-dszmolka: also re: `certs := tlsState.PeerCertificates` vs insecureMode
diagnostic client uses the transport from getTransport

so if the user configures disableOCSPChecks || insecureMode in the generic driver context, the diagnoser will also use no-OCSP Transport in the HTTPS checks

connectivity_diagnosis.go
sfc-gh-dszmolka: good point, edited

connectivity_diagnosis.go
sfc-gh-dszmolka: actually we don't make a very dummy request, we try to connect to the valid, existing endpoints listed in the ALLOWLIST json. the idea behind 'acceptable codes' is to see if there's any very fishy going on, probably due to proxy or such things. 

for snowflake endpoint, we expect 200, for storage endpoint we expect 403, anything else is something which might be worth looking at

connectivity_diagnosis.go
sfc-gh-dszmolka: ok - checking err first now, and existence of proxy later

connectivity_diagnosis.go
sfc-gh-dszmolka: today we only have port 80 and port 443 endpoints. i dont expect this to change in the future but what if it does, or i somehow missed some service which is not listening on 80 or 443 . we should know about it at least ... but open to suggestions.

connectivity_diagnosis.go
sfc-gh-dszmolka: they should be the same as they are mapped 1:1. but i can agree maybe protocol looks a little bit nicer. modified.

connectivity_diagnosis.go
sfc-gh-dszmolka: 1. it did not yet exist when i started with this pr :) 
2. still, i would like to give separate control to the user, whether they want to try download the CRL or not, what if its enabled and they don't want to download the CRL?

connectivity_diagnosis_test.go
sfc-gh-dszmolka: nevermind the previous comment, i got you now. moved it out of t.Run

connectivity_diagnosis_test.go
sfc-gh-dszmolka: yes, testcases based on the original struct. its now also removed.

connectivity_diagnosis_test.go
sfc-gh-dszmolka: removed

connectivity_diagnosis_test.go
sfc-gh-dszmolka: see above please; its from an earlier recommendation

sfc-gh-dszmolka: moved it out of t.Run

connectivity_diagnosis_test.go
sfc-gh-dszmolka: removed

connectivity_diagnosis_test.go
sfc-gh-dszmolka: i remember having some problem with that approach, i'll double check if really, and what

connectivity_diagnosis_test.go
sfc-gh-dszmolka: got it out of t.Run too

connectivity_diagnosis_test.go
sfc-gh-dszmolka: good point. added assertion

connectivity_diagnosis_test.go
sfc-gh-dszmolka: nope - removed

connectivity_diagnosis.go
sfc-gh-dszmolka: yes, the approach has been refactored heeding your recommendation on 'general notice', creating a separate struct for diagnoser

connectivity_diagnosis_test.go
sfc-gh-dszmolka: yup , modified the approach

connectivity_diagnosis_test.go
sfc-gh-dszmolka: this is indeed the generic idea, to log stuff and make it visible for the user, who deliberately chose to enable the diagnostic. for now i'd like to leave it like this, but i'm open to suggestions if you think i should change it after all.

connectivity_diagnosis_test.go
sfc-gh-dszmolka: yes but i don't want to fail the test at this point if it doesn't contain CRL endpoint, just simply step over it gracefully and not even attempt the next part (which _does_ hard fail if CRL fetching is not logged)

dsn.go
sfc-gh-dszmolka: you're absolutely right, i should have thought about this. added.

connectivity_diagnosis_test.go
sfc-gh-dszmolka: yes it panicks with
```
runtime error: comparing uncomparable type func(*http.Request) (*url.URL, error)
```

so for now, i used the original approach

connectivity_diagnosis_test.go
sfc-gh-dszmolka: modified

connectivity_diagnosis.go
sfc-gh-dszmolka: re: certs := tlsState.PeerCertificates vs insecureMode
diagnostic client uses the transport from getTransport

so if the user configures disableOCSPChecks || insecureMode in the generic driver Config, the diagnoser will also use no-OCSP Transport in the HTTPS checks

connectivity_diagnosis.go
sfc-gh-dszmolka: looks like this when run manually
```
INFO[0001]log.go:192 gosnowflake.(*defaultLogger).Infof [doHTTPSGetCerts] Certificate 1, serial number: 221ab1ba870027c5512741c69ce0807 
INFO[0001]log.go:192 gosnowflake.(*defaultLogger).Infof [doHTTPSGetCerts]   Subject: CN=*.us-central1.gcp.snowflakecomputing.com,O=Snowflake Inc.,L=Bozeman,ST=Montana,C=US 
INFO[0001]log.go:192 gosnowflake.(*defaultLogger).Infof [doHTTPSGetCerts]   Issuer:  CN=DigiCert Global G2 TLS RSA SHA256 2020 CA1,O=DigiCert Inc,C=US 
INFO[0001]log.go:192 gosnowflake.(*defaultLogger).Infof [doHTTPSGetCerts]   Valid:   2025-04-29 00:00:00 +0000 UTC to 2026-04-29 23:59:59 +0000 UTC 
INFO[0001]log.go:192 gosnowflake.(*defaultLogger).Infof [doHTTPSGetCerts]   For further details, check https://crt.sh/?serial=221ab1ba870027c5512741c69ce0807 (non-Snowflake site) 
```

connectivity_diagnosis.go
sfc-gh-pfus: You can always extract it from the `client`, right?

connectivity_diagnosis.go
sfc-gh-pfus: Very good answer!

connectivity_diagnosis.go
sfc-gh-pfus: My thinking was that this diagnostic tool runs checks based on the client connection config, to replicate the behaviour and give as accurate replication as possible. WDYT?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+type allowlistEntry struct {
+	Host string `json:"host"`
+	Port int    `json:"port"`
+	Type string `json:"type"`
+}
+
+type allowlist struct {
+	Entries []allowlistEntry
```
sfc-gh-pfus, line 30: lowercase? Or is it required to deserialize JSON?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// acceptable HTTP status codes for connectivity diagnosis
+// for the sake of connectivity, e.g. 403 is perfectly fine
+var acceptableStatusCodes = []int{http.StatusOK, http.StatusForbidden}
```
sfc-gh-pfus: It's too general a name. Either put it in `connectivityDiagnoser` (but I doubt it's a good idea because structs fields can't have default values) or prepend it with meaningful prefix.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+
+	var httpTransport *http.Transport
+	if t, ok := baseTransport.(*http.Transport); ok {
+		httpTransport = t
+	} else {
+		httpTransport = SnowflakeTransport
+	}
+
+	// return a new transport enhanced with remote IP logging
+	// for SnowflakeNoOcspTransport, TLSClientConfig is nil
```
sfc-gh-pfus, line 105: Is there something wrong with having nil TLSClientConfig? Do we want OCSP checks to be triggered by diagnostic tool by default?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+	return &http.Transport{
+		TLSClientConfig: httpTransport.TLSClientConfig,
+		MaxIdleConns:    httpTransport.MaxIdleConns,
+		IdleConnTimeout: httpTransport.IdleConnTimeout,
+		Proxy:           httpTransport.Proxy,
+		DialContext:     cd.createDiagnosticDialContext(),
+	}
+}
+
+func (cd *connectivityDiagnoser) openAndReadAllowlistJSON(filePath string) (allowlist allowlist, err error) {
```
sfc-gh-pfus, line 115: Maybe just `createAllowlist`?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	logger.Infof("[openAndReadAllowlistJSON] reading allowlist from %s.\n", filePath)
+	fileContent, err := os.ReadFile(filePath)
+	if err != nil {
+		return allowlist, err
+	}
+
+	logger.Debug("[openAndReadAllowlistJSON] parsing allowlist.json")
+	err = json.Unmarshal(fileContent, &allowlist.Entries)
+	if err != nil {
```
sfc-gh-pfus: Just `return allowlist, err`? If you still want to use naked return it can be even `return` instead.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("[fetchCRL] HTTP response status from endpoint: %s", resp.Status)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("[fetchCRL] failed to read response body: %w", err)
+	}
+	logger.Infof("[fetchCRL] %s retrieved successfully (%d bytes), not persisting to disk.\n", uri, len(body))
```
sfc-gh-pfus: Can we try parsing this? For OCSP we sometimes had problems with parsing OCSP response. I believe it is more stable for CRL, but still - if we already have it, it's worth trying to parse it.

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+	return nil
+}
+
+func (cd *connectivityDiagnoser) doHTTP(request *http.Request) error {
+	if strings.HasPrefix(request.URL.Host, "ocsp.snowflakecomputing.") {
+		fullOCSPCacheURI := request.URL.String() + "/ocsp_response_cache.json"
+		newURL, _ := url.Parse(fullOCSPCacheURI)
```
sfc-gh-pfus: Why do we omit errors?

connectivity_diagnosis.go
```diff
@@ -0,0 +1,363 @@
...
+	if cd.diagnosticTransport.Proxy == nil {
+		// no proxy configured, nothing to log
+		return
+	}
+	p, err := cd.diagnosticTransport.Proxy(req)
+	if err != nil {
+		logger.Errorf("[checkProxy] problem determining PROXY: %v\n", err)
+	}
+	if p != nil {
+		logger.Infof("[checkProxy] PROXY detected in the connection: %v\n", p)
```
sfc-gh-pfus: What is expected to be printed here?

connectivity_diagnosis.go
sfc-gh-pfus: For local snowflake instance we connect on 8082 or sth :) But we can leave it as is.

driver.go
```diff
@@ -54,6 +55,12 @@ func (d SnowflakeDriver) OpenWithConfig(ctx context.Context, config Config) (dri
 		}
 	}
 	logger.WithContext(ctx).Info("OpenWithConfig")
+	if config.ConnectionDiagnosticsEnabled {
+		logger.WithContext(ctx).Infof("Connection diagnostics enabled. Allowlist file from config: %s, will check CRLs in certificates: %s",
+			config.ConnectionDiagnosticsAllowlistFile, strconv.FormatBool(config.ConnectionDiagnosticsDownloadCRL))
+		performDiagnosis(&config)
+		logger.WithContext(ctx).Info("Connection diagnostics finished.")
```
sfc-gh-pfus, line 64: IIRC in JDBC we stop after running diagnostics. I believe it is safer. Can you check it in JDBC and do the same?

connectivity_diagnosis.go
sfc-gh-dszmolka: yes i'm using it internally to deserialize the JSON

connectivity_diagnosis.go
sfc-gh-dszmolka: agreed! renamed to `connDiagAcceptableStatusCodes`

connectivity_diagnosis.go
sfc-gh-dszmolka: the generic idea behind this approach is that this is not a separate tool (as it is not a separate tool in Python, JDBC, and ODBC either), but part of the driver. Takes the same config as the driver, and should use the ~same configuration as the driver where it makes sense.

if the OCSP checks are enabled on the driver and working or failing, they should work (or fail) on the conndiag too. 
if the OCSP checks are disabled by the user on the driver (and their use-case in the driver) then it will be disabled in the conndiag too

Of course this thought can be flawed or suboptimal, i'm open to suggestions

connectivity_diagnosis.go
sfc-gh-dszmolka: i see where you coming from; we're creating and populating an `allowlist` struct, but however we're not creating and allowlist (as in: input allowlist.json). I want to avoid confusing the user. 

Is the function name too long, or against some convention? Tried to name it to reflect what it does, but please suggest what it should be (which is not confusing for the user; we're not _creating_ an allowlist (JSON) here) :)

connectivity_diagnosis.go
sfc-gh-dszmolka: okay this is a good idea, modified!

connectivity_diagnosis.go
sfc-gh-dszmolka: original idea was, this is a connectivity diagnosis tool. it tries to connect to the OCSP endpoint (TCP connectivity ok) and fetch the crl (HTTP connectivity ok)

it does not analyze or parse the CRL. 

But i see why your idea is good, as it can detect additional errors which are beyond the connectivity. However i'm super concerned about the massive CRLs which even today during normal operations, are 10-15 megs. Without microsoft revoking the 30 million certs..

Question is, do we want to parse the 15m crl stuff , and if so, probably it would need to be implemented to parse each individual CRL only once, even though they _will_ appear in multiple certificates. 
Should perhaps this put behind another config flag, this 'parsing' functionality?

connectivity_diagnosis.go
sfc-gh-dszmolka: fair point, added error handling

connectivity_diagnosis.go
sfc-gh-dszmolka: the proxy configured for the connection. if a proxy is used. e.g. for `HTTPS_PROXY=http://my.pro.xy:8080 go run ...` it's
```
time="2025-07-24T09:23:31Z" level=info msg="[checkProxy] PROXY detected in the connection: http://my.pro.xy:8080" func="gosnowflake.(*defaultLogger).Infof" file="log.go:192"
```

driver.go
sfc-gh-dszmolka: yes, JDBC indeed exists. modified the approach here accordingly.

connectivity_diagnosis.go
sfc-gh-dszmolka: the current approach also allows me to set it once, and use it 'globally' everywhere across all the diagnostic functions (without setting anything really global). it's very convenient. 

i can extract it from the client, and then create it as a global variable because it's needed in multiple functions. isn't the current approach nicer and cleaner? i mean this is a genuine question, what is the preferred approach, i thought we want to use local variables wherever possible

connectivity_diagnosis.go
sfc-gh-dszmolka: so this is only created once, and used in all places where Transport is used, or that was my intent:
```
	return &http.Transport{
		TLSClientConfig: httpTransport.TLSClientConfig,
		MaxIdleConns:    httpTransport.MaxIdleConns,
		IdleConnTimeout: httpTransport.IdleConnTimeout,
		Proxy:           httpTransport.Proxy,
>>		DialContext:     cd.createDiagnosticDialContext(),
	}
```

connectivity_diagnosis.go
sfc-gh-dszmolka: i feel kinda strongly about _not_ reaching out to CRL whenever that is possible. They are big and ugly. Especially if we now want to also parse them. 

But with the approach that 'exit after diag test', this is now entirely different, so conn.diag need to check the CRLs for the user because the 'regular driver' will not check it. (it will not even connect anywhere, exit after conn.diag)
Let me change this.

connectivity_diagnosis.go
sfc-gh-pfus: I meant across the whole driver, but after implementing CRLs and reading more about timeouts I'm not as convinced.

connectivity_diagnosis.go
sfc-gh-pfus: But it creates `allowlist` struct, right? I'm not saying anything about JSON ;) This name exposes some unneeded internal details. From caller perspective it what it does is just it creates allowlist struct. But this is a minor comment, I'm fine with leaving it as-is.

connectivity_diagnosis.go
sfc-gh-pfus: Having in mind "run diag, exit", I think it's fine to parse it.

connectivity_diagnosis.go
sfc-gh-pfus: Hmmm, I can see, that you use it only in checking proxy. So instead of having this:
```
if cd.diagnosticTransport.Proxy == nil {
```
We can have:
```
if cd.diagnosticClient.Transport.Proxy == nil {
```
For me it's not a big deal, but it's safer. Otherwise, you manage two objects and it's easy to make a mistake and override this global `diagnosticTransport` variable and not set it in `diagnosticClient`. It's easy to not make such a mistake for you, because you know it cold now, but for others looking at it in a year or two, it may not be that obvious. I'm a fan of writing code in a manner that is less error-prone, sometimes sacrificing "nicety and cleanity" as you mentioned :) WDYT?

connectivity_diagnosis.go
sfc-gh-dszmolka: Pending: https://github.com/snowflakedb/gosnowflake/pull/1476 to be merged so detection of 'is crl enabled' could be added.

connectivity_diagnosis.go
sfc-gh-dszmolka: this has been implemented with the latest comment (parse crl, do not dump it)

connectivity_diagnosis.go
sfc-gh-dszmolka: okay, you convinced me, removed this from the struct :)

connectivity_diagnosis.go
sfc-gh-dszmolka: considering `CertRevocationCheckMode` has now been added

sfc-gh-pbulawa (approved):

PR #1446, opened 2025-06-18
SNOW-2161709: Fix permission handling for configuration.toml and session/token files in auto configuration mode

sfc-gh-gmerticariu:
### Description

SNOW-2161709 

#### Summary
This PR updates the file permission validation logic in `connection_configuration.go` and its corresponding tests in `connection_configuration_test.go`. The changes are necessary to allow more flexible file permissions in environments where configuration files need to be readable by users other than the owner.

#### Changes
1. **File Permission Validation Logic**:
   - The previous implementation enforced strict `0600` file permissions for configuration files, which caused issues in environments where files need to be readable by others.
   - Updated the `validateFilePermission` function to:
     - Log a warning if the file is readable by someone other than the owner.
     - Return an error if the file is writable or executable by others.
   - Introduced the `shouldSkipWarningForReadPermissions` function to allow skipping warnings for readable permissions via the `GOSNOWFLAKE_SKIP_FILE_READ_PERMISSION_WARNING` environment variable.

2. **Error Messages**:
   - Updated error messages in `errors.go` to provide more specific feedback:
     - `errMsgInvalidWritablePermissionToFile`: Indicates the file is writable by someone other than the owner.
     - `errMsgInvalidExecutablePermissionToFile`: Indicates the file is executable.

3. **Unit Tests**:
   - Refactored `TestTokenFilePermission` in `connection_configuration_test.go` to use `t.Run` for better organization and readability.
   - Added tests for:
     - Warning logs when files are readable by others.
     - Skipping warnings for readable permissions using the environment variable.
     - Errors for writable and executable file permissions.

#### Reason for Change
The old implementation enforced `0600` file permissions, which was overly restrictive for certain environments. For example, shared configurations may need to be readable by multiple users. This change introduces flexibility while maintaining security by:
- Allowing readable permissions with warnings.
- Enforcing stricter checks for writable and executable permissions.

#### Testing
- Verified the updated logic with unit tests covering all scenarios (readable, writable, executable permissions).
- Ensured backward compatibility by maintaining strict checks for writable and executable permissions.

### Checklist

### DIFF
--- a/connection_configuration.go
+++ b/connection_configuration.go
@@ -340,16 +340,36 @@ func validateFilePermission(filePath string) error {
 	if err != nil {
 		return err
 	}
-	if permission := fileInfo.Mode().Perm(); permission != os.FileMode(0600) {
+
+	permission := fileInfo.Mode().Perm()
+
+	if !shouldSkipWarningForReadPermissions() && (permission&os.FileMode(0040) != 0 || permission&os.FileMode(0004) != 0) {
+		logger.Warnf("file '%v' is readable by someone other than the owner. Your Permission: %v", filePath, permission)
+	}
+
+	if permission&os.FileMode(0111) != 0 {
 		return &SnowflakeError{
 			Number:      ErrCodeInvalidFilePermission,
-			Message:     errMsgInvalidPermissionToTomlFile,
-			MessageArgs: []interface{}{permission},
+			Message:     errMsgInvalidExecutablePermissionToFile,
+			MessageArgs: []interface{}{filePath, permission},
 		}
 	}
+
+	if permission&os.FileMode(0020) != 0 || permission&os.FileMode(0002) != 0 {
+		return &SnowflakeError{
+			Number:      ErrCodeInvalidFilePermission,
+			Message:     errMsgInvalidWritablePermissionToFile,
+			MessageArgs: []interface{}{filePath, permission},
+		}
+	}
+
 	return nil
 }
 
 func shouldReadTokenFromFile(cfg *Config) bool {
 	return cfg != nil && cfg.Authenticator == AuthTypeOAuth && len(cfg.Token) == 0
+}
+
+func shouldSkipWarningForReadPermissions() bool {
+	return os.Getenv("GOSNOWFLAKE_SKIP_FILE_READ_PERMISSION_WARNING") != ""
 }
--- a/connection_configuration_test.go
+++ b/connection_configuration_test.go
@@ -1,6 +1,8 @@
 package gosnowflake
 
 import (
+	"bytes"
+	"fmt"
 	"io/fs"
 	"os"
 	path "path/filepath"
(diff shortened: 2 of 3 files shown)


connection_configuration.go
```diff
@@ -336,20 +340,41 @@ func validateFilePermission(filePath string) error {
...
+
 	fileInfo, err := os.Stat(filePath)
 	if err != nil {
 		return err
 	}
-	if permission := fileInfo.Mode().Perm(); permission != os.FileMode(0600) {
+
+	permission := fileInfo.Mode().Perm()
+
+	if !shouldSkipWarningForReadPermissions() && permission&othersCanReadFilePermission != 0 {
```
sfc-gh-rsavenok, line 353: Since SNOWFLAKE_HOME can be customized, I wonder if we should allow disabling both warning and error with a single parameter.

WDYT? (dont edit this code yet, lets discuss in the thread and w8 for @sfc-gh-dheyman opinion)

connection_configuration.go
```diff
@@ -336,20 +340,41 @@ func validateFilePermission(filePath string) error {
...
+
 	return nil
 }
 
 func shouldReadTokenFromFile(cfg *Config) bool {
 	return cfg != nil && cfg.Authenticator == AuthTypeOAuth && len(cfg.Token) == 0
 }
+
+func shouldSkipWarningForReadPermissions() bool {
+	return os.Getenv("GOSNOWFLAKE_SKIP_FILE_READ_PERMISSION_WARNING") != ""
```
sfc-gh-rsavenok: All of driver env variables are prefixed with `SNOWFLAKE` instead of `GOSNOWFLAKE`

Also, WDYT about naming it`SKIP_CONFIGURATION/CONNECTION_FILE....` to indicate the usage context?

connection_configuration.go
```diff
@@ -336,20 +340,41 @@ func validateFilePermission(filePath string) error {
...
 	fileInfo, err := os.Stat(filePath)
 	if err != nil {
 		return err
 	}
-	if permission := fileInfo.Mode().Perm(); permission != os.FileMode(0600) {
+
+	permission := fileInfo.Mode().Perm()
+
+	if !shouldSkipWarningForReadPermissions() && permission&othersCanReadFilePermission != 0 {
+		logger.Warnf("file '%v' is readable by someone other than the owner. Your Permission: %v", filePath, permission)
```
sfc-gh-rsavenok: I think this warning should mention that there's an env variable to skip this check.

Maybe write something like... `If this is intentional, set SNOWFLAKE_..... to skip the check`

errors.go
```diff
@@ -325,7 +325,8 @@ const (
 	errMsgNullValueInMap                     = "for handling null values in maps use WithMapValuesNullable(ctx)"
 	errMsgFailedToParseTomlFile              = "failed to parse toml file. the params %v occurred error with value %v"
 	errMsgFailedToFindDSNInTomlFile          = "failed to find DSN in toml file."
-	errMsgInvalidPermissionToTomlFile        = "file permissions different than read/write for user. Your Permission: %v"
+	errMsgInvalidWritablePermissionToFile    = "file '%v' is writable by someone other than the owner. Your Permission: %v"
```
sfc-gh-rsavenok: It would be nice to have more descriptive error messages here explaining why both cases are a security risk. 

Also, mention that it can be disabled with env variable (if my other comment would apply)

sfc-gh-fochnik (commented):
ProdSec review: LGTM

connection_configuration.go
sfc-gh-gmerticariu: I followed the same model here https://github.com/snowflakedb/gosnowflake/blob/dd94b3bafbc9c52aa017fd5fddaef2526db74da5/driver.go#L89, for naming the variable. I will update the rest of the variable to reflect better what it does.

connection_configuration.go
sfc-gh-gmerticariu: I'm a bit on the fence with disabling the error. We should encourage using correct permissions where possible and more than `644` is really not a good practice. Moreover, right now we error when the permission is `600` and nobody complained (based on what I found in threads). I will defer to @sfc-gh-dheyman for the final decision though.

connection_configuration.go
sfc-gh-rsavenok: Ah, the driver has a mix of `GOSNOWFLAKE_*` and `SNOWFLAKE_*` environment variables, like `SNOWFLAKE_DEFAULT_CONNECTION_NAME` or `SNOWFLAKE_HOME`.

My thinking was that we probably want to have the same env var name in every driver, but looking at other drivers, they have their own prefixes 😭

Do you think there will ever be a need to set such a variable on SPCS (by snowflake) so the warning is suppressed for every driver? If not, I'm fine with having GOSNOWFLAKE, if yes - I'd like to document this decision in the jira epic so other drivers have exactly the same env name

sfc-gh-gmerticariu: In SPCS we would need to supress both GO and Python. By your logic, it makes sense to prefix with `SNOWFLAKE_`, and not with `GOSNOWFLAKE_`

connection_configuration.go
```diff
@@ -16,6 +16,12 @@ const (
 	snowflakeConnectionName = "SNOWFLAKE_DEFAULT_CONNECTION_NAME"
 	snowflakeHome           = "SNOWFLAKE_HOME"
 	defaultTokenPath        = "/snowflake/session/token"
+
+	othersCanReadFilePermission  = os.FileMode(0044)
+	othersCanWriteFilePermission = os.FileMode(0022)
+	executableFilePermission     = os.FileMode(0111)
+
+	skipWarningForReadPermissionsEnv = "SNOWFLAKE_SKIP_WARNING_FOR_READ_PERMISSIONS_ON_CONFIG_FILE"
```
sfc-gh-dheyman: We tend to add SF_ prefix instead of SNOWFLAKE_

sfc-gh-dheyman: I'm also wondering if it's worth to add a new parameter for this, since warnings are not that invasive. Is it just for the sake of reducing logs clutter?

connection_configuration.go
sfc-gh-dheyman: Agreed with @sfc-gh-gmerticariu. I don't think disabling errors would be correct from security standpoint.

connection_configuration.go
sfc-gh-dheyman: If we'll decide to leave this param, we should also include this information in doc.go file.

connection_configuration_test.go
```diff
@@ -9,50 +11,124 @@ import (
...
+	})
+
+	t.Run("test warning skipped logger for readable outside owner", func(t *testing.T) {
+		os.Setenv(skipWarningForReadPermissionsEnv, "true")
+		var originalLogger = logger
+		logger = CreateDefaultLogger()
+		buf := &bytes.Buffer{}
+		logger.SetOutput(buf)
+
+		defer func() {
```
sfc-gh-dheyman, line 69: I'd split this into two `defer` clauses to avoid it not being run if e.g. creating logger causes panic.

connection_configuration_test.go
```diff
@@ -9,50 +11,124 @@ import (
...
+		assertNilF(t, err, "The error occurred because you cannot change the file permission")
+
+		_, err = loadConnectionConfig()
+		assertNilF(t, err, "The error should not occur")
+
+		assertEmptyStringE(t, buf.String())
+	})
+
+	t.Run("test writable connection file other than owner", func(t *testing.T) {
+		err = os.Chmod("./test_data/connections.toml", 0666)
```
sfc-gh-dheyman, line 83: shouldn't we clean up and set the permissions to the correct ones after the test?

connection_configuration.go
sfc-gh-gmerticariu: It is done exclusively for avoiding log clutter. Given that the file is read every time `Open` is called, we would create quite some logs. I will change the prefix to `SF_` and update the doc.go file

connection_configuration_test.go
sfc-gh-gmerticariu: Since each test is setting its own permissions, I created a defer on line 26 (https://github.com/snowflakedb/gosnowflake/pull/1446/files#diff-152b5a9b206b2e7b0600f48d01f7180d5a7fa58bb524df9264bc8f27bcd6805eR26) which is resetting the file to the original permissions before the test bundle is run.

sfc-gh-dheyman: Alright, I somehow missed this.

sfc-gh-dheyman (approved):

sfc-gh-rsavenok (approved):

PR #1463, opened 2025-07-11
SNOW-2021015: test optimisation

sfc-gh-pcyrek:
### Description

SNOW-2021015: test optimisation

### Checklist

### DIFF
--- a/put_get_test.go
+++ b/put_get_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/sha256"
 	"database/sql"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


bindings_test.go
```diff
@@ -45,17 +45,14 @@ const (
...
 	// range to use for generating random numbers
 	lobRandomRange = 100000
 )
 
 var (
-	// maxLOBSize = 128 * 1024 * 1024 // new max LOB size
-	maxLOBSize = 16 * 1024 * 1024 // current max LOB size
-	largeSize  = maxLOBSize / 2
-	mediumSize = largeSize / 2
+// No additional size variables needed - just use smallSize and largeSize
```
sfc-gh-pfus: Can we remove this section?

bindings_test.go
```diff
@@ -1552,6 +1542,34 @@ func testInsertLOBData(t *testing.T, useArrowFormat bool, isLiteral bool) {
 	})
 }
 
+func fastStringGeneration(size int) string {
```
sfc-gh-pfus, line 1541: Why do we need this? If this one is better maybe we can just replace `randomString`?

put_get_test.go
```diff
@@ -1039,28 +1036,76 @@ func testPutGetLargeFile(t *testing.T, isStream bool, autoCompress bool) {
...
 		assertNilE(t, err)
-		assertEqualF(t, contents, string(originalContents), "data did not match content")
+		originalChecksum := fmt.Sprintf("%x", originalHash.Sum(nil))
+
+		assertEqualF(t, downloadedChecksum, originalChecksum, "file integrity check failed - checksums don't match")
 	})
 }
+
+func createLimitedRealFile(t *testing.T, sourceDir string) (string, func()) {
+	tmpFile, err := os.CreateTemp(sourceDir, "largefile_*.txt")
```
sfc-gh-pfus: Maybe we can use `t.TempFile()`? We don't have to clean it after the test then.

bindings_test.go
sfc-gh-pcyrek: ok

bindings_test.go
sfc-gh-pcyrek: fastStringGeneration is faster, but works for different purposes ;) 

tl;dr instead of randomly generating a string it uses a repeating pattern of:
pattern := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"


Serves different needs. I would keep them both.

put_get_test.go
sfc-gh-pcyrek: used t.tempdir()

PR #1465, opened 2025-07-14
SNOW-2117152 Implement CRL algorithm

sfc-gh-pfus:
### Description

SNOW-2117152 Implemented the main CRL algorithm.

### Checklist

### DIFF
--- a/crl.go
+++ b/crl.go
@@ -114,20 +114,17 @@ func (cv *crlValidator) verifyPeerCertificates(rawCerts [][]byte, verifiedChains
 	}
 	crlValidationResults := cv.validateChains(verifiedChains)
 
+	allRevoked := true
 	for _, result := range crlValidationResults {
 		if result == crlUnrevoked {
 			logger.Debug("found certificate chain with no revoked certificates")
 			return nil
 		}
-	}
-
-	allRevoked := true
-	for _, result := range crlValidationResults {
 		if result != crlRevoked {
 			allRevoked = false
-			break
 		}
 	}
+
 	if allRevoked {
 		return fmt.Errorf("every verified certificate chain contained revoked certificates")
 	}
@@ -350,7 +347,9 @@ func (cv *crlValidator) getFromCache(crlURL string) (*x509.RevocationList, *time
 		// promote CRL to in-memory cache
 		cv.inMemoryCacheMutex.Lock()
 		cv.inMemoryCache[crlURL] = &crlInMemoryCacheValueType{
-			crl:          crl,
+			crl: crl,
+			// modTime is not the exact time the CRL was downloaded, but rather the last modification time of the file
+			// still, it is good enough for our purposes
 			downloadTime: &modTime,
 		}
 		cv.inMemoryCacheMutex.Unlock()
@@ -429,6 +428,17 @@ func (cv *crlValidator) verifyAgainstIdpExtension(crl *x509.RevocationList, dist
 		}
 	}
 	return nil
+}
+
+func (cv *crlValidator) getOrCreateMutex(crlURL string) *sync.Mutex {
+	cv.inMemoryCacheMutex.Lock()
+	mu, ok := cv.crlURLMutexes[crlURL]
+	if !ok {
+		mu = &sync.Mutex{}
+		cv.crlURLMutexes[crlURL] = mu
+	}
+	cv.inMemoryCacheMutex.Unlock()
+	return mu
 }
 
 func isShortLivedCertificate(cert *x509.Certificate) bool {
--- a/crl.go
+++ b/crl.go
@@ -370,28 +370,31 @@ func (cv *crlValidator) updateCache(crlURL string, crl *x509.RevocationList, dow
 		return
 	}
 	crlFilePath := cv.crlURLToPath(crlURL)
-	err := os.WriteFile(crlFilePath, crl.Raw, 0600)
-	if err != nil {
+	if err := os.MkdirAll(filepath.Dir(crlFilePath), 0755); err != nil {
+		logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
+		return
+	}
+	if err := os.WriteFile(crlFilePath, crl.Raw, 0600); err != nil {
 		logger.Warnf("failed to write CRL to disk for %v (%v): %v", crlURL, crlFilePath, err)
 	}
 }
 
-func (cv *crlValidator) downloadCrl(url string) (*x509.RevocationList, *time.Time, error) {
-	logger.Debugf("downloading CRL from %v", url)
+func (cv *crlValidator) downloadCrl(crlURL string) (*x509.RevocationList, *time.Time, error) {
+	logger.Debugf("downloading CRL from %v", crlURL)
 	now := time.Now()
-	resp, err := cv.httpClient.Get(url)
+	resp, err := cv.httpClient.Get(crlURL)
 	if err != nil {
 		return nil, nil, err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode >= 400 {
-		return nil, nil, fmt.Errorf("failed to download CRL from %v, status code: %v", url, resp.StatusCode)
+		return nil, nil, fmt.Errorf("failed to download CRL from %v, status code: %v", crlURL, resp.StatusCode)
 	}
 	crlBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, nil, err
 	}
-	logger.Debugf("downloaded %v bytes for CRL %v", len(crlBytes), url)
+	logger.Debugf("downloaded %v bytes for CRL %v", len(crlBytes), crlURL)
 	crl, err := x509.ParseRevocationList(crlBytes)
 	if err != nil {
 		return nil, nil, err
--- a/crl.go
+++ b/crl.go
@@ -1,0 +1,434 @@
+package gosnowflake
+
+import (
+	"crypto/x509"
+	"encoding/asn1"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+)
+
+var idpOID = asn1.ObjectIdentifier{2, 5, 29, 28}
+
+type distributionPointName struct {
+	FullName []asn1.RawValue `asn1:"optional,tag:0"`
+}
+
+type issuingDistributionPoint struct {
+	DistributionPoint distributionPointName `asn1:"optional,tag:0"`
+}
+
+type crlValidator struct {
+	certRevocationCheckMode        CertRevocationCheckMode
+	serialCertificateValidation    bool
+	allowCertificatesWithoutCrlUrl bool
+	cacheValidityTime              time.Duration
+	inMemoryCacheDisabled          bool
+	inMemoryCache                  map[string]*crlInMemoryCacheValueType
+	inMemoryCacheMutex             sync.Mutex
+	onDiskCacheDisabled            bool
+	onDiskCacheDir                 string
+	crlUrlMus                      map[string]*sync.Mutex
+	httpClient                     *http.Client
+}
+
+type crlInMemoryCacheValueType struct {
+	crl          *x509.RevocationList
+	downloadTime *time.Time
+}
+
+func newCrlValidator(certRevocationCheckMode CertRevocationCheckMode, serialCertificateValidation, allowCertificatesWithoutCrlUrl bool, cacheValidityTime time.Duration, inMemoryCacheDisabled, onDiskCacheDisabled bool, onDiskCacheDir string, httpClient *http.Client) *crlValidator {
+	var inMemoryCache map[string]*crlInMemoryCacheValueType
+	if !inMemoryCacheDisabled {
+		inMemoryCache = make(map[string]*crlInMemoryCacheValueType)
+	}
+	return &crlValidator{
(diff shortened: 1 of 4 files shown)


crl.go
```diff
@@ -0,0 +1,448 @@
...
+
+	for _, result := range crlValidationResults {
+		if result == crlUnrevoked {
+			logger.Debug("found certificate chain with no revoked certificates")
+			return nil
+		}
+	}
+
+	allRevoked := true
+	for _, result := range crlValidationResults {
```
sfc-gh-dheyman: Can't we condense both of the loops (this and above one)?

crl.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+	}
+	return certUnrevoked
+}
+
+func (cv *crlValidator) validateCrlAgainstCrlURL(cert *x509.Certificate, crlURL string, parent *x509.Certificate) certValidationResult {
+	now := time.Now()
+
+	cv.inMemoryCacheMutex.Lock()
+	mu, ok := cv.crlURLMutexes[crlURL]
```
sfc-gh-dheyman: perhaps can be extracted to getorCreateMutex(...) for better clarity?

crl.go
```diff
@@ -0,0 +1,448 @@
...
+		if cert.SerialNumber.Cmp(rce.SerialNumber) == 0 {
+			logger.Warnf("certificate for %v (serial number %v) has been revoked at %v, reason: %v", cert.Subject, rce.SerialNumber, rce.RevocationTime, rce.ReasonCode)
+			return certRevoked
+		}
+	}
+
+	return certUnrevoked
+}
+
+func (cv *crlValidator) validateCrl(crl *x509.RevocationList, parent *x509.Certificate, crlURL string) error {
```
sfc-gh-dheyman, line 250: is `parent` the correct nomenclature? Shouldn't we use `issuer`?

crl.go
```diff
@@ -0,0 +1,448 @@
...
+		return nil, nil
+	}
+	modTime := stat.ModTime()
+
+	if !cv.inMemoryCacheDisabled {
+		// promote CRL to in-memory cache
+		cv.inMemoryCacheMutex.Lock()
+		cv.inMemoryCache[crlURL] = &crlInMemoryCacheValueType{
+			crl:          crl,
+			downloadTime: &modTime,
```
sfc-gh-dheyman, line 318: perhaps comment pointing out repercussions of using modTime might be useful?

crl.go
```diff
@@ -0,0 +1,448 @@
...
+				continue
+			}
+
+			if isShortLivedCertificate(cert) {
+				logger.Debugf("certificate %v is short-lived, skipping CRL validation", cert.Subject)
+				continue
+			}
+
+			if len(cert.CRLDistributionPoints) == 0 {
+				if cv.allowCertificatesWithoutCrlURL {
```
sfc-gh-dheyman, line 159: this function overall has really deep nesting. Can we reduce it by code extraction?

crl.go
```diff
@@ -0,0 +1,448 @@
...
+	if cv.onDiskCacheDisabled {
+		logger.Debugf("CRL cache is disabled, not writing to disk for %v", crlURL)
+		return
+	}
+	crlFilePath := cv.crlURLToPath(crlURL)
+	if err := os.MkdirAll(filepath.Dir(crlFilePath), 0755); err != nil {
+		logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
+		return
+	}
+	if err := os.WriteFile(crlFilePath, crl.Raw, 0600); err != nil {
```
sfc-gh-dheyman, line 345: what about 644? Are we sure that 0600 will always work?

crl_test.go
```diff
@@ -0,0 +1,778 @@
...
+	assertNilF(t, err)
+	cv := newTestCrlValidator(t, CertRevocationCheckEnabled, false)
+	err = cv.verifyAgainstIdpExtension(crl, "http://c.pki.goog/we2/yK5nPhtHKQs.crl")
+	assertNilE(t, err)
+	err = cv.verifyAgainstIdpExtension(crl, "http://c.pki.goog/we2/other.crl")
+	assertNotNilF(t, err)
+	assertStringContainsE(t, err.Error(), "distribution point http://c.pki.goog/we2/other.crl not found in CRL IDP extension")
+}
+
+func TestParallelRequestToTheSameCrl(t *testing.T) {
```
sfc-gh-dheyman, line 547: perhaps it would be worth to leverage `t.Parallel()` here? https://brandur.org/t-parallel

crl_test.go
```diff
@@ -0,0 +1,778 @@
...
+			Locality:           []string{"Warsaw"},
+			CommonName:         "localhost",
+		},
+		NotBefore: time.Now(),
+		NotAfter:  notAfter,
+		IsCA:      false,
+	}
+	return createCert(t, certTemplate, issuerCert, issuerPrivateKey, crlEndpoint)
+}
+
```
sfc-gh-dheyman, line 706: any tests for allowCertificatesWithoutCrlURLType?

crl_test.go
```diff
@@ -0,0 +1,778 @@
...
+		case onDiskCacheDisabledType:
+			onDiskCacheDisabled = bool(v)
+		default:
+			t.Fatalf("unexpected argument type %T", v)
+		}
+	}
+	cacheDir := t.TempDir()
+	return newCrlValidator(checkMode, serialCertificateValidation, allowCertificatesWithoutCrlURL, cacheValidityTime, inMemoryCacheDisabled, onDiskCacheDisabled, cacheDir, httpClient)
+}
+
```
sfc-gh-dheyman, line 64: test case for cache expiry behaviour?

crl.go
sfc-gh-pfus: Yes, we can.

crl.go
sfc-gh-pfus: I wanted it to be as close to the design pseudocode as possible, to make it "reference implementation".

crl.go
sfc-gh-pfus: Ok, done.

crl.go
sfc-gh-pfus: Parent is a nomenclature used in Golang, example:
image

crl.go
sfc-gh-pfus: Added.

crl.go
sfc-gh-pfus: Why not?

crl_test.go
sfc-gh-pfus: There already is "should redownload when evicted in cache" test - isn't it good enough?

crl_test.go
sfc-gh-pfus: No, because it will create separate instances of crlValidator and countingRoundTripper.

crl_test.go
sfc-gh-pfus: There is `CertWithNoCrlDistributionPointsAllowed` - isn't it enough?

crl.go
sfc-gh-dheyman: I know in JDBC it was connected to SPCS. Perhaps here there's no need for this.

sfc-gh-dheyman (approved):

crl.go
```diff
@@ -0,0 +1,452 @@
...
+	crlURLMutexes                  map[string]*sync.Mutex
+	httpClient                     *http.Client
+}
+
+type crlInMemoryCacheValueType struct {
+	crl          *x509.RevocationList
+	downloadTime *time.Time
+}
+
+func newCrlValidator(certRevocationCheckMode CertRevocationCheckMode, serialCertificateValidation, allowCertificatesWithoutCrlURL bool, cacheValidityTime time.Duration, inMemoryCacheDisabled, onDiskCacheDisabled bool, onDiskCacheDir string, httpClient *http.Client) *crlValidator {
```
sfc-gh-jkasten: no action required: I find negative names like `inMemoryCacheDisabled` harder to read.

`if !inMemoryCacheDisabled` vs `if inMemoryCacheEnabled` -- I realize it is subjective though.

crl.go
```diff
@@ -0,0 +1,452 @@
...
+
+			certStatus := cv.validateCertificate(cert, chain[j+1])
+			if certStatus == certRevoked {
+				crlValidationResults[i] = crlRevoked
+				break
+			}
+
+			if certStatus == certError {
+				crlValidationResults[i] = crlError
+				break
```
sfc-gh-jkasten: nit: I think that this break could potentially cover up a revocation higher up in the chain. (Downgrading the problem  and allowing the connection to succeed depending on config)

crl.go
```diff
@@ -0,0 +1,452 @@
...
+		return cv.validateCertificateSerial(cert, parent)
+	}
+
+	return cv.validateCertificateInParallel(cert, parent)
+}
+
+func (cv *crlValidator) validateCertificateSerial(cert *x509.Certificate, parent *x509.Certificate) certValidationResult {
+	for _, crlURL := range cert.CRLDistributionPoints {
+		result := cv.validateCrlAgainstCrlURL(cert, crlURL, parent)
+		if result == certRevoked || result == certError {
```
sfc-gh-jkasten, line 192: nit/optional: This has the same sort of certError covering up revocation, but it seems less problematic (it should be very rare to see a cert with multiple CRL DPs) (same below if we feel inclined to change it)

crl.go
```diff
@@ -0,0 +1,452 @@
...
+			return result
+		}
+	}
+	return certUnrevoked
+}
+
+func (cv *crlValidator) validateCertificateInParallel(cert *x509.Certificate, parent *x509.Certificate) certValidationResult {
```
sfc-gh-jkasten: Note: I think certs with have multiple CRL DPs are explicitly discouraged, so this might not get much use. (I haven't run a censys.io query to come back with exact numbers)
https://github.com/cabforum/servercert/blob/main/docs/BR.md#:~:text=When%20present%2C%20the%20CRL%20Distribution%20Points%20extension%20MUST%20contain%20at%20least%20one%20DistributionPoint%3B%20containing%20more%20than%20one%20is%20NOT%20RECOMMENDED.%20All%20DistributionPoint%20items%20must%20be%20formatted%20as%20follows%3A

(Parallelizing the validation of each cert in the chain though would be utilized heavily)

sfc-gh-jkasten (commented):
Thanks, Piotr! Looks great thus far. I made it down to `validateCrlAgainstCrlURL`, but it got late on me and I am quite tired.

I will try to finish tomorrow morning with a fresh set of eyes. I can also review it after submission if you need to get moving on it. Sorry for the delay.

crl.go
sfc-gh-pfus: Generally agreed, but I wanted to have a default behaviour to match default bool value which is false - which means, that if a client doesn't set this parameter (means false), caching still works. Thus this negative name.

crl.go
sfc-gh-pfus: It's not a "nit", very valid comment!

crl.go
sfc-gh-pfus: Good one!

crl.go
sfc-gh-pfus: Ok, so leaving as is to keep implementation simple.

crl.go
```diff
@@ -0,0 +1,423 @@
...
+	shouldUpdateCrl := false
+
+	if needsFreshCrl {
+		newCrl, newDownloadTime, err := cv.downloadCrl(crlURL)
+		if err != nil {
+			logger.Warnf("failed to download CRL from %v: %v", crlURL, err)
+		}
+		shouldUpdateCrl = newCrl != nil && (crl == nil || newCrl.ThisUpdate.After(crl.ThisUpdate))
+		if shouldUpdateCrl {
+			logger.Debugf("Updating CRL for %v", crlURL)
```
sfc-gh-jkasten:
```suggestion
			logger.Debugf("Found updated CRL for %v", crlURL)
```

nit/optional: If the CRL is not valid we technically will not use it and cache it below. The current text sounds like have already decided to substitute the new for the old.

sfc-gh-jkasten (approved):
Awesome work! Thanks Piotr!

crl.go
sfc-gh-jkasten: I think switching this to a continue should resolve it?

sfc-gh-rsavenok (approved):

crl.go
```diff
@@ -0,0 +1,423 @@
...
+			crl = newCrl
+			downloadTime = newDownloadTime
+		} else {
+			if crl != nil && crl.NextUpdate.Before(now) {
```
sfc-gh-fochnik: Shouldn't this be `After`?

crl.go
```diff
@@ -0,0 +1,423 @@
...
+
+	logger.Debugf("CRL has %v entries, next update at %v", len(crl.RevokedCertificateEntries), crl.NextUpdate)
+	if err := cv.validateCrl(crl, parent, crlURL); err != nil {
+		return certError
+	}
+
+	if shouldUpdateCrl {
+		logger.Debugf("CRL for %v is valid, updating cache", crlURL)
+		cv.updateCache(crlURL, crl, downloadTime)
+	}
```
sfc-gh-fochnik, line 238: Nit: it feels like everything up to this point should be in a separate function.

sfc-gh-pfus: I tried to have it as close to the design as possible, to call it "reference implementation".

crl.go
sfc-gh-pfus: Right, done!

Issue #1351, opened 2025-04-03
Cancel context cause 2 different errors


nearsyh:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of GO driver are you using?

**v1.12.1**

   
2. What operating system and processor architecture are you using?

**MacOS darwin/arm64**

   
3. What version of GO are you using?
run `go version` in your console

**go version go1.22.4 darwin/arm64**

4.Server version:* E.g. 1.90.1
You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```

**9.8.1**

5. What did you do?

```
func Test_CancelContext(t *testing.T) {

	db, err := // open db
        defer db.Close()

	conn, err := db.Conn(context.Background())
	if err != nil {
		t.Fatal(err)
	}
	defer conn.Close()

	if _, err := conn.ExecContext(context.Background(), "CREATE TEMP TABLE test AS SELECT random() AS num FROM table(generator(rowcount => 10000))"); err != nil {
		t.Fatal(err)
	}

	for i := 0; i < 10; i++ {
		ctx, cancel := context.WithCancel(context.Background())

		var g errgroup.Group
		g.Go(func() error {
			var avg float64
			err := conn.QueryRowContext(ctx, "SELECT avg(num) FROM test;").Scan(&avg)
			return err
		})

		time.Sleep(500 * time.Millisecond)
		cancel()

		if err := g.Wait(); err == nil || !errors.Is(err, context.Canceled) {
			t.Errorf("Got err %v, wanted context.Canceled", err)
		}
	}
}
```

6. What did you expect to see?

I expect to see that all returned errors are `context.Canceled`. However, **I also saw `000605: Identified SQL statement is not currently executing` sometime.**

sfc-gh-dszmolka:
hello - can you please review https://github.com/snowflakedb/gosnowflake/issues/1292 and see if it already captures the issue you're seeing? To me it looks similar and I would like to make sure we focus efforts. Thank you !

nearsyh:
They look similar, but I'm not sure if they have the same root cause. In my case, this issue can be reproduced without concurrent queries.

sfc-gh-dszmolka:
okay, thank you. we'll take a look

sfc-gh-dszmolka:
(https://github.com/snowflakedb/gosnowflake/issues/1351 and https://github.com/snowflakedb/gosnowflake/issues/1292 likely related)

[...]

PR #1466, opened 2025-07-15
SNOW-2022394 Fix issues with cancel query

sfc-gh-rsavenok:
### Description
* Fix an issue in `OpenWithConfig` where a cancelled context during `connectionTelemetry` was treated as a successful connection. This resulted in a cancel query operation for a query that was never sent to the back-end.
* Changed the behaviour of `cancelQuery` to return success if `000605` error was returned after retries. https://github.com/snowflakedb/gosnowflake/issues/1351
* created `test_utils_test.go` file to store testing utils in a common place instead of test files

### Checklist


driver_test.go
```diff
@@ -1978,6 +1978,32 @@ func TestOpenWithConfig(t *testing.T) {
 	db.Close()
 }
 
+func TestOpenWithConfigCancel(t *testing.T) {
+	wiremock = newWiremock()
```
sfc-gh-pfus: We don't need to reinitialize it, right? Just use `wiremock` global variable.

driver_test.go
```diff
@@ -1978,6 +1978,32 @@ func TestOpenWithConfig(t *testing.T) {
...
+	)
+	driver := SnowflakeDriver{}
+	config := wiremock.connectionConfig()
+	blockingRoundTripper := newBlockingRoundTripper(snowflakeNoOcspTransport, nil)
+	countingRoundTripper := newCountingRoundTripper(blockingRoundTripper)
+	config.Transporter = countingRoundTripper
+
+	paths := []string{"/session/v1/login-request", "/telemetry/send"}
+	for _, path := range paths {
+		t.Run("canceled during request:"+path, func(t *testing.T) {
```
sfc-gh-pfus: Maybe it should be two separate tests, because they need a bit different asserts? As for now, if a timeout happens during auth call, telemetry is never sent. But auth request is sent in both cases, right?

restful.go
```diff
@@ -473,8 +473,13 @@ func cancelQuery(ctx context.Context, sr *snowflakeRestful, requestID UUID, time
 				return err
 			}
 			return sr.FuncCancelQuery(ctx, sr, requestID, timeout)
-		} else if !respd.Success && respd.Code == queryNotExecutingCode && ctxRetry != 0 {
-			return sr.FuncCancelQuery(context.WithValue(ctx, cancelRetry, ctxRetry-1), sr, requestID, timeout)
+		} else if !respd.Success && respd.Code == queryNotExecutingCode {
+			if ctxRetry != 0 {
+				return sr.FuncCancelQuery(context.WithValue(ctx, cancelRetry, ctxRetry-1), sr, requestID, timeout)
```
sfc-gh-pfus, line 478: Maybe we can also decrease the number of retires? Currently it takes 5s, right? How much time do we need to ensure that the query might have had enough time to reach GS (to avoid cancelling before query even reached GS)?

restful.go
```diff
@@ -473,8 +473,13 @@ func cancelQuery(ctx context.Context, sr *snowflakeRestful, requestID UUID, time
...
+			}
+			// After exhausting retries, we can safely treat queryNotExecutingCode as success
```
sfc-gh-pfus, line 480: Worth logging something, right?

restful_test.go
```diff
@@ -611,11 +584,67 @@ func TestCancelRetry(t *testing.T) {
...
+	t.Run("cancel failure returns wrapped error", func(t *testing.T) {
+		sr := newRestfulWithError(context.Canceled)
+		sr.FuncCancelQuery = func(context.Context, *snowflakeRestful, UUID, time.Duration) error {
+			cancelCalled = true
+			return fmt.Errorf("fatal failure")
+		}
+		_, err := runPostRestfulQuery(sr)
+		assertTrueF(t, cancelCalled)
+		assertNotNilF(t, err)
+		assertEqualF(t, "failed to cancel query. cancelErr: fatal failure, queryErr: context canceled", err.Error())
```
sfc-gh-pfus: Can we ensure, that error unwrapping works fine? https://rollbar.com/blog/golang-wrap-and-unwrap-error/

restful_test.go
```diff
@@ -611,11 +584,67 @@ func TestCancelRetry(t *testing.T) {
...
+		}
+	}
+	runPostRestfulQuery := func(sr *snowflakeRestful) (data *execResponse, err error) {
+		return postRestfulQuery(context.Background(), sr, &url.Values{}, nil, nil, 0, NewUUID(), nil)
+	}
+
+	t.Run("postRestfulQuery error does not trigger cancel", func(t *testing.T) {
+		sr := newRestfulWithError(fmt.Errorf("query error"))
+		_, err := runPostRestfulQuery(sr)
+		assertFalseF(t, cancelCalled)
```
sfc-gh-pfus: We use `assert***F` to mark "preconditions" and `assert***E` to mark inconsistencies on expected outputs. So in that case it should be:

```
assertFalseE(t, cancelCalled)
assertNotNilF(t, err)
assertEqualE(t, "query error", err.Error())
```

restful_test.go
```diff
@@ -611,11 +584,67 @@ func TestCancelRetry(t *testing.T) {
...
+		assertNotNilF(t, err)
+		assertEqualF(t, "query error", err.Error())
+	})
+
+	t.Run("context.Canceled triggers cancel", func(t *testing.T) {
+		sr := newRestfulWithError(context.Canceled)
+		_, err := runPostRestfulQuery(sr)
+		assertTrueF(t, cancelCalled)
+		assertNotNilF(t, err)
+		assertEqualF(t, context.Canceled.Error(), err.Error())
```
sfc-gh-pfus: Maybe it would be better to use:

```
assertErrIsE
```

restful.go
```diff
@@ -473,8 +473,13 @@ func cancelQuery(ctx context.Context, sr *snowflakeRestful, requestID UUID, time
...
+			// since it indicates the query has already completed and there's nothing left to cancel
+			return nil
```
sfc-gh-pfus, line 483: Just to be sure - what if a context is cancelled or timed out? Do we return such an error on the other level?

sfc-gh-rsavenok: Yes, the 2 new tests "context.NAME triggers cancel" check for it. Since `FuncCancelQuery` returns nil, it will use the original query error.

This block will get cancelErr = nul and will return original error
```
if err == context.Canceled || err == context.DeadlineExceeded {
		// For context cancel/timeout cases, a special cancel request needs to be sent.
		if cancelErr := sr.FuncCancelQuery(context.Background(), sr, requestID, timeout); cancelErr != nil {
			// Wrap the original error with the cancel error.
			err = fmt.Errorf("failed to cancel query. cancelErr: %w, queryErr: %w", cancelErr, err)
		}
	}

	return data, err
```

restful.go
sfc-gh-rsavenok: I'd rather keep it as is and maybe revisit in universal driver implementation because it might break something.

https://snowflakecomputing.atlassian.net/browse/SNOW-342743 this ticket from 2021 doesn't give enough context on why 5 times were used.

restful_test.go
sfc-gh-rsavenok: Not sure why it doesn't work, but unwrap gives me nil:
```
fmt.Printf("Error: %v\n", err)
fmt.Printf("Unwrapped error: %v\n", errors.Unwrap(err))
		
Error: failed to cancel query. cancelErr: fatal failure, queryErr: context canceled
Unwrapped error: <nil>
```

driver_test.go
sfc-gh-rsavenok: They need the same assertions but a different path. However, I agree with you that having 2 tests instead of a loop would make tests much easier to understand.
```
	t.Run("canceled during request:login-request", func(t *testing.T) {
		blockingRoundTripper.setPathBlockTime("/session/v1/login-request", 50*time.Millisecond)
		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
		defer cancel()
		_, err := driver.OpenWithConfig(ctx, *config)
		assertErrIsE(t, context.DeadlineExceeded, err)
		assertEqualE(t, countingRoundTripper.totalRequestsByPath("/session/v1/login-request"), 1)
	})

	t.Run("canceled during request:telemetry/send", func(t *testing.T) {
		blockingRoundTripper.setPathBlockTime("/telemetry/send", 50*time.Millisecond)
		ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
		defer cancel()
		_, err := driver.OpenWithConfig(ctx, *config)
		assertErrIsE(t, context.DeadlineExceeded, err)
		assertEqualE(t, countingRoundTripper.totalRequestsByPath("/telemetry/send"), 1)
	})
```

restful.go
sfc-gh-rsavenok: Added

restful_test.go
sfc-gh-pfus: Maybe we somewhere on the upper level incorrectly create the error?

sfc-gh-pfus: Ok, I checked, that unwrap can be used only if one error is wrapped. You have to use `errors.Is` instead.

driver.go
```diff
@@ -72,6 +72,9 @@ func (d SnowflakeDriver) OpenWithConfig(ctx context.Context, config Config) (dri
 
 	sc.startHeartBeat()
 	sc.internal = &httpClient{sr: sc.rest}
+	if ctx.Err() != nil {
```
sfc-gh-pfus, line 76: Maybe worth adding a comment why we need this?

test_utils_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	for _, reqs := range crt.postReqCount {
+		total += reqs
+	}
+	return total
+}
+
+type blockingRoundTripper struct {
+	delegate      http.RoundTripper
+	blockTime     time.Duration
```
sfc-gh-pfus: `defaultBlockTime`?

sfc-gh-pfus (approved):

restful_test.go
```diff
@@ -611,11 +584,67 @@ func TestCancelRetry(t *testing.T) {
...
+	t.Run("cancel failure returns wrapped error", func(t *testing.T) {
+		fatalCancelErr := fmt.Errorf("fatal failure")
+		sr := newRestfulWithError(context.Canceled)
+		sr.FuncCancelQuery = func(context.Context, *snowflakeRestful, UUID, time.Duration) error {
+			cancelCalled = true
+			return fatalCancelErr
+		}
+		_, err := runPostRestfulQuery(sr)
+		assertTrueE(t, cancelCalled)
+		assertTrueE(t, errors.Is(err, context.Canceled))
```
sfc-gh-pfus: We have `assertErrIsE`.

sfc-gh-pfus (approved):

PR #1468, opened 2025-07-16
Respect max retry count when no timeout

jlourenc:
### Description

The HTTP retrier ignores the max retry count when no timeout (0 value) is set. In most of the cases, the timeout comes from the REST Request Timeout, which defaults to 0. This means the HTTP retrier may retry indefinitely (e.g. fails to open connections).

This changeset updates the retry logic to always consider max retry count, alongside the timeout if its value is greater than 0.

### Checklist


jlourenc:

chunk_downloader.go
```diff
@@ -677,7 +679,7 @@ func (f *httpStreamChunkFetcher) fetch(URL string, rows chan<- []*string) error
 	if err != nil {
 		return fmt.Errorf("parsing URL: %w", err)
 	}
-	res, err := newRetryHTTP(context.Background(), f.client, http.NewRequest, fullURL, f.headers, 0, 0, defaultTimeProvider, nil).execute()
+	res, err := newRetryHTTP(f.ctx, f.client, http.NewRequest, fullURL, f.headers, f.timeout, f.maxRetryCount, defaultTimeProvider, nil).execute()
```
sfc-gh-pfus, line 682: I'm a bit afraid of passing this context here. Probably it's fine, and we even want to honour cancels/timeouts...

retry_test.go
```diff
@@ -344,29 +344,47 @@ func TestRetryQueryFailWithTimeout(t *testing.T) {
...
+				success:    false,
+			}
+			urlPtr, err := url.Parse("https://fakeaccountretryfail.snowflakecomputing.com:443/queries/v1/query-request?" + requestIDKey)
+			assertNilF(t, err, "failed to parse the test URL")
+			_, err = newRetryHTTP(context.Background(),
+				client,
+				emptyRequest, urlPtr, make(map[string]string), tc.timeout, maxRetryCount, defaultTimeProvider, nil).doPost().setBody([]byte{0}).execute()
+			assertNotNilF(t, err, "should fail to run retry")
+			var values url.Values
+			values, err = url.ParseQuery(urlPtr.RawQuery)
```
sfc-gh-pfus, line 376: Why don't use `values, err := url.Parse...`?

retry_test.go
```diff
@@ -344,29 +344,47 @@ func TestRetryQueryFailWithTimeout(t *testing.T) {
...
+			if err != nil {
```
sfc-gh-pfus, line 377: Can you use our assertions (`assertNilF`)?

chunk_downloader.go
jlourenc: Let me know how you'd like me to proceed. My main concern here was that previously it'd retry on retryable errors (because of timeout set to `0`). Without updating at least the max retry count, the call would no longer retry on its first retryable error.

retry_test.go
jlourenc: The test in itself has not been touched, I have simply added an additional test case for when timeout is set to `0`.
Happy to follow your lead though.

chunk_downloader.go
sfc-gh-pfus: Ok, let's keep it as you changed.

retry_test.go
sfc-gh-pfus: Ah, ok, so it's not your responsibility to fix it :)

sfc-gh-pfus (approved):

sfc-gh-rsavenok (approved):

sfc-gh-pfus:
Hi @jlourenc ! One test is failing on your PR, can you take a look?
`TestOpenWithConfigCancel`

jlourenc:
Hi @sfc-gh-pfus ! Sure thing. I have not run the tests but given the CI error and the util function, it looks like the `actual` and `expected` in the assertions were reversed which caused the ultimate `errors.Is` to be reversed as well. That would trigger failures in case of error wrapping.

https://github.com/snowflakedb/gosnowflake/pull/1468/commits/926b3a2f6a49cf42c0c5345ca59b966620a89db3

sfc-gh-pfus:
Thanks @jlourenc for your input! Your change is merged and should be released this week.

PR #1471, opened 2025-07-18
Return storage errors and log from callsites

felipecrv:
### Description

Return errors from `secureCredentialStorage` implementations and:

1. Log them at the callsites
2. Assert on them in unit tests

This makes it easier to identify why higher level flows (e.g. external browser auth) are failing and makes the unit tests more strict. Without these errors, it's very easy to break the storage managers and not break tests because the code is written to tolerate storage errors.

### Checklist


felipecrv:
recheck

felipecrv:
I expect some unit tests to fail in CI now that I'm asserting on `setCredential` and `deleteCredential` errors. I will adapt after I see them.

felipecrv:

felipecrv:
recheck

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
-			})
-			account := strings.ToUpper(tokenSpec.user)
-			item := keyring.Item{
-				Key:  account,
-				Data: []byte(value),
-			}
-			if err := ring.Set(item); err != nil {
-				logger.Debugf("Failed to write to keychain. Err: %v", err)
-			}
+			// don't leak any info in the error message
```
sfc-gh-pfus: Have you encountered any leaks here?

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
+			return fmt.Errorf("Failed to open Windows credential manager.")
+		}
+		item := keyring.Item{
+			Key:  credentialsKey,
+			Data: []byte(value),
+		}
+		if err := ring.Set(item); err != nil {
+			logger.Debugf("Failed to write to Windows credential manager. Err: %v", err)
```
sfc-gh-pfus: Nothing is leaked here? Shouldn't we return the error as well?

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
+			// don't leak any info in the error message
+			return fmt.Errorf("Failed to open macOS keychain.")
+		}
+		account := strings.ToUpper(tokenSpec.user)
+		item := keyring.Item{
+			Key:  account,
+			Data: []byte(value),
+		}
+		if err := ring.Set(item); err != nil {
+			return fmt.Errorf("Failed to write to keychain. Err: %v", err)
```
sfc-gh-pfus, line 501: Same here - no leaking here?

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
 		}
 	}
+	return nil
 }
 
-func (ssm *keyringSecureStorageManager) getCredential(tokenSpec *secureTokenSpec) string {
+func (ssm *keyringSecureStorageManager) getCredential(tokenSpec *secureTokenSpec) (string, error) {
 	cred := ""
```
sfc-gh-pfus, line 509: Maybe we can get rid of this var here, and return where it should be returned instead? Seems cleaner. I know this is not your code, but boy scout rule :)

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
 		})
+		if err != nil {
+			// don't leak any info in the error message
+			return "", fmt.Errorf("Failed to open macOS keychain.")
+		}
 		account := strings.ToUpper(tokenSpec.user)
 		i, err := ring.Get(account)
 		if err != nil {
-			logger.Debugf("Failed to find the item in keychain or item does not exist. Error: %v", err)
+			return "", fmt.Errorf("Failed to find the item in keychain or item does not exist. Error: %v", err)
```
sfc-gh-pfus, line 537: Same here.

secure_storage_manager.go
```diff
@@ -467,68 +459,83 @@ func newKeyringBasedSecureStorageManager() *keyringSecureStorageManager {
...
+		account := strings.ToUpper(tokenSpec.user)
+		item := keyring.Item{
+			Key:  account,
+			Data: []byte(value),
+		}
+		if err := ring.Set(item); err != nil {
+			return fmt.Errorf("Failed to write to keychain. Err: %v", err)
 		}
 	}
+	return nil
```
sfc-gh-pfus: Giving that another thought - maybe we should return error that no supported credential manager found?

auth.go
```diff
@@ -403,13 +403,22 @@ func authenticate(
 		logger.WithContext(ctx).Errorln("Authentication FAILED")
```
sfc-gh-pfus, line 421: Hi Felipe! I generally agree with the direction of your changes (and I'm willing to merge it), but I'm wondering, why is it needed? Treating driver as a black-box, nothing changes, right?

secure_storage_manager.go
felipecrv: I'm just being extra careful with *new* keychain-related error messages.

felipecrv: But I can return the error here if it's ok.

secure_storage_manager.go
felipecrv: It should be a return here. I missed it.

secure_storage_manager.go
felipecrv: I'm going to wrap and return

auth.go
felipecrv: > why is it needed?

There has been many mysterious credentials storage related issues and ignoring errors makes it really difficult to debug. There are many distinct failure modes that were all converging to `""` being returned from `getCredential`.

> Treating driver as a black-box, nothing changes, right?

Nothing should change. I took care to not fail the auth process if credentials storage fails [1]. But unit tests got more strict: they will fail if the crendentials storage fails.

[1] changing this would happen in another PR if we consider necessary

secure_storage_manager.go
felipecrv: We can't. There is a common tail in the if/else that is de-duplicated by having this variable. I wish Go supported: `x := if ... else ...`.

I moved the debug-logging code to make it even more evident why this variable is a benefit.

secure_storage_manager.go
felipecrv: At the moment, only Windows and macOS use the keychain-based credentials storage.

sfc-gh-pfus: Exactly - so ending up in this line means that we have an exceptional situation, right? Otherwise this SSM is a NOOP for non-Windows/MacOS OSes and it should rather emit error, WDYT?

felipecrv: This line is unreachable.

sfc-gh-pfus: It currently is, because we create this SSM only for Windows and Mac. However, I'm not a fan of having "this is fine here, because it's unreachable because of somewhere else" - it's easy to change this "somewhere else" and forget this line. It would be better to return "should never happen" error here.

felipecrv: Added an error message return.

felipecrv:
@sfc-gh-pfus can you approve the CI runs? I'm curious to see if the more strict asserts are breaking in a full run of the test suite in all the OSes -- it's hard to test all that locally.

felipecrv:
@sfc-gh-pfus are the `go test -v ...` errors uploaded somewhere?

Screenshot 2025-07-24 at 10 52 18


UPDATE: a local run is giving me many errors already. Investigating.

sfc-gh-rsavenok:
@felipecrv running all tests locally requires wiremock process spawned and real snowflake account to execute some of the tests. Additionally, our GH actions always fail when someone submits a PR from a fork.

I triggered GH actions with correct permissions. Lets wait for them to complete and see if there are any failures caused by your PR :)

felipecrv:
> @felipecrv running all tests locally requires wiremock process spawned and real snowflake account to execute some of the tests.

Yeah. I've been running selected tests with it locally on macOS and Windows (sometimes).

> Additionally, our GH actions always fail when someone submits a PR from a fork.

Good to know. :)

sfc-gh-pfus:
> are the go test -v ... errors uploaded somewhere?

Sorry, we made a mistake with configuring codecov and it grabbed all logs from Linux and Mac machines. I fixed it yesterday.

sfc-gh-pfus:
@felipecrv can you rebase your change?

felipecrv:
I think this will fail again. I currently don't have the time to investigate.

sfc-gh-pfus:
Hi @felipecrv ! Are you still interested in merging this PR?

felipecrv:
> Hi @felipecrv ! Are you still interested in merging this PR?

Feel free to close.

PR #1473, opened 2025-07-21
NO-SNOW: adding wildcards tests in oauth

sfc-gh-pcyrek:
### Description

NO-SNOW: adding wildcards tests in oauth

### Checklist

### DIFF
--- a/auth_with_oauth_snowflake_authorization_code_wildcards_test.go
+++ b/auth_with_oauth_snowflake_authorization_code_wildcards_test.go
@@ -1,0 +1,147 @@
+package gosnowflake
+
+import (
+	"errors"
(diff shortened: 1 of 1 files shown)


auth_with_oauth_snowflake_authorization_code_wildcards_test.go
```diff
@@ -0,0 +1,111 @@
+package gosnowflake
```
sfc-gh-pfus, line 1: Why can't we move it to the existing file?

sfc-gh-pcyrek: to keep the same convention between drivers

PR #1474, opened 2025-07-21
fix: do not panic when `asb.scd` is nil

prsabahrami:
### Description
This is an attempt to fix the panic in the below case: 
I wouldn't say with full confidence that this is the source of the issue but it happened upon a single select query execution.
```go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x3f2dca8]

goroutine 593403 [running]:
github.com/snowflakedb/gosnowflake.(*ArrowStreamBatch).downloadChunkStreamHelper(0xc002bceed0, {0x6168b88, 0xc001572ff0})
	/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.14.0/connection.go:600 +0x228
github.com/snowflakedb/gosnowflake.(*ArrowStreamBatch).GetStream(...)
	/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.14.0/connection.go:655
github.com/apache/arrow-adbc/go/adbc/driver/snowflake.newRecordReader.func3.1()
	/go/pkg/mod/github.com/apache/arrow-adbc/go/adbc@v1.6.0/driver/snowflake/record_reader.go:655 +0x106
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/go/pkg/mod/golang.org/x/sync@v0.16.0/errgroup/errgroup.go:93 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 593359
	/go/pkg/mod/golang.org/x/sync@v0.16.0/errgroup/errgroup.go:78 +0x93
```

### Checklist
- [ ] Created tests which fail without the change (if possible) (not exactly sure ho

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -590,6 +590,13 @@ func (w *wrapReader) Close() error {
 }
 
 func (asb *ArrowStreamBatch) downloadChunkStreamHelper(ctx context.Context) error {
+	if asb == nil || asb.scd == nil {
+		return &SnowflakeError{
+			Number:   ErrNilArrowStreamBatch,
+			SQLState: SQLStateConnectionFailure,
+			Message:  errMsgNilArrowStreamBatch,
+		}
+	}
 	headers := make(map[string]string)
 	if len(asb.scd.ChunkHeader) > 0 {
 		logger.WithContext(ctx).Debug("chunk header is provided")
--- a/connection.go
+++ b/connection.go
@@ -597,6 +597,14 @@ func (asb *ArrowStreamBatch) downloadChunkStreamHelper(ctx context.Context) erro
 			Message:  errMsgNilArrowStreamBatch,
 		}
 	}
+	if asb.scd.ChunkMetas == nil || asb.idx >= len(asb.scd.ChunkMetas) {
+		return &SnowflakeError{
+			Number:      ErrFailedToGetChunk,
+			SQLState:    SQLStateConnectionFailure,
+			Message:     errMsgFailedToGetChunk,
+			MessageArgs: []interface{}{asb.idx},
+		}
+	}
 	headers := make(map[string]string)
 	if len(asb.scd.ChunkHeader) > 0 {
 		logger.WithContext(ctx).Debug("chunk header is provided")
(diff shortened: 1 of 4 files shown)


prsabahrami:
I have read the CLA Document and I hereby sign the CLA.

prsabahrami:
recheck

sfc-gh-pfus:
Hi @prsabahrami ! Is it a problem that you have a reproduction of? I'm leaning towards finding a root cause instead of transforming a panic into an error. This situation should not happen at all.

prsabahrami:
Hi @sfc-gh-pfus!
Yeah that's fair. It's not as easy to repro. However, for some context, the query that's running returns a high dimensional table with around 1m rows through Arrow (you can see adbc error's in the description) which naturally streams through Arrow batches so shouldn't really result in OOM but I guess the issue has something to do with that? Not sure if the fix for the panic should be on gosnowflake's end or adbc.
Another thing to note is that if I do not use adbc and directly write an arrow record reader of my own but still use gosnowflake driver, I would get OOMKilled.

sfc-gh-pfus:
How often does it happen? Is there a repro?

prsabahrami:
Okay I believe I found the issue. 
I'm getting OOM Killed (not a panic) when using `gosnowflake` directly without `adbc` but ig that actually has something to do with the actual memory limit cuz with pure `gosnowflake` and increased memory the query did work. 
I'm not sure if you'd be interested in investing some time in having a safeguard against OOMKilled when querying `ArrowBatches` but it should be interesting to consider. 
Regarding the panic, I believe `adbc` is here to blame and they somehow pass a nil object somewhere to the function. 
Also, an answer to ur question, it happens whenever I query a considerably high dimensional table. 

In short, I don't rly think there is much on ur side to do about this apart from perhaps considering adding a safeguard for OOMKilled. Would you be interested in doing that? 
I'll raise the panic to `adbc` people and keep u in the loop. 

@sfc-gh-pfus would love to know what you think.

sfc-gh-pfus:
Hi @prsabahrami , thanks for the explanation! Do you have a rough explanation of what this OOM safeguard would look like? Any doc/blog/anything?

sfc-gh-pfus:
Hi @prsabahrami ! Any comments?

sfc-gh-pfus:
Closing due to no response.

PR #1476, opened 2025-07-23
SNOW-2117152 Add CRLs to DSN

sfc-gh-pfus:
### Description

SNOW-2117152 Enable CRLs to be used with connections.

### Checklist

### DIFF
--- a/auth_wif.go
+++ b/auth_wif.go
@@ -304,7 +304,12 @@ func (c *gcpIdentityAttestationCreator) createTokenRequest() (*http.Request, err
 }
 
 func fetchTokenFromMetadataService(req *http.Request, cfg *Config) string {
-	client := &http.Client{Transport: getTransport(cfg)}
+	transport, err := getTransport(cfg)
+	if err != nil {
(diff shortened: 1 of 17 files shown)


auth_wif.go
```diff
@@ -304,7 +304,12 @@ func (c *gcpIdentityAttestationCreator) createTokenRequest() (*http.Request, err
 }
 
 func fetchTokenFromMetadataService(req *http.Request, cfg *Config) string {
-	client := &http.Client{Transport: getTransport(cfg)}
+	transport, err := getTransport(cfg)
+	if err != nil {
+		logger.Debugf("Failed to create HTTP transport: %v", err)
```
sfc-gh-rsavenok, line 309: Question: Why here you log failure, but in other files (auth_oath or azure storage client) it's just a return? 

Maybe just log inside getTransport function before returning an error?

sfc-gh-pfus: Because it has to be logged somewhere (when it is not propagated forward). In the remaining places we just propagate it. IIUC Dawid's intention around WIF was in case of problems we continue going to the next WIF provider (for the sake of autodiscovery).

PR #1486, opened 2025-07-25
NO-SNOW Fix aws config for WIF

sfc-gh-rsavenok:
### Description
`LoadWithConfig` resolves region using 3 resolvers:
```
resolveRegion,                
resolveEC2IMDSRegion,
resolveDefaultRegion, 
```

When running auth on EC2 with no region in config, we should fetch it from the metadata service (resolveEC2IMDSRegion), but it returns "" unless the config option is set:
```
if o.UseEC2IMDSRegion == nil {
  return "", false, nil
}
```

PR doesn't contain any unit test changes, as they used a mocked AWS provider and can't catch such issues. Future e2e tests will catch such issues.

### Checklist

### DIFF
--- a/auth_wif.go
+++ b/auth_wif.go
@@ -8,15 +8,16 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/aws/aws-sdk-go-v2/aws"
-	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
-	"github.com/aws/aws-sdk-go-v2/config"
-	"github.com/golang-jwt/jwt/v5"
 	"io"
 	"net/http"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/golang-jwt/jwt/v5"
 )
 
 const (
@@ -153,7 +154,7 @@ type defaultAwsAttestationMetadataProvider struct {
 }
 
 func createDefaultAwsAttestationMetadataProvider(ctx context.Context) *defaultAwsAttestationMetadataProvider {
-	cfg, err := config.LoadDefaultConfig(ctx)
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithEC2IMDSRegion())
 	if err != nil {
 		logger.Debugf("Unable to load AWS config: %v", err)
 		return nil


sfc-gh-rkowalski (approved):

sfc-gh-rsavenok:
@sfc-gh-pmansour 

> Can we also ensure it's tested on AWS serverless (e.g. Lambda)?
@sfc-gh-akolodziejczyk is going to write automatic tests for Lambda, so if it doesn't work there, there will be another PR. Lambda adds region as env variable and I checked that LoadWithConfig correctly reads env var

auth_test.go
```diff
@@ -1094,3 +1096,65 @@ func TestWithOAuthClientCredentialsFlowManual(t *testing.T) {
...
+	if os.Getenv("SKIP_SETUP") == "" {
+		t.Skip("Skipping test - meant to be run on remote VM only")
+	}
+	level := logger.GetLogLevel()
+	_ = logger.SetLogLevel("trace")
+	defer func() {
+		_ = logger.SetLogLevel(level)
+	}()
+	getWIFConfig := func() *Config {
+		config := &Config{}
```
sfc-gh-pfus: Why you don't want to initialize it immediately?

auth_test.go
```diff
@@ -1094,3 +1096,65 @@ func TestWithOAuthClientCredentialsFlowManual(t *testing.T) {
...
+func TestWorkloadIdentityOnRemoteVM(t *testing.T) {
+	if os.Getenv("SKIP_SETUP") == "" {
+		t.Skip("Skipping test - meant to be run on remote VM only")
+	}
+	level := logger.GetLogLevel()
+	_ = logger.SetLogLevel("trace")
+	defer func() {
+		_ = logger.SetLogLevel(level)
+	}()
+	getWIFConfig := func() *Config {
```
sfc-gh-pfus: This function doesn't have parameters, why do we need it as a function?

auth_test.go
```diff
@@ -1094,3 +1096,65 @@ func TestWithOAuthClientCredentialsFlowManual(t *testing.T) {
...
+	}
+	runWIFTest := func(t *testing.T, config *Config) {
+		logger.Debugf("Running WIF test with config: %+v", config)
+		connector := NewConnector(SnowflakeDriver{}, *config)
+		db := sql.OpenDB(connector)
+		defer db.Close()
+		runSmokeQuery(t, db)
+	}
+
+	t.Run("autodetected provider", func(t *testing.T) {
```
sfc-gh-pfus: We could rewrite everything to loop over configs? See examples with `range testcases`.

auth_test.go
sfc-gh-rsavenok: Because I need a fresh copy for every test case since I'm modifying it. But the suggestion of having a loop of test cases would change this anyways ;)

ci/test_wif.sh
```diff
@@ -0,0 +1,84 @@
...
+run_tests_and_set_result() {
+  local provider="$1"
+  local host="$2"
+  local snowflake_host="$3"
+  local rsa_key_path="$4"
+
+  # TODO: replace
+  # * docker image
+  # * remove apt-get
+  # * remove SF_ENABLE_EXPERIMENTAL_AUTHENTICATION=true
```
sfc-gh-rsavenok: This will be removed in PR that removes experimental flag

ci/test_wif.sh
```diff
@@ -0,0 +1,84 @@
...
+export PARAMETERS_FILE_PATH="$THIS_DIR/wif/parameters/parameters_wif.json"
+
+run_tests_and_set_result() {
+  local provider="$1"
+  local host="$2"
+  local snowflake_host="$3"
+  local rsa_key_path="$4"
+
+  # TODO: replace
+  # * docker image
```
sfc-gh-rsavenok: docker image and apt-get will be changed before merging PR

ci/test_wif.sh
```diff
@@ -0,0 +1,84 @@
...
+        --rm \
+        -e BRANCH \
+        -e SNOWFLAKE_TEST_WIF_PROVIDER \
+        -e SNOWFLAKE_TEST_WIF_HOST \
+        -e SNOWFLAKE_TEST_WIF_ACCOUNT \
+        -e SF_ENABLE_EXPERIMENTAL_AUTHENTICATION=true \
+        golang:1.22 \
+          bash -c "
+            apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
+            echo 'Running tests on branch: \$BRANCH, provider: \$SNOWFLAKE_TEST_WIF_PROVIDER'
```
sfc-gh-rsavenok, line 31: Changes compared to Java PR:
1) echo prining provider
2) no copy command after curl to fix branch names with slash e.g. `rsavenok/some-feature`

sfc-gh-pfus (approved):

PR #1492, opened 2025-07-29
SNOW-2229416 log session id for connection and heartbeat

sfc-gh-dszmolka:
### Description

This is coming from a live support case where we have `error: 390111: Session no longer exists.  New login required to access the service` back from Snowflake, and driver logs provide no clue at the moment, _which_ session does not exist anymore. So we cannot track it down what happened, when it was opened, was there any heartbeats made into it, when it was logout, etc.

This is a little bit inconvenient when tens of sessions are open in parallel and one of them is expired, there's no way to tell which, making debugging the issue hard. 

PR aims to use the already existing tokenaccessor to get the current sessionID and log it. This is not privileged or new information, we already log it a couple places (LOG_SESSION_ID). 

also introduced function to get the tokens in a more safe way even if for some reason `snowflakeRestful` itself, or its `TokenAccessor` would be `nil` on the connection (and log an error in such occasion)

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -180,12 +180,13 @@ var userAgent = fmt.Sprintf("%v/%v (%v-%v) %v/%v",
 	runtime.Version())
 
 type authRequestClientEnvironment struct {
-	Application     string `json:"APPLICATION"`
-	ApplicationPath string `json:"APPLICATION_PATH"`
-	Os              string `json:"OS"`
-	OsVersion       string `json:"OS_VERSION"`
-	OCSPMode        string `json:"OCSP_MODE"`
-	GoVersion       string `json:"GO_VERSION"`
+	Application             string `json:"APPLICATION"`
+	ApplicationPath         string `json:"APPLICATION_PATH"`
+	Os                      string `json:"OS"`
+	OsVersion               string `json:"OS_VERSION"`
+	OCSPMode                string `json:"OCSP_MODE"`
+	GoVersion               string `json:"GO_VERSION"`
+	CertRevocationCheckMode string `json:"CERT_REVOCATION_CHECK_MODE,omitempty"`
 }
 
 type authRequestData struct {
@@ -358,12 +359,13 @@ func authenticate(
 		applicationPath = "unknown"
 	}
 	clientEnvironment := authRequestClientEnvironment{
-		Application:     sc.cfg.Application,
-		ApplicationPath: applicationPath,
-		Os:              operatingSystem,
-		OsVersion:       platform,
-		OCSPMode:        sc.cfg.ocspMode(),
-		GoVersion:       runtime.Version(),
+		Application:             sc.cfg.Application,
+		ApplicationPath:         applicationPath,
+		Os:                      operatingSystem,
+		OsVersion:               platform,
+		OCSPMode:                sc.cfg.ocspMode(),
+		GoVersion:               runtime.Version(),
+		CertRevocationCheckMode: sc.cfg.CertRevocationCheckMode.String(),
 	}
 
 	sessionParameters := make(map[string]interface{})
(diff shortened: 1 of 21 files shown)


sfc-gh-pfus (approved):

heartbeat.go
```diff
@@ -21,41 +21,48 @@ type heartbeat struct {
...
+	ctx := context.WithValue(context.Background(), SFSessionIDKey, sessionID)
 	hbTicker := time.NewTicker(heartBeatInterval)
 	defer hbTicker.Stop()
 	for {
 		select {
 		case <-hbTicker.C:
 			err := hc.heartbeatMain()
 			if err != nil {
-				logger.Error("failed to heartbeat")
+				logger.WithContext(ctx).Error("failed to heartbeat")
```
sfc-gh-pbulawa: Could we log the error here?

heartbeat.go
```diff
@@ -21,41 +21,48 @@ type heartbeat struct {
...
 func (hc *heartbeat) heartbeatMain() error {
-	logger.Info("Heartbeating!")
 	params := &url.Values{}
 	params.Set(requestIDKey, NewUUID().String())
 	params.Set(requestGUIDKey, NewUUID().String())
 	headers := getHeaders()
-	token, _, _ := hc.restful.TokenAccessor.GetTokens()
+	token, _, sessionID := hc.restful.TokenAccessor.GetTokens()
+	ctx := context.WithValue(context.Background(), SFSessionIDKey, sessionID)
+	logger.WithContext(ctx).Info("Heartbeating!")
```
sfc-gh-pbulawa, line 65: IMO, there's no need to log the heartbeat on Info. WDYT about changing it to debug?

heartbeat.go
```diff
@@ -66,14 +73,15 @@ func (hc *heartbeat) heartbeatMain() error {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode == http.StatusOK {
-		logger.Infof("heartbeatMain: resp: %v", resp)
+		logger.WithContext(ctx).Infof("heartbeatMain: resp: %v", resp)
```
sfc-gh-pbulawa: What about changing it to debug?

heartbeat.go
```diff
@@ -83,11 +91,11 @@ func (hc *heartbeat) heartbeatMain() error {
 	}
 	b, err := io.ReadAll(resp.Body)
 	if err != nil {
-		logger.Errorf("failed to extract HTTP response body. err: %v", err)
+		logger.WithContext(ctx).Errorf("failed to extract HTTP response body. err: %v", err)
 		return err
 	}
-	logger.Infof("HTTP: %v, URL: %v, Body: %v", resp.StatusCode, fullURL, b)
-	logger.Infof("Header: %v", resp.Header)
+	logger.WithContext(ctx).Infof("HTTP: %v, URL: %v, Body: %v", resp.StatusCode, fullURL, b)
```
sfc-gh-pbulawa: IMO there's no need to log these details on info. WDYT about changing it to debug?

heartbeat.go
sfc-gh-dszmolka: `error` was not logged originally either, but this is a good idea, added

heartbeat.go
sfc-gh-dszmolka: again this is how the function looked before my change, logging on Info even today. I think it's good to keep on Info, especially that it's a very infrequent operation (at most, it happens every 15m but by default once per hour so not expecting a lot of traffic from it)

heartbeat.go
sfc-gh-dszmolka: my preference keeping current loglevel, as for 'Heartbeating!'. but if there's a strong feeling moving to debug from multiple people, i can of course change it

heartbeat.go
sfc-gh-dszmolka: changed both lines (HTTP: and Header:) to debug

heartbeat.go
sfc-gh-pbulawa: Ok, thanks.

heartbeat.go
sfc-gh-pbulawa: IMO, logging the response when the flow proceeds correctly is unnecessary, as the log does not provide any vital information for the user who is using the driver or support. We inform the user that heartbeat starts with the log `Heartbeating!`

sfc-gh-pbulawa (approved):

heartbeat.go
sfc-gh-dszmolka: this has been replaced with Debug level

sfc-gh-pfus (approved):

PR #1484, opened 2025-07-25
SNOW-2117152 Add CRL telemetry

sfc-gh-pfus:
### Description

SNOW-2117152 Added telemetry to CRL process.

### Checklist
- [x] Created tests which fail without the change (if possible)
- [x] Extended the README / documentation, if necessary

PR #1494, opened 2025-07-29
SNOW-2176203: Add RegisterTLSConfig support

sfc-gh-snoonan:
This PR adds support for custom TLS configurations while preserving Snowflake's certificate revocation checking (OCSP/CRL). Users can now register custom TLS configs with their own root CAs, client certificates, and verification callbacks, which will be automatically chained with Snowflake's security validation.

Configs each have a name and are registered; the `tlsConfigName` parameter in the DSN can be used to reference them.

### Usage Example

```go
import (
    "crypto/x509"
    "database/sql"
    _ "github.com/snowflakedb/gosnowflake"
)

// Register custom TLS config with your root CAs
rootCertPool := x509.NewCertPool()
pem, _ := os.ReadFile("/path/to/ca-cert.pem")
rootCertPool.AppendCertsFromPEM(pem)

err := gosnowflake.RegisterTLSConfig("myCustomTLS", &tls.Config{
    RootCAs: rootCertPool,
})
if err != nil {
    // handle error case...
}

// Use in connection string
db, err := sql.Open("snowflake", "user:pass@account/db?tlsConfigName=myCustomTLS")
```

### Checklist

### DIFF
--- a/.cursor/rules/overall-guidelines.mdc
+++ b/.cursor/rules/overall-guidelines.mdc
@@ -1,0 +1,26 @@
+# Cursor Rules for Go Snowflake Driver
+
+## General Development Standards
+
+### Code Quality
+- Follow Go formatting standards (use `gofmt`)
+- Use meaningful variable and function names
+- Include error handling for all operations that can fail
+- Write comprehensive documentation for public APIs
+
+### Project Structure
+- Place test files in the same package as the code being tested
+- Use `testdata/` directory for test fixtures and sample data
+- Group related functionality in logical packages
+
+### Testing
+- Test files should be named `*_test.go`
+- **For test-specific rules, see `testing.mdc`**
+- Write both positive and negative test cases
+- Use table-driven tests for testing multiple scenarios
+
+### Code Review Guidelines
+- Ensure code follows Go best practices
+- Verify comprehensive test coverage
+- Check that error messages are descriptive and helpful for debugging
+- Validate that public APIs are properly documented
--- a/.cursor/rules/overall-guidelines.mdc
+++ b/.cursor/rules/overall-guidelines.mdc
@@ -1,3 +1,33 @@
+---
+alwaysApply: true
+---
+
+# Cursor Rules for Go Snowflake Driver
+
+## General Development Standards
+
+### Code Quality
+- Follow Go formatting standards (use `gofmt`)
+- Use meaningful variable and function names
+- Include error handling for all operations that can fail
+- Write comprehensive documentation for public APIs
+
+### Project Structure
+- Place test files in the same package as the code being tested
+- Use `testdata/` directory for test fixtures and sample data
+- Group related functionality in logical packages
+
+### Testing
+- Test files should be named `*_test.go`
+- **For test-specific rules, see `testing.mdc`**
+- Write both positive and negative test cases
+- Use table-driven tests for testing multiple scenarios
+
+### Code Review Guidelines
+- Ensure code follows Go best practices
+- Verify comprehensive test coverage
+- Check that error messages are descriptive and helpful for debugging
+- Validate that public APIs are properly documented
 # Cursor Rules for Go Snowflake Driver
 
 ## General Development Standards
--- a/.cursor/rules/overall-guidelines.mdc
+++ b/.cursor/rules/overall-guidelines.mdc
@@ -14,33 +14,7 @@ alwaysApply: true
 
 ### Project Structure
 - Place test files in the same package as the code being tested
-- Use `testdata/` directory for test fixtures and sample data
-- Group related functionality in logical packages
-
-### Testing
-- Test files should be named `*_test.go`
-- **For test-specific rules, see `testing.mdc`**
-- Write both positive and negative test cases
-- Use table-driven tests for testing multiple scenarios
-
-### Code Review Guidelines
-- Ensure code follows Go best practices
-- Verify comprehensive test coverage
-- Check that error messages are descriptive and helpful for debugging
-- Validate that public APIs are properly documented
-# Cursor Rules for Go Snowflake Driver
-
-## General Development Standards
-
-### Code Quality
-- Follow Go formatting standards (use `gofmt`)
-- Use meaningful variable and function names
-- Include error handling for all operations that can fail
-- Write comprehensive documentation for public APIs
-
-### Project Structure
-- Place test files in the same package as the code being tested
-- Use `testdata/` directory for test fixtures and sample data
+- Use `test_data/` directory for test fixtures and sample data
 - Group related functionality in logical packages
 
 ### Testing
--- a/.cursor/rules/testing.mdc
+++ b/.cursor/rules/testing.mdc
@@ -1,0 +1,86 @@
+# Cursor Rules for Go Test Files
+
+This file automatically applies when working on `*_test.go` files.
+
+## Testing Standards
+
+### Assertion Helper Usage
+- **ALWAYS** use assertion helpers from `assert_test.go` instead of direct `t.Fatal`, `t.Fatalf`, `t.Error`, or `t.Errorf` calls
+- **NEVER** write manual if-then-fatal patterns in test functions
+
+#### Common Assertion Patterns:
+
+**Error Checking:**
+```go
+// ❌ WRONG
+if err != nil {
+    t.Fatalf("Unexpected error: %v", err)
+}
+
+// ✅ CORRECT  
+assertNilF(t, err, "Unexpected error")
+```
+
+**Nil Checking:**
+```go
+// ❌ WRONG
+if obj == nil {
+    t.Fatal("Expected non-nil object")
+}
+
+// ✅ CORRECT
+assertNotNilF(t, obj, "Expected non-nil object")
+```
+
+**Equality Checking:**
+```go
+// ❌ WRONG
+if actual != expected {
+    t.Fatalf("Expected %v, got %v", expected, actual)
+}
+
+// ✅ CORRECT
+assertEqualF(t, actual, expected, "Values should match")
+```
+
+**Error Message Validation:**
+```go
+// ❌ WRONG
+if err.Error() != expectedMsg {
+    t.Fatalf("Expected error: %s, got: %s", expectedMsg, err.Error())
+}
+
+// ✅ CORRECT
+assertEqualF(t, err.Error(), expectedMsg, "Error message should match")
+```
+
+**Boolean Assertions:**
+```go
+// ❌ WRONG
+if !condition {
+    t.Fatal("Condition should be true")
+}
+
+// ✅ CORRECT
+assertTrueF(t, condition, "Condition should be true")
+```
+
+#### Helper Function Reference:
+- `assertNilF/E(t, value, description)` - Assert value is nil
+- `assertNotNilF/E(t, value, description)` - Assert value is not nil  
+- `assertEqualF/E(t, actual, expected, description)` - Assert equality
+- `assertNotEqualF/E(t, actual, expected, description)` - Assert inequality
+- `assertTrueF/E(t, value, description)` - Assert boolean is true
+- `assertFalseF/E(t, value, description)` - Assert boolean is false
+- `assertStringContainsF/E(t, str, substring, description)` - Assert string contains substring
+- `assertErrIsF/E(t, actual, expected, description)` - Assert error matches expected error
+
+#### When to Use F vs E:
+- Use `F` suffix (Fatal) for critical failures that should stop the test immediately
+- Use `E` suffix (Error) for non-critical failures that allow the test to continue
+
+## Code Review Guidelines:
+- Flag any direct use of `t.Fatal*` or `t.Error*` in new code
+- Ensure all test functions use appropriate assertion helpers
+- Verify that error messages are descriptive and helpful for debugging
+- Check that tests are comprehensive and cover edge cases
--- a/.cursor/rules/testing.mdc
+++ b/.cursor/rules/testing.mdc
@@ -1,3 +1,7 @@
+---
+alwaysApply: true
+---
+
 # Cursor Rules for Go Test Files
 
 This file automatically applies when working on `*_test.go` files.
@@ -5,8 +9,8 @@ This file automatically applies when working on `*_test.go` files.
 ## Testing Standards
 
 ### Assertion Helper Usage
-- **ALWAYS** use assertion helpers from `assert_test.go` instead of direct `t.Fatal`, `t.Fatalf`, `t.Error`, or `t.Errorf` calls
-- **NEVER** write manual if-then-fatal patterns in test functions
+- **ALWAYS** Attempt to use assertion helpers from `assert_test.go` instead of direct `t.Fatal`, `t.Fatalf`, `t.Error`, or `t.Errorf` calls. Where it makes sense, add new assertion helpers.
+- **NEVER** write manual if-then-fatal patterns in test functions when a suitable assertion helper exists.
 
 #### Common Assertion Patterns:
 
@@ -66,6 +70,7 @@ assertTrueF(t, condition, "Condition should be true")
 ```
 
 #### Helper Function Reference:
+Always examine `assertion_helpers.go` for the latest set of helpers. Consider these existing examples below.
 - `assertNilF/E(t, value, description)` - Assert value is nil
 - `assertNotNilF/E(t, value, description)` - Assert value is not nil  
 - `assertEqualF/E(t, actual, expected, description)` - Assert equality
@@ -83,4 +88,4 @@ assertTrueF(t, condition, "Condition should be true")
 - Flag any direct use of `t.Fatal*` or `t.Error*` in new code
 - Ensure all test functions use appropriate assertion helpers
 - Verify that error messages are descriptive and helpful for debugging
-- Check that tests are comprehensive and cover edge cases
+- Check that tests are comprehensive and cover edge cases# Cursor Rules for Go Test Files
--- a/.cursor/rules/testing.mdc
+++ b/.cursor/rules/testing.mdc
@@ -81,7 +81,7 @@ Always examine `assertion_helpers.go` for the latest set of helpers. Consider th
 - `assertErrIsF/E(t, actual, expected, description)` - Assert error matches expected error
 
 #### When to Use F vs E:
-- Use `F` suffix (Fatal) for critical failures that should stop the test immediately
+- Use `F` suffix (Fatal) for critical failures that should stop the test immediately as well as for preconditions
 - Use `E` suffix (Error) for non-critical failures that allow the test to continue
 
 ## Code Review Guidelines:
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test_setup: test_teardown
 	python3 ci/scripts/hang_webserver.py 12345 &
 
 test_teardown:
-	kill -9 $$(ps -ef | grep hang_webserver | grep -v grep | awk '{print $$1}') || true
+	pkill -9 hang_webserver || true
 
 test: deps test_setup
 	./ci/scripts/execute_tests.sh
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,9 @@ lint: clint
 ## Format source codes
 fmt: cfmt
 	@for c in $$(ls cmd); do \
-		(cd cmd/$$c;  make fmt); \
+		if [ -f cmd/$$c/Makefile ]; then \
+			(cd cmd/$$c; make fmt); \
+		fi; \
 	done
 
 ## Install sample programs
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,7 @@ lint: clint
 ## Format source codes
 fmt: cfmt
 	@for c in $$(ls cmd); do \
-		if [ -f cmd/$$c/Makefile ]; then \
-			(cd cmd/$$c; make fmt); \
-		fi; \
+		(cd cmd/$$c; make fmt); \
 	done
 
 ## Install sample programs
--- a/assert_test.go
+++ b/assert_test.go
@@ -325,18 +325,3 @@ func assertTLSConfigsEqual(t *testing.T, expected, actual *tls.Config, msg strin
 	actualHasVerifier := actual.VerifyPeerCertificate != nil
 	assertEqualF(t, expectedHasVerifier, actualHasVerifier, "%s VerifyPeerCertificate presence", msg)
 }
-
-// assertTransportHasVerifier checks if transport has a TLS verifier function
-func assertTransportHasVerifier(t *testing.T, rt http.RoundTripper, shouldHave bool, msg string) {
-	transport := castToTransport(rt)
-	assertNotNilF(t, transport, "%s - expected http.Transport", msg)
-
-	if shouldHave {
-		assertNotNilF(t, transport.TLSClientConfig, "%s - expected TLSClientConfig", msg)
-		assertNotNilF(t, transport.TLSClientConfig.VerifyPeerCertificate, "%s - expected VerifyPeerCertificate", msg)
-	} else {
-		if transport.TLSClientConfig != nil {
-			assertNilF(t, transport.TLSClientConfig.VerifyPeerCertificate, "%s - expected no VerifyPeerCertificate", msg)
-		}
-	}
-}
--- a/assert_test.go
+++ b/assert_test.go
@@ -2,10 +2,8 @@ package gosnowflake
 
 import (
 	"bytes"
-	"crypto/tls"
 	"errors"
 	"fmt"
-	"net/http"
 	"reflect"
 	"regexp"
 	"runtime"
@@ -269,59 +267,4 @@ func thrownFrom() string {
 		}
 	}
 	return stack
-}
-
-// castToTransport safely casts http.RoundTripper to *http.Transport
-// Returns nil if the cast fails
-func castToTransport(rt http.RoundTripper) *http.Transport {
-	if transport, ok := rt.(*http.Transport); ok {
-		return transport
-	}
-	return nil
-}
-
-// assertTransportsEqual compares two transports, excluding function fields and other non-comparable fields
-// that may vary between instances but represent equivalent configurations
-func assertTransportsEqual(t *testing.T, expected, actual *http.Transport, msg string) {
-	if expected == nil && actual == nil {
-		return
-	}
-	assertNotNilF(t, expected, "Expected transport should not be nil in %s", msg)
-	assertNotNilF(t, actual, "Actual transport should not be nil in %s", msg)
-
-	// Compare TLS configurations
-	assertTLSConfigsEqual(t, expected.TLSClientConfig, actual.TLSClientConfig, msg+" TLS config")
-
-	// Compare other relevant transport fields (excluding function fields)
-	assertEqualF(t, expected.MaxIdleConns, actual.MaxIdleConns, "%s MaxIdleConns", msg)
-	assertEqualF(t, expected.MaxIdleConnsPerHost, actual.MaxIdleConnsPerHost, "%s MaxIdleConnsPerHost", msg)
-	assertEqualF(t, expected.MaxConnsPerHost, actual.MaxConnsPerHost, "%s MaxConnsPerHost", msg)
-	assertEqualF(t, expected.IdleConnTimeout, actual.IdleConnTimeout, "%s IdleConnTimeout", msg)
-	assertEqualF(t, expected.ResponseHeaderTimeout, actual.ResponseHeaderTimeout, "%s ResponseHeaderTimeout", msg)
-	assertEqualF(t, expected.ExpectContinueTimeout, actual.ExpectContinueTimeout, "%s ExpectContinueTimeout", msg)
-	assertEqualF(t, expected.TLSHandshakeTimeout, actual.TLSHandshakeTimeout, "%s TLSHandshakeTimeout", msg)
-	assertEqualF(t, expected.DisableKeepAlives, actual.DisableKeepAlives, "%s DisableKeepAlives", msg)
-	assertEqualF(t, expected.DisableCompression, actual.DisableCompression, "%s DisableCompression", msg)
-	assertEqualF(t, expected.ForceAttemptHTTP2, actual.ForceAttemptHTTP2, "%s ForceAttemptHTTP2", msg)
-}
-
-// assertTLSConfigsEqual compares two TLS configurations, excluding function fields
-// like VerifyPeerCertificate which may point to different but equivalent functions
-func assertTLSConfigsEqual(t *testing.T, expected, actual *tls.Config, msg string) {
-	if expected == nil && actual == nil {
-		return
-	}
-	assertNotNilF(t, expected, "Expected TLS config should not be nil in %s", msg)
-	assertNotNilF(t, actual, "Actual TLS config should not be nil in %s", msg)
-
-	// Compare non-function fields
-	assertEqualF(t, expected.InsecureSkipVerify, actual.InsecureSkipVerify, "%s InsecureSkipVerify", msg)
-	assertEqualF(t, expected.ServerName, actual.ServerName, "%s ServerName", msg)
-	assertEqualF(t, expected.MinVersion, actual.MinVersion, "%s MinVersion", msg)
-	assertEqualF(t, expected.MaxVersion, actual.MaxVersion, "%s MaxVersion", msg)
-
-	// For VerifyPeerCertificate, just check presence/absence since function pointers can't be compared
-	expectedHasVerifier := expected.VerifyPeerCertificate != nil
-	actualHasVerifier := actual.VerifyPeerCertificate != nil
-	assertEqualF(t, expectedHasVerifier, actualHasVerifier, "%s VerifyPeerCertificate presence", msg)
 }
--- a/assert_test.go
+++ b/assert_test.go
@@ -2,8 +2,10 @@ package gosnowflake
 
 import (
 	"bytes"
+	"crypto/tls"
 	"errors"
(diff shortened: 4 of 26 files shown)


Makefile
```diff
@@ -30,7 +30,9 @@ lint: clint
 ## Format source codes
 fmt: cfmt
 	@for c in $$(ls cmd); do \
-		(cd cmd/$$c;  make fmt); \
+		if [ -f cmd/$$c/Makefile ]; then \
```
sfc-gh-pfus: Why do we need this change? I suppose we should rather add missing makefiles to have everything formatted instead?

connection.go
```diff
@@ -797,32 +796,19 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
-			// set OCSP fail open mode
-			ocspResponseCacheLock.Lock()
-			atomic.StoreUint32((*uint32)(&ocspFailOpen), uint32(sc.cfg.OCSPFailOpen))
-			ocspResponseCacheLock.Unlock()
-		}
-	} else {
-		// use the custom transport
-		st = sc.cfg.Transporter
+
+	transportFactory := NewTransportFactory(sc.cfg)
```
sfc-gh-pfus: Do we need this function to be exported? Is it supposed to be used by clients?

connection.go
```diff
@@ -797,32 +796,19 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
+	st, cv, err := transportFactory.createTransport()
```
sfc-gh-pfus: Please, wait with this PR until I merge CRL telemetry, as mine is a bit more urgent and there will be some conflicts.

connection.go
```diff
@@ -880,27 +866,6 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
-		return transport, nil
-	}
-	if cfg.DisableOCSPChecks || cfg.InsecureMode {
-		logger.Debug("getTransport: skipping OCSP validation for cloud storage")
-		return snowflakeNoRevocationCheckTransport, nil
-	}
-	logger.Debug("getTransport: will perform OCSP validation for cloud storage")
-	return SnowflakeTransport, nil
+	transportFactory := NewTransportFactory(cfg)
+	return transportFactory.CreateFileTransferTransport()
```
sfc-gh-pfus: The original function name `getTransport` doesn't mention anything about using it for file transport only. Especially it is used by external authenticators like SAML or OAuth as well.

dsn.go
```diff
@@ -101,6 +101,8 @@ type Config struct {
 
 	Transporter http.RoundTripper // RoundTripper to intercept HTTP requests and responses
 
+	TLSConfig string // TLS configuration name registered with RegisterTLSConfig
```
sfc-gh-pfus: There is a discrepancy between DSN and config names (`tls` vs `TLSConfig`) that serve the same purpose. My suggestions are:
1) Use the same name in both places (wdyt about changing it to `tlsConfigName`? unless it is a standard name (answer that only MySQL uses this name is not satisfactory :) ))
2) Use string in DSN with some name (`tls`/`tlsConfigName`), but in Config struct keep already `*tls.Config` instance. Customers using `Config` instead of DSN can omit registration.
3) As in 2), but in Config keep two fields - one string and one struct, return error on conflict.

I like the 2) option the most.

dsn_test.go
```diff
@@ -1203,6 +1203,42 @@ func TestParseDSN(t *testing.T) {
 			},
 			ocspMode: ocspModeFailOpen,
 		},
+		// TLS config test cases
```
sfc-gh-pfus: Does it work without registering? I'm surprised a bit, I would expect that DSN parser should already verify it. But maybe this is connected to my previous comment (keeping `*tls.Config` instance as soon as possible).

tls_config.go
```diff
@@ -0,0 +1,82 @@
...
+	"sync"
+)
+
+var (
+	tlsConfigLock     sync.RWMutex
+	tlsConfigRegistry map[string]*tls.Config
+)
+
+func init() {
+	tlsConfigRegistry = make(map[string]*tls.Config)
```
sfc-gh-pfus, line 10: Can't we init it on var declaration?

tls_config.go
```diff
@@ -0,0 +1,82 @@
...
+//	    log.Fatal("Failed to append PEM.")
+//	}
+//
+//	gosnowflake.RegisterTLSConfig("custom", &tls.Config{
+//	    RootCAs: rootCertPool,
+//	    // OCSP/CRL validation will be automatically added by the driver
+//	})
+//
+//	db, err := sql.Open("snowflake", "user:pass@account/db?tls=custom")
+func RegisterTLSConfig(key string, config *tls.Config) error {
```
sfc-gh-pfus, line 15: I'm not sure if having a pointer here is ok. It's fine to keep pointers in map, but it would be better to receive this param as a value. Otherwise, it can be changed later, without our control.

sfc-gh-pfus, line 15: WDYT about some logging here?

tls_config.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	tlsConfigLock.Lock()
+	tlsConfigRegistry[key] = config
+	tlsConfigLock.Unlock()
+	return nil
+}
+
+// DeregisterTLSConfig removes the tls.Config associated with key.
+func DeregisterTLSConfig(key string) {
```
sfc-gh-pfus: This is a public function, so once defined its contract, we won't be able to change it. Maybe it's good to return error, even if now it is always nil.

tls_config.go
```diff
@@ -0,0 +1,82 @@
...
+	tlsConfigLock.Lock()
+	delete(tlsConfigRegistry, key)
+	tlsConfigLock.Unlock()
+}
+
+// getTLSConfigClone returns a clone of a TLS config from the registry.
+// This prevents the user config from being modified by the driver.
+func getTLSConfigClone(key string) (*tls.Config, bool) {
```
sfc-gh-pfus: Ah, so this is what you planned. I like my suggestion above more, WDYT?

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"sync"
+	"testing"
+)
+
+func TestRegisterTLSConfig(t *testing.T) {
+	// Clean up any existing configs before testing
```
sfc-gh-pfus: Before? It cleans up after.

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	testConfig := &tls.Config{
+		InsecureSkipVerify: true,
+		ServerName:         "test-server",
+	}
+
+	// Test successful registration
+	err := RegisterTLSConfig("test", testConfig)
+	if err != nil {
+		t.Fatalf("RegisterTLSConfig failed: %v", err)
```
sfc-gh-pfus: We use some assertion helpers, can you use them?

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+	clone, exists := getTLSConfigClone("test")
+	if !exists {
+		t.Fatal("TLS config was not registered")
+	}
+
+	// Verify the clone is correct but not the same object (to prevent modification)
+	if clone == testConfig {
+		t.Fatal("getTLSConfigClone should return a clone, not the original")
+	}
+	if clone.InsecureSkipVerify != testConfig.InsecureSkipVerify {
```
sfc-gh-pfus: There is a `DeepEqual` function to compare the whole object, maybe we can use it? If yes, can we add it to our helpers?

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+		tlsConfigLock.Unlock()
+	}()
+
+	// Test concurrent access to ensure thread safety
+	var wg sync.WaitGroup
+	numGoroutines := 10
+
+	// Concurrently register configs
+	for i := 0; i < numGoroutines; i++ {
+		wg.Add(1)
```
sfc-gh-pfus: If the number of goroutines is constant, you can `add` before the loop.

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+func TestGetTLSConfigCloneNonExistent(t *testing.T) {
+	_, exists := getTLSConfigClone("nonexistent")
+	if exists {
+		t.Fatal("getTLSConfigClone should return false for non-existent config")
+	}
+}
+
+func TestTLSConfigConcurrency(t *testing.T) {
```
sfc-gh-pfus: What is the purpose of this test? Detect races?

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+			if !exists {
+				t.Errorf("Config test%d should exist", index)
+			}
+		}(i)
+	}
+
+	wg.Wait()
+}
+
+func TestDSNParsingWithTLSConfig(t *testing.T) {
```
sfc-gh-pfus: We have all tests related to DSN/Config in dsn_test.go.

dsn_test.go
```diff
@@ -1203,6 +1203,42 @@ func TestParseDSN(t *testing.T) {
...
+				ValidateDefaultParameters: ConfigBoolTrue,
+				ClientTimeout:             defaultClientTimeout,
+				JWTClientTimeout:          defaultJWTClientTimeout,
+				ExternalBrowserTimeout:    defaultExternalBrowserTimeout,
+				CloudStorageTimeout:       defaultCloudStorageTimeout,
+				IncludeRetryReason:        ConfigBoolTrue,
+			},
+			ocspMode: ocspModeFailOpen,
+			err:      nil,
+		},
```
sfc-gh-pfus, line 1209: There is the reversed direction test: config -> DSN. It should also be extended.

tls_config_test.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	clone, _ := getTLSConfigClone("insecure")
+	if clone.InsecureSkipVerify != false {
+		t.Fatal("Config should have been overwritten")
+	}
+}
+
+// Test that custom TLS configs preserve OCSP validation
+func TestRegisterTLSConfigWithOCSPValidation(t *testing.T) {
```
sfc-gh-pfus: What is the purpose of this test?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+// TransportConfig holds the configuration for creating HTTP transports
+type TransportConfig struct {
+	MaxIdleConns    int
+	IdleConnTimeout time.Duration
+	DialTimeout     time.Duration
+	KeepAlive       time.Duration
+}
+
+// DefaultTransportConfig returns the standard transport configuration
+func DefaultTransportConfig() TransportConfig {
```
sfc-gh-pfus: Why not return a pointer?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+	return &TransportFactory{config: config}
+}
+
+// createBaseTransport creates a base HTTP transport with the given configuration
+func (tf *TransportFactory) createBaseTransport(transportConfig TransportConfig, tlsConfig *tls.Config) *http.Transport {
+	dialer := &net.Dialer{
+		Timeout: transportConfig.DialTimeout,
+	}
+	if transportConfig.KeepAlive > 0 {
+		dialer.KeepAlive = transportConfig.KeepAlive
```
sfc-gh-pfus: The default value is 0, so we can omit the if, right?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// CreateCRLTransport creates a transport with CRL validation
+func (tf *TransportFactory) CreateCRLTransport() (*http.Transport, error) {
+	transport, _, err := tf.createCRLTransportInternal()
+	return transport, err
+}
+
+// createCRLTransportInternal creates a transport with CRL validation and returns the validator
+func (tf *TransportFactory) createCRLTransportInternal() (*http.Transport, *crlValidator, error) {
```
sfc-gh-pfus: Why can't we inline this function?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+	return transport, nil
+}
+
+// createStandardTransportInternal creates a transport without custom TLS configuration and returns the validator
+func (tf *TransportFactory) createStandardTransportInternal() (http.RoundTripper, *crlValidator, error) {
+	// Validate configuration
+	if err := tf.validateRevocationConfig(); err != nil {
+		return nil, nil, err
+	}
+
```
sfc-gh-pfus, line 188: Should we handle transport passed by a client as well?

sfc-gh-pfus, line 128: Should we handle transport passed by a client as well?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+	// Handle CRL validation path
+	if tf.config.CertRevocationCheckMode != CertRevocationCheckDisabled {
```
sfc-gh-pfus, line 130: Same here re OCSP check.

transport.go
```diff
@@ -0,0 +1,291 @@
...
+		return tf.createCustomTLSTransportInternal(customTLSConfig)
+	}
+
+	// Handle standard transport configuration
+	return tf.createStandardTransportInternal()
+}
+
+// CreateFileTransferTransport creates a transport for file transfer operations
+// This replaces the getTransport function
+func (tf *TransportFactory) CreateFileTransferTransport() (http.RoundTripper, error) {
```
sfc-gh-pfus: Why is this different than regular transport? Customer configure driver once (OCSP, CRL, custom transport, TLS Config etc), why do we need distinction internally?

transport.go
```diff
@@ -0,0 +1,291 @@
...
+	if !tf.config.DisableOCSPChecks && !tf.config.InsecureMode && tf.config.CertRevocationCheckMode != CertRevocationCheckDisabled {
+		return errors.New("both OCSP and CRL cannot be enabled at the same time, please disable one of them")
+	}
+	return nil
+}
+
+// chainVerificationCallbacks chains a user's custom verification with the provided verification function
+func (tf *TransportFactory) chainVerificationCallbacks(customTLSConfig *tls.Config, verificationFunc func([][]byte, [][]*x509.Certificate) error) {
+	if customTLSConfig.VerifyPeerCertificate == nil {
+		customTLSConfig.VerifyPeerCertificate = verificationFunc
```
sfc-gh-pfus: Maybe instead of configuring object we could just return the final function?

Makefile
sfc-gh-snoonan: Or we could just do `gofmt -w .`

I just wanted to stop seeing the errors when I ran make :-). The main problem is that there are subdirs under cmd/arrow. I'll do something else here.

connection.go
sfc-gh-snoonan: Unexported 🤦

transport.go
sfc-gh-snoonan: It isn't, and I've removed it.

dsn.go
sfc-gh-snoonan: I agree, 2) makes the most sense.

tls_config.go
sfc-gh-snoonan: tls.Config has a mutex in it, so we can't store the value trivially.

tls_config.go
sfc-gh-snoonan: SGTM

tls_config_test.go
sfc-gh-snoonan: Yeah. It isn't really necessary if we trust how the mutex is used. I've removed it.

dsn_test.go
sfc-gh-snoonan: It doesn't now that we've converged on *tls.Config as soon as possible. This also means we can't transform back and forth. Is that okay?

transport.go
sfc-gh-snoonan: Could you clarify what you mean?

dsn.go
sfc-gh-snoonan: The one issue I have here is that we can't easily go back and forth between the DSN and the Config if we don't have a key here (to your point about adding the DSN tests.) WDYT?

Makefile
sfc-gh-snoonan: Added a makefile that recurses properly.

dsn.go
sfc-gh-pfus: Ah, you are right! So we can keep it as you implemented, but with aligned names.

transport.go
sfc-gh-pfus: `Config` struct allows clients to pass their own `Transporter` - it should have a precedence over own transports.

sfc-gh-snoonan: Ah. We do handle this - we get passed in the config, and we always check config.Transporter first.

tls_config_test.go
sfc-gh-snoonan: Done, and added some cursor rules to the repo to have it do that when generating tests in the future.

connection.go
sfc-gh-snoonan: I assume this is https://github.com/snowflakedb/gosnowflake/pull/1484 right?

.cursor/rules/testing.mdc
```diff
@@ -0,0 +1,86 @@
+# Cursor Rules for Go Test Files
+
+This file automatically applies when working on `*_test.go` files.
+
+## Testing Standards
+
+### Assertion Helper Usage
+- **ALWAYS** use assertion helpers from `assert_test.go` instead of direct `t.Fatal`, `t.Fatalf`, `t.Error`, or `t.Errorf` calls
```
sfc-gh-pfus: I'm not a cursor expert, but ALWAYS is very strong, we don't have assertions for everything. Can it be relaxed to "wherever possible"? Or Cursor is smart enough to add a new assertion if missing?

.cursor/rules/testing.mdc
```diff
@@ -0,0 +1,86 @@
...
+- **NEVER** write manual if-then-fatal patterns in test functions
```
sfc-gh-pfus: Can you explain what you mean? I may not be as smart as cursor, but I'm not sure how to understand it :)

.cursor/rules/testing.mdc
```diff
@@ -0,0 +1,86 @@
...
+**Boolean Assertions:**
+```go
+// ❌ WRONG
+if !condition {
+    t.Fatal("Condition should be true")
+}
+
+// ✅ CORRECT
+assertTrueF(t, condition, "Condition should be true")
+```
```
sfc-gh-pfus, line 70: Do we need to extend this on every new assertion added?

.cursor/rules/testing.mdc
```diff
@@ -0,0 +1,86 @@
...
+- `assertNotNilF/E(t, value, description)` - Assert value is not nil  
+- `assertEqualF/E(t, actual, expected, description)` - Assert equality
+- `assertNotEqualF/E(t, actual, expected, description)` - Assert inequality
+- `assertTrueF/E(t, value, description)` - Assert boolean is true
+- `assertFalseF/E(t, value, description)` - Assert boolean is false
+- `assertStringContainsF/E(t, str, substring, description)` - Assert string contains substring
+- `assertErrIsF/E(t, actual, expected, description)` - Assert error matches expected error
+
+#### When to Use F vs E:
+- Use `F` suffix (Fatal) for critical failures that should stop the test immediately
```
sfc-gh-pfus: And for preconditions.

assert_test.go
```diff
@@ -268,3 +270,58 @@ func thrownFrom() string {
...
+func castToTransport(rt http.RoundTripper) *http.Transport {
+	if transport, ok := rt.(*http.Transport); ok {
+		return transport
+	}
+	return nil
+}
+
+// assertTransportsEqual compares two transports, excluding function fields and other non-comparable fields
+// that may vary between instances but represent equivalent configurations
+func assertTransportsEqual(t *testing.T, expected, actual *http.Transport, msg string) {
```
sfc-gh-pfus: Can we move all this to `transport_test.go`?

assert_test.go
```diff
@@ -268,3 +270,58 @@ func thrownFrom() string {
...
+	assertEqualF(t, expected.ExpectContinueTimeout, actual.ExpectContinueTimeout, "%s ExpectContinueTimeout", msg)
+	assertEqualF(t, expected.TLSHandshakeTimeout, actual.TLSHandshakeTimeout, "%s TLSHandshakeTimeout", msg)
+	assertEqualF(t, expected.DisableKeepAlives, actual.DisableKeepAlives, "%s DisableKeepAlives", msg)
+	assertEqualF(t, expected.DisableCompression, actual.DisableCompression, "%s DisableCompression", msg)
+	assertEqualF(t, expected.ForceAttemptHTTP2, actual.ForceAttemptHTTP2, "%s ForceAttemptHTTP2", msg)
+}
+
+// assertTLSConfigsEqual compares two TLS configurations, excluding function fields
+// like VerifyPeerCertificate which may point to different but equivalent functions
+func assertTLSConfigsEqual(t *testing.T, expected, actual *tls.Config, msg string) {
```
sfc-gh-pfus: And this to tls_config_test.go?

assert_test.go
```diff
@@ -268,3 +270,58 @@ func thrownFrom() string {
 	}
 	return stack
 }
+
+// castToTransport safely casts http.RoundTripper to *http.Transport
+// Returns nil if the cast fails
+func castToTransport(rt http.RoundTripper) *http.Transport {
```
sfc-gh-pfus: Is this the best file for this function?

connection.go
sfc-gh-pfus: Correct, you can merge now!

connection_test.go
```diff
@@ -887,64 +887,84 @@ func (t EmptyTransporter) RoundTrip(*http.Request) (*http.Response, error) {
...
-	}
-	crlTransport, _, err := createCrlTransport(crlCfg)
-	assertNilF(t, err)
 	testcases := []struct {
-		name      string
-		cfg       *Config
-		transport http.RoundTripper
+		name           string
+		cfg            *Config
+		transportCheck func(transport http.RoundTripper) bool
```
sfc-gh-pfus: Maybe instead of returning bool, we could pass `t` and do meaningful assertions?

connection_test.go
```diff
@@ -887,64 +887,84 @@ func (t EmptyTransporter) RoundTrip(*http.Request) (*http.Response, error) {
...
 	}{
 		{
-			name:      "DisableOCSPChecks and InsecureMode false",
-			cfg:       &Config{Account: "one", DisableOCSPChecks: false, InsecureMode: false},
-			transport: SnowflakeTransport,
+			name: "DisableOCSPChecks and InsecureMode false",
+			cfg:  &Config{Account: "one", DisableOCSPChecks: false, InsecureMode: false},
+			transportCheck: func(transport http.RoundTripper) bool {
+				// We should have a verifier function
+				t := castToTransport(transport)
```
sfc-gh-pfus: In case of my previous comments, this function would be obsolete.

connectivity_diagnosis_test.go
```diff
@@ -731,10 +726,12 @@ func TestPerformDiagnosis(t *testing.T) {
 
 		// perform the diagnosis without downloading CRL
 		performDiagnosis(config, false)
+		t.Logf("after performDiagnosis")
```
sfc-gh-pfus: Is it still needed?

dsn.go
```diff
@@ -101,6 +102,9 @@ type Config struct {
 
 	Transporter http.RoundTripper // RoundTripper to intercept HTTP requests and responses
 
+	TLSConfig     *tls.Config // Custom TLS configuration
```
sfc-gh-pfus: Do we still want it? Even if it is not serializable? IMHO we could keep it for clients that use `NewConnector`, but we should return error if `DSN` function is used and this var is set, wdyt? Moreover, what do we do in case of conflicts?

dsn_test.go
```diff
@@ -2359,3 +2380,75 @@ func TestUrlDecodeIfNeededE2E(t *testing.T) {
...
+	tlsConfigLock.Unlock()
+
+	// Register test TLS config
+	testTLSConfig := tls.Config{
+		InsecureSkipVerify: true,
+		ServerName:         "custom.test.com",
+	}
+	err := RegisterTLSConfig("custom", &testTLSConfig)
+	if err != nil {
+		t.Fatalf("Failed to register test TLS config: %v", err)
```
sfc-gh-pfus: `assertNilF`?

dsn_test.go
```diff
@@ -2359,3 +2380,75 @@ func TestUrlDecodeIfNeededE2E(t *testing.T) {
...
+	}
+	defer func() {
+		err := DeregisterTLSConfig("custom")
+		if err != nil {
```
sfc-gh-pfus: Same.

tls_config.go
```diff
@@ -0,0 +1,41 @@
...
+var (
+	tlsConfigLock     sync.RWMutex
+	tlsConfigRegistry = make(map[string]*tls.Config)
+)
+
+// RegisterTLSConfig registers a custom tls.Config to be used with sql.Open.
+// Use the key as a value in the DSN where tls=value.
+func RegisterTLSConfig(key string, config *tls.Config) error {
+	tlsConfigLock.Lock()
+	logger.Debugf("Registering TLS config for key: %s", key)
```
sfc-gh-pfus: This is a direct user action, I believe it's worth having `info`. Deregistering the same.

transport.go
```diff
@@ -0,0 +1,199 @@
...
+		tf.config.CrlOnDiskCacheDisabled,
+		tf.config.CrlOnDiskCacheDir,
+		tf.config.CrlOnDiskCacheRemovalDelay,
+		client,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return crlValidator, nil
```
sfc-gh-pfus: The last couple of lines can be merged to `return newCrlValidator(...)`.

transport.go
```diff
@@ -0,0 +1,199 @@
...
+	return tf.createBaseTransport(defaultTransportConfig(), tlsConfig)
+}
+
+// createNoRevocationTransport creates a transport without certificate revocation checking
+func (tf *transportFactory) createNoRevocationTransport() *http.Transport {
+	return tf.createBaseTransport(defaultTransportConfig(), nil)
+}
+
+// createCRLTransport creates a transport with CRL validation
+func (tf *transportFactory) createCRLValidator() (*crlValidator, error) {
```
sfc-gh-pfus, line 95: Should it be `transportFactory` responsibility to create CRL validator?

transport.go
```diff
@@ -0,0 +1,199 @@
...
+		client,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return crlValidator, nil
+}
+
+// TODO(snoonan): Better interface for crlValidator return
```
sfc-gh-pfus: Maybe we could have our custom `RoundTripper` implementation for CRLs, that contains (and expose?) crlValidator?

transport.go
```diff
@@ -0,0 +1,199 @@
...
+			return err
+		}
+		// Then run the provided verification
+		return verificationFunc(rawCerts, verifiedChains)
+	}
+	return newVerify
+}
+
+// getConfigDuration returns the config duration if non-zero, otherwise returns the default
+func getConfigDuration(configValue, defaultValue time.Duration) time.Duration {
```
sfc-gh-pfus: It's easier to write `cmp.Or(configValue, defaultValue)` and this function would be obsolete.

.cursor/rules/testing.mdc
sfc-gh-snoonan: As I understand it, the always here is more "include this in your model context" as opposed to a strict interpretation the way you or I would read it.

.cursor/rules/testing.mdc
sfc-gh-snoonan: No, but I've made that clearer.
FWIW, the file itself was originally generated by cursor.

.cursor/rules/testing.mdc
sfc-gh-snoonan: Ah, I was mistaken. That's the 'alwaysApply' bit I added above. This is adding to the prompt/context.

connectivity_diagnosis_test.go
sfc-gh-snoonan: I've removed it, but performDiagnosis has a logger.Fatalf in it, which is a frustrating experience. I can fix it in a follow-up PR.

transport.go
sfc-gh-snoonan: Thanks!

transport.go
sfc-gh-snoonan: Yeah, I've been struggling to think of the right abstraction here. I do think it provides a nicer interface if we isolate this within the connection/transport, as that's where we want retry policies and things eventually.

In fact, the only reason we're returning it is addressed by https://github.com/snowflakedb/gosnowflake/pull/1500 - I didn't want to move the cache cleanup start into here, and I was making the minimal changes (lol, not so minimal now...)

dsn.go
sfc-gh-snoonan: I think we still want it as a private field for ease of using the config, but I'm also okay with just keeping the registry/name. (I like the behavior that if someone does something silly like call deregister after we've started, we don't break.)

transport.go
sfc-gh-snoonan: I think so with the rewrite. Reopen if you disagree.

connection_test.go
sfc-gh-snoonan: Moved to passing t and performing assertions. If you prefer, I can just split these into separate top level tests.

connection.go
```diff
@@ -884,28 +862,10 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
...
-		if err != nil {
-			return nil, err
-		}
-		return transport, nil
-	}
-	if cfg.DisableOCSPChecks || cfg.InsecureMode {
-		logger.Debug("getTransport: skipping OCSP validation")
-		return snowflakeNoRevocationCheckTransport, nil
+	transportFactory := newTransportFactory(cfg, telemetry)
+	st, err := transportFactory.createTransport()
```
sfc-gh-pfus: Just `return transportFactory.createTransport()`?

connection_test.go
```diff
@@ -886,66 +886,103 @@ func (t EmptyTransporter) RoundTrip(*http.Request) (*http.Response, error) {
...
-			assertNilE(t, err)
-			if test.name == "Using CRLs" {
-				// we can't use default comparison for transport here, because this type is not comparable
-				// it works for other cases, because they use the same pointer
-				// for this case, we assume that only CRL transport uses such a small number of idle conns
-				assertEqualE(t, result.(*http.Transport).MaxIdleConns, 5)
-			} else {
-				assertEqualE(t, result, test.transport)
+			if test.transportCheck != nil {
+				test.transportCheck(t, castToTransport(result))
```
sfc-gh-pfus, line 967: Do we need this cast? I would get rid of this function if possible...

connection.go
sfc-gh-snoonan: There's few enough calls to getTransport now that I've just removed them all.

connection_test.go
sfc-gh-snoonan: Yes, if we want to assert anything useful about the properties of the transport. We return a RoundTripper, which has a minimal interface: https://github.com/golang/go/blob/8282b72d627269ed2c2f7db7ea744e109667d7b1/src/net/http/client.go#L116

I think this is okay for the test, but would be a smell if we were doing it in the code itself; we're trying to make the client interface easy, but we want to ensure we're doing the right thing internally.

.cursor/rules/overall-guidelines.mdc
```diff
@@ -0,0 +1,56 @@
...
+- **For test-specific rules, see `testing.mdc`**
+- Write both positive and negative test cases
+- Use table-driven tests for testing multiple scenarios
+
+### Code Review Guidelines
+- Ensure code follows Go best practices
+- Verify comprehensive test coverage
+- Check that error messages are descriptive and helpful for debugging
+- Validate that public APIs are properly documented
+# Cursor Rules for Go Snowflake Driver
```
sfc-gh-rsavenok: Is it a bug or a feature that we have same rules declared twice in this file (everything below this line)?

.cursor/rules/overall-guidelines.mdc
```diff
@@ -0,0 +1,56 @@
...
+
+### Code Quality
+- Follow Go formatting standards (use `gofmt`)
+- Use meaningful variable and function names
+- Include error handling for all operations that can fail
+- Write comprehensive documentation for public APIs
+
+### Project Structure
+- Place test files in the same package as the code being tested
+- Use `testdata/` directory for test fixtures and sample data
```
sfc-gh-rsavenok: Is ai smart enough to figure out that the folder is `test_data`?

tls_config.go
```diff
@@ -0,0 +1,41 @@
...
+	"sync"
+)
+
+var (
+	tlsConfigLock     sync.RWMutex
+	tlsConfigRegistry = make(map[string]*tls.Config)
+)
+
+// RegisterTLSConfig registers a custom tls.Config to be used with sql.Open.
+// Use the key as a value in the DSN where tls=value.
```
sfc-gh-rsavenok: Is `tls=value` correct? Reading tests, seems like it should document tlsConfigName=value

sfc-gh-rsavenok (approved):

.cursor/rules/overall-guidelines.mdc
sfc-gh-snoonan: Definitely a bug!

tls_config.go
sfc-gh-snoonan: Thanks! I had originally done it as `tls=value` but we decided to rename it after some discussion. I missed this one.

connection_test.go
sfc-gh-snoonan: I'm going to resolve. If it's a sticking point, I'll fix in a follow-up PR.

PR #1504, opened 2025-08-07
SNOW-2184865 Implement decfloat read path

sfc-gh-pfus:
### Description

SNOW-2184865 Implemented decfloat read path

### Checklist


converter.go
```diff
@@ -2451,6 +2476,50 @@ func arrowRealToValue(srcValue *array.Float64, rowIdx int) snowflakeValue {
...
+		result.Mul(mantissa, powerOfTen)
+		if !decfloatEnabled(ctx) {
+			return result.String(), nil
+		}
+		return result, nil
+	}
+	return nil, nil
+}
+
+func parseTwosComplementBigEndian(b []byte) (*big.Int, error) {
```
sfc-gh-snoonan, line 2519: I thought there must have been a built in that does this, but no. 🤦

converter.go
```diff
@@ -161,6 +161,14 @@ func snowflakeTypeToGo(ctx context.Context, dbtype snowflakeType, precision int6
 		return reflect.TypeOf(float64(0))
 	case realType:
 		return reflect.TypeOf(float64(0))
+	case decfloatType:
+		if !decfloatEnabled(ctx) {
+			return reflect.TypeOf("")
```
sfc-gh-snoonan, line 166: IMO it is worth a function level comment that for numerical types we do not handle, we default to string.

sfc-gh-pfus: This is a deeply hidden private function, so these comment will be visible only while browsing the code. I generally think, that this code speaks volumes - all types are clearly understandable.

converter.go
sfc-gh-pfus: Me too. But there is only a function that parses to big.Float.

converter.go
sfc-gh-pfus: There is a comment describing this behaviour on `WithDecfloatEnabled` - this is much more important, as this is used by customers to enable this feature.

sfc-gh-pfus: I added also a mention to our docs.

converter.go
```diff
@@ -2451,6 +2476,70 @@ func arrowRealToValue(srcValue *array.Float64, rowIdx int) snowflakeValue {
...
+			return nil, fmt.Errorf("failed to parse mantissa bytes: %s, error: %v", hex.EncodeToString(mantissaBytes), err)
+		}
+		if decfloatMappingEnabled(ctx) {
+			mantissa := new(big.Float).SetPrec(127).SetInt(mantissaInt)
+			powerOfTen, ok := new(big.Float).SetPrec(127).SetString(fmt.Sprintf("1e%d", exponent))
+			if !ok {
+				return nil, fmt.Errorf("failed to create power of ten for exponent %d", exponent)
+			}
+			result := &big.Float{}
+			result.Mul(mantissa, powerOfTen)
```
sfc-gh-tkissinger: This multiplication could lead to more inexactness as we are not computing with exact values here. Can we change this to parse this whole thing from a string?

converter.go
```diff
@@ -2451,6 +2476,70 @@ func arrowRealToValue(srcValue *array.Float64, rowIdx int) snowflakeValue {
...
+			result = mantissaStr + "e" + strconv.Itoa(int(exponent))
+		}
+		return result, nil
+	}
+	return nil, nil
+}
+
+func parseTwosComplementBigEndian(b []byte) (*big.Int, error) {
+	if len(b) > 16 {
+		return nil, fmt.Errorf("input byte slice is too long (max 16 bytes)")
```
sfc-gh-tkissinger, line 2521: Might also be a signal that the driver should be updates. E.g., if we support higher precisions in the future. Might be something we could point the customer to

converter.go
sfc-gh-pfus: Hmmm, how? This is Arrow, we have a separate mantissa and exponent.

converter.go
sfc-gh-pfus: If we support higher precision, we should just update the driver. I don't think such messages suggesting updates are popular in drivers, we just return some kind of "not implemented" error then.

converter.go
sfc-gh-tkissinger: mantissa to string; then parse "`mantissaStr`e`exponent`"

PR #1509, opened 2025-08-11
Support using pre-signed JWT for AuthTypeJwt authentication

radu-todirica:
**Summary**
This PR adds support for passing an already signed JWT to the Snowflake AuthTypeJwt authenticator.
If a JWT is provided, it will be used directly for authentication.
If not, the existing flow remains unchanged — the authenticator will generate a new JWT using the provided private key.

**Background / Reason for Change**
In our project, the private key used for Snowflake authentication is stored in AWS Secrets Manager.
Currently, every lambda function that uses the Snowflake Go driver’s requires retrieving the privateKey from AWS Secrets.
To minimize calls to AWS Secrets Manager, we want the option to sign our own JWT tokens in the authorizer lambda function, where the private key can be kept in an in-memory cache, and pass the already signed jwt from authorizer to the rest of the lambda function instead of sending the privateKey.

**Changes in this PR**

- Added a check in the AuthTypeJwt authenticator to use a provided JWT if available.
- Fallback to the existing implementation of generating a JWT with the private key if no token is passed.
- No breaking changes — current behavior is preserved when no JWT is supplied.

**Benefits**

- Allows caching the private key in our authorizer lambda to avoid repeated AWS Secrets lookups.
- Reduces latency and AWS Secrets Manager costs.
- Maintains backward compatibility.

### DIFF
--- a/.cursor/rules/overall-guidelines.mdc
+++ b/.cursor/rules/overall-guidelines.mdc
@@ -1,0 +1,30 @@
+---
+alwaysApply: true
+---
+
+# Cursor Rules for Go Snowflake Driver
+
+## General Development Standards
+
+### Code Quality
+- Follow Go formatting standards (use `gofmt`)
+- Use meaningful variable and function names
+- Include error handling for all operations that can fail
+- Write comprehensive documentation for public APIs
+
+### Project Structure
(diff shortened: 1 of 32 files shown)


radu-todirica:

PR #1510, opened 2025-08-11
SNOW-2176203 Remove cacert.go and custom certificate bundle

sfc-gh-pfus:
### Description

SNOW-2176203 Removed custom, static certificate bundle.

This certificate bundle was used only during OCSP. Such an approach had multiple problems:
1. Discrepancy between cert pool used to validate if cert is trusted and OCSP.
2. Cert pool was not updated regularly.
3. It was changed in runtime in a strange policy.

### Checklist


ocsp.go
```diff
@@ -681,22 +676,8 @@ func isValidOCSPStatus(status ocspStatusCode) bool {
...
-			rca := caRoot[string(verifiedChains[i][numberOfNoneRootCerts].RawIssuer)]
-			if rca == nil {
-				return fmt.Errorf("failed to find root CA. pkix.name: %v", verifiedChains[i][numberOfNoneRootCerts].Issuer)
-			}
-			verifiedChains[i] = append(verifiedChains[i], rca)
-			numberOfNoneRootCerts++
-		}
 		results := getAllRevocationStatus(ctx, verifiedChains[i])
-		if r := canEarlyExitForOCSP(results, numberOfNoneRootCerts); r != nil {
+		if r := canEarlyExitForOCSP(results, len(verifiedChains[i])-1); r != nil {
```
sfc-gh-snoonan: The rest of the removal LGTM. Need to spend some time looking at the functions here.

sfc-gh-snoonan: I rewrote this as below because I think it is easier to see the problem

```
// verifyPeerCertificate verifies all of certificate revocation status
func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certificate) (err error) {
	// Each chain passed to us has already been validated; we are just checking revocation via ocsp here.
	// Each chain represents a path through the graph to a valid root. The root will be the last cert in the chain.
	for i, chain := range verifiedChains {
		numberOfNonRootCerts := len(chain) - 1
		// pfus and snoonan both think this is an off by one error in the original code
		lastCert := chain[numberOfNonRootCerts]
		if !lastCert.IsCA || string(lastCert.RawIssuer) != string(lastCert.RawSubject) {
			// if the last certificate is not a CA or is not self signed and it is part of our caRoot, add it to the list
			rca := caRoot[string(lastCert.RawIssuer)]
			if rca == nil {
				return fmt.Errorf("failed to find root CA. pkix.name: %v", lastCert.Issuer)
			}
			// So this would add the same cert to the end, and increase the number of certs?
			numberOfNonRootCerts++
		}
		results := getAllRevocationStatus(ctx, verifiedChains[i])
		if r := canEarlyExitForOCSP(results, numberOfNonRootCerts); r != nil {
			return r.err
		}
	}
```

Suggest we rewrite this as:
```
// verifyPeerCertificate verifies all of certificate revocation status
func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certificate) (err error) {
	for _, chain := range verifiedChains {
		if r := canEarlyExitForOCSP(ctx, chain); r != nil {
			return r.err
		}
	}

	ocspResponseCacheLock.Lock()
	if cacheUpdated {
		writeOCSPCacheFile()
	}
	cacheUpdated = false
	ocspResponseCacheLock.Unlock()
	return nil
}
func canEarlyExitForOCSP(ctx context.Context, chain []*x509.Certificate) *ocspStatus {
	results := getAllRevocationStatus(ctx, chain)
	msg := ""
	if atomic.LoadUint32((*uint32)(&ocspFailOpen)) == (uint32)(OCSPFailOpenFalse) {
		// Fail closed. any error is returned to stop connection
		for _, r := range results {
			if r.err != nil {
				return r
			}
		}
		return nil
	}
	// Fail open and all results are valid.
	allValid := len(results) == (len(chain)-1)
```

What do you think?

sfc-gh-snoonan (approved):
I think we can rip out the 'test ocsp responder' stuff as well.

sfc-gh-dszmolka:
do we also want to remove https://github.com/snowflakedb/gosnowflake/blob/master/doc/runbook-update-ca-certs.md @sfc-gh-pfus please?

ocsp.go
sfc-gh-pfus: Totally agree!

ocsp.go
```diff
@@ -681,22 +653,8 @@ func isValidOCSPStatus(status ocspStatusCode) bool {
 // verifyPeerCertificate verifies all of certificate revocation status
 func verifyPeerCertificate(ctx context.Context, verifiedChains [][]*x509.Certificate) (err error) {
 	for i := 0; i < len(verifiedChains); i++ {
```
sfc-gh-snoonan: nit: `for _, chain := range verifiedChains` and then we don't need i.

ocsp.go
```diff
@@ -721,7 +679,7 @@ func canEarlyExitForOCSP(results []*ocspStatus, chainSize int) *ocspStatus {
 		}
 	} else {
 		// Fail open and all results are valid.
-		allValid := len(results) == chainSize
+		allValid := len(results) == len(verifiedChain)-1
```
sfc-gh-snoonan: nit: add a comment on the -1 since we had a side conversation on it?

ocsp.go
sfc-gh-pfus: Done.

ocsp.go
sfc-gh-pfus: Done.

sfc-gh-pbulawa (approved):

PR #1511, opened 2025-08-11
SNOW-2230380 [GO] Support proxy parameters in the connection string

sfc-gh-ext-simba-jy:
### Description

SNOW-2230380
## New Parameters

| Parameter                  | Description                                                                                          | Default   |
|----------------------------|------------------------------------------------------------------------------------------------------|-----------|
| `proxyHost`                 | Hostname or IP address of the proxy server.                                                          |           |
| `proxyPort`                 | Port number of the proxy server.                                                                     |           |
| `proxyUser`                 | Username for proxy authentication.                                                                  |           |
| `proxyPassword`              | Password for proxy authentication.                                                                  |           |
| `proxyProtocol`              | Protocol to use for proxy connection.                                                               | `http`    |
| `noProxy`                    | Comma-separated list of hosts that should bypass the proxy.                                         |           |

---

### Checklist


cmd/connectionproxy/connectionproxy.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		log.Fatalf("failed to create Config, err: %v", err)
+	}
+	cfg.ProxyHost = "<Proxy Host>"
+	cfg.ProxyPort = 1234       // Replace with your proxy port
+	cfg.ProxyProtocol = "http" // or "https". Default value is "http"
+
+	//If you need to use a proxy with authentication, set the following:
+	cfg.ProxyUser = "<Proxy User>"
+	cfg.ProxyPassword = "<Proxy Password>"
```
sfc-gh-pfus, line 34: Missing `noProxy`?

cmd/connectionproxy/connectionproxy.go
```diff
@@ -0,0 +1,67 @@
...
+		log.Fatalf("failed to create DSN from Config, err: %v", err)
+	}
+
+	db, err := sql.Open("snowflake", dsn)
+	if err != nil {
+		log.Fatalf("failed to connect. err: %v", err)
+	}
+	defer db.Close()
+	query := "SELECT 1"
+	rows, err := db.Query(query) // no cancel is allowed
```
sfc-gh-pfus: I know this comment exists everywhere, but it's nonsense, let's not add it in the new code :)

cmd/connectionproxy/connectionproxy.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		log.Fatalf("failed to run a query. %v, err: %v", query, err)
+	}
+	defer rows.Close()
+	var v int
+	for rows.Next() {
```
sfc-gh-pfus: Same here - we should replace it with:
```
if !rows.Next() {
  log.Fatalf("no rows returned")
}
```

driver.go
```diff
@@ -34,7 +34,14 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
 
 // OpenConnector creates a new connector with parsed DSN.
 func (d SnowflakeDriver) OpenConnector(dsn string) (driver.Connector, error) {
-	cfg, err := ParseDSN(dsn)
+		var cfg *Config
+	var err error
+	if dsn == "autoConfig" {
```
sfc-gh-pfus: Is it connected to https://github.com/snowflakedb/gosnowflake/issues/1523 ? If yes, can you make it in a separate change?

driver.go
```diff
@@ -34,7 +34,14 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
 
 // OpenConnector creates a new connector with parsed DSN.
 func (d SnowflakeDriver) OpenConnector(dsn string) (driver.Connector, error) {
-	cfg, err := ParseDSN(dsn)
+		var cfg *Config
```
sfc-gh-pfus: Formatting.

driver.go
```diff
@@ -34,7 +34,14 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
...
+	var err error
+	if dsn == "autoConfig" {
+		cfg, err = loadConnectionConfig()	
+	}else {
```
sfc-gh-pfus: Formatting.

transport.go
```diff
@@ -51,6 +55,35 @@ func newTransportFactory(config *Config, telemetry *snowflakeTelemetry) *transpo
 	return &transportFactory{config: config, telemetry: telemetry}
 }
 
+func (tf *transportFactory) createProxy() func(*http.Request) (*url.URL, error) {
```
sfc-gh-pfus, line 57: I think it is worth adding some logs here.

transport.go
```diff
@@ -51,6 +55,35 @@ func newTransportFactory(config *Config, telemetry *snowflakeTelemetry) *transpo
...
+	}
+
+	cfg := httpproxy.Config{
+		HTTPSProxy: connectionProxy.String(),
+		HTTPProxy:  connectionProxy.String(),
+		NoProxy:    tf.config.NoProxy,
+	}
+	proxyURLFunc := cfg.ProxyFunc()
+
+	return func(req *http.Request) (*url.URL, error) {
```
sfc-gh-pfus, line 82: Hmmm, isn't it enough to just `return proxyURLFunc`?

transport_test.go
```diff
@@ -172,3 +172,63 @@ func TestDirectTLSConfigOnly(t *testing.T) {
 	assertNilF(t, err, "Unexpected error")
 	assertNotNilF(t, transport, "Expected non-nil transport")
 }
+
+type proxyTest struct {
```
sfc-gh-pfus: Can we inline this struct?

transport_test.go
```diff
@@ -172,3 +172,63 @@ func TestDirectTLSConfigOnly(t *testing.T) {
...
+		req, _ := http.NewRequest("GET", "https://testing.snowflakecomputing.com", nil)
+		proxyURL, _ := proxyFunc(req)
+
+		if test.proxyURL == "" {
+			assertNilF(t, proxyURL, "Expected nil proxy for https request")
+		} else {
+			assertEqualF(t, proxyURL.String(), test.proxyURL)
+		}
+	}
+}
```
sfc-gh-pfus, line 240: Have you checked it in action using MITM Proxy for instance? Does it work for both HTTP (OCSP/CRL) and HTTPS traffic?

cmd/connectionproxy/connectionproxy.go
sfc-gh-ext-simba-jy: I added it.

cmd/connectionproxy/connectionproxy.go
sfc-gh-ext-simba-jy: Deleted the comment.

cmd/connectionproxy/connectionproxy.go
sfc-gh-ext-simba-jy: I changed it.

driver.go
sfc-gh-ext-simba-jy: When I tested the connection proxy with the toml file configuration, I found that the toml config connection is not working due to the new function. I agree that we need to make this change in a separate ticket. I reverted back whole driver.go file.

driver.go
sfc-gh-ext-simba-jy: Reverted back.

driver.go
sfc-gh-ext-simba-jy: reverted back.

transport.go
sfc-gh-ext-simba-jy: Not exactly. The proxy field in the transport expects a function of type func(req *http.Request) (*url.URL, error), whereas proxyURLFunc has the type func(reqURL *url.URL) (*url.URL, error). Therefore, we need to manually pass the URL from the http.Request to proxyURLFunc.

transport.go
sfc-gh-ext-simba-jy: I added some info logs.

transport_test.go
sfc-gh-ext-simba-jy: Done. Thank you!

transport_test.go
sfc-gh-ext-simba-jy: I have tested the proxy manually with Squid proxy in Docker. In my testing, it works both HTTP and HTTPS. I will test this again after we fix the OCSP init issue.

driver.go
sfc-gh-ext-simba-jy: I created another PR here: https://github.com/snowflakedb/gosnowflake/pull/1530

transport_test.go
sfc-gh-ext-simba-jy: I have modified the connection.go to apply the proxy for the OCSP connection. Please have a look. @sfc-gh-pfus

connection.go
```diff
@@ -813,6 +813,12 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
 		return nil, err
 	}
 
+	if config.ProxyHost != "" {
+		//Reinitialize the transport with the new factory
+		SnowflakeTransport = transportFactory.createOCSPTransport()
```
sfc-gh-pfus: I don't think we should mess up with global variables. Especially that they are deprecated.

sfc-gh-pbulawa: +1

sfc-gh-ext-simba-jy: Hello @sfc-gh-pfus, @sfc-gh-pbulawa. I’m aware that those global variables are deprecated. However, in the current OCSP certificate download workflow, we are passing snowflakeNoRevocationCheckTransport as the transport [here](https://github.com/snowflakedb/gosnowflake/blob/master/ocsp.go#L784). Since I haven’t seen any new design regarding how to handle this after the deprecation of the global variables, I updated it this way. Could you suggest what would be the best approach here? If there’s a design doc or any direction, I’d appreciate it if you could share your thoughts.

sfc-gh-ext-simba-jy: Could you please confirm whether snowflakeNoRevocationCheckTransport is also deprecated?
From what I see, we don’t need to reinitialize SnowflakeTransport, but only snowflakeNoRevocationCheckTransport. In the current OCSP certificate download workflow, we are passing snowflakeNoRevocationCheckTransport as the transport [here](https://github.com/snowflakedb/gosnowflake/blob/master/ocsp.go#L784). If you think this approach would cause issues with the global variables, I’d appreciate it if you could share your thoughts on how we should replace this transport — for example, a new global variable in the OCSP.go with a new transport instance that incorporates the connection proxy or another approach you’d recommend.

If we are going to deprecate snowflakeNoRevocationCheckTransport, could you suggest the best approach to handle this case? Is there a design doc or any direction on how this should be managed? Since I haven’t seen any new design regarding handling this after the global variable deprecation, I updated it this way for now.

sfc-gh-pfus: My thinking is:
- `SnowflakeTransport` is kept here only for compatibility. It is initialized with some default config, but shouldn't be used, either by customers or by us. I checked that it is used by connectivity diagnoser - to be honest, I would get rid of the `if` where it is used and just panic if things go south.
- `SnowflakeTransportTest` - the same.
- `snowflakeNoRevocationCheckTransport` - this is an internal var, so we can do whatever we want. I think we should get rid of it too. It is used in OCSP module only. We can create it in OCSP module using `factory.createNoRevocationTransport()` (probably we would need to pass config to `ocspValidator`.

sfc-gh-ext-simba-jy: I added the config in the ocspValidator struct, and updated the transport workflow. Please have a look.

cmd/connectionproxy/.gitignore
```diff
@@ -0,0 +1 @@
+connectionproxy
```
sfc-gh-pfus, line 1: I forgot - can you add these new parameters to doc.go? Othwerise LGTM. Add it and merge it.

ocsp.go
```diff
@@ -777,11 +779,13 @@ func (ov *ocspValidator) downloadOCSPCacheServer() {
 	if err != nil {
 		return
 	}
+
+	factory := newTransportFactory(ov.cfg, nil)
 	logger.Infof("downloading OCSP Cache from server %v", ocspCacheServerURL)
 	timeout := OcspCacheServerTimeout
 	ocspClient := &http.Client{
 		Timeout:   timeout,
-		Transport: snowflakeNoRevocationCheckTransport,
```
sfc-gh-pfus, line 796: We can remove `snowflakeNoRevocationCheckTransport` now, right?

cmd/connectionproxy/.gitignore
sfc-gh-ext-simba-jy: I updated the doc.

ocsp.go
```diff
@@ -1142,7 +1149,7 @@ func (occ *ocspCacheClearerType) stop() {
 	}
 }
 
-// snowflakeNoRevocationCheckTransport is the transport object that doesn't do certificate revocation check with OCSP.
+// only testing purpose to count the round trip.
```
sfc-gh-pfus: Can we just remove it? If OCSP/CRL is not important, we can just use Go's default `http.Transport`.

sfc-gh-pfus (approved):

ocsp.go
sfc-gh-ext-simba-jy: Oh, my message was missing. I replaced the snowflakeNoRevocationCheckTransport in the testing code with other functions, except for one test called TestOCSPUnexpectedResponses. In this test, the OCSP client transport is set to a round tripper in order to count how many times the request is called. I tried to replace this part as well, but I haven’t been able to. Do you have any suggestions or ideas on how to handle this? [here](https://github.com/snowflakedb/gosnowflake/blob/f76304431bff16e2d1c542bf35738db0e2420b49/driver_ocsp_test.go#L706)

sfc-gh-pfus: https://github.com/snowflakedb/gosnowflake/pull/1553/files - you can merge this into your branch.

dsn.go
```diff
@@ -614,6 +639,12 @@ func fillMissingConfigParameters(cfg *Config) error {
 		cfg.IncludeRetryReason = ConfigBoolTrue
 	}
 
+	if cfg.ProxyHost != "" {
```
sfc-gh-dheyman: can we replace this with one if statement for more concise code?

sfc-gh-ext-simba-jy: Done!

PR #1519, opened 2025-08-13
NO-SNOW Fix TestOpenWithConfigCancel

sfc-gh-pfus:
### Description

### Checklist


sfc-gh-pfus:
> LGTM to unblock the release / deflake, but it would be nice if we weren't relying on the equivalent of sleep(1) here :-).

@sfc-gh-snoonan actually we have to - this is a test of timeouts. And the problem was that this timeout was too low and we won't even reach the function that handles timeout.

sfc-gh-snoonan:
> > LGTM to unblock the release / deflake, but it would be nice if we weren't relying on the equivalent of sleep(1) here :-).
> 
> @sfc-gh-snoonan actually we have to - this is a test of timeouts. And the problem was that this timeout was too low and we won't even reach the function that handles timeout.

Yes, but it will always be susceptible to flakes unless we switch to using a virtual clock.

PR #1521, opened 2025-08-13
Avoid wasteful processing in exceptionTelemetry

obs-gh-felixbowing:
In case telemetry is disabled, this function still calls the expensive generateTelemetryExceptionData function.

This commit skips all unnecessary steps
unless telemetry is explicitly enabled. This should save quite some resources as
generateTelemetryExceptionData processes potentially large stack traces.

Testing:
No semantic changes, hence all existing tests should pass. No extra tests needed.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -63,16 +63,13 @@ func (se *SnowflakeError) generateTelemetryExceptionData() *telemetryData {
 	return data
 }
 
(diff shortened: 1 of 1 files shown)


sfc-gh-pfus:
Thanks again, @obs-gh-felixbowing , your PR is merged!

PR #1530, opened 2025-08-15
SNOW-2270316: [Go] Toml file connection is not working.

sfc-gh-ext-simba-jy:
### Description

SNOW-2270316: The TOML file connection is not working properly because the Go connector uses the OpenConnector function instead of Open. We should add the autoConfig option to address this.

### Checklist

### DIFF
--- a/driver.go
+++ b/driver.go
@@ -34,7 +34,13 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
 
 // OpenConnector creates a new connector with parsed DSN.
 func (d SnowflakeDriver) OpenConnector(dsn string) (driver.Connector, error) {
-	cfg, err := ParseDSN(dsn)
+	var cfg *Config
+	var err error
+	if dsn == "autoConfig" {
+		cfg, err = loadConnectionConfig()
+	} else {
+		cfg, err = ParseDSN(dsn)
+	}
 	if err != nil {
 		return Connector{}, err
 	}


driver.go
```diff
@@ -34,7 +34,13 @@ func (d SnowflakeDriver) Open(dsn string) (driver.Conn, error) {
 
 // OpenConnector creates a new connector with parsed DSN.
 func (d SnowflakeDriver) OpenConnector(dsn string) (driver.Connector, error) {
-	cfg, err := ParseDSN(dsn)
+	var cfg *Config
```
sfc-gh-pfus, line 37: Can we have a test to see if it didn't work previously and it works now?

sfc-gh-ext-simba-jy: It is difficult to create a test for this on my side. To properly test it, we would need to implement an integration test. However, in that case, we would also need to find a way to encrypt the .toml file to ensure that credentials are not exposed.

sfc-gh-pfus: Can we use wiremock for this? We already have some tests based on wiremock.

sfc-gh-ext-simba-jy: Please correct me if I am wrong. As far as I know, we need to set a custom transport if we want to use WireMock. Unfortunately, the TOML file connection feature does not provide an option to set a custom transport or TLSClientConfig. In the testing workflow with WireMock, the steps are: 1) create the DSN, 2) create the connector, and 3) run sql.OpenDB. However, the TOML-based connection uses sql.Open, which is a different scenario. I have tried creating a unit test and succeeded in instantiating the DB without establishing a connection. IIn Go, extracting internal connector details is difficult because the *sql.DB type is opaque and its internal fields are unexported.

sfc-gh-pfus: No, we don't need a custom transport, we just need to create a connection parameters (host, port, user and so on) that points to the wiremock instance. [Here](https://github.com/snowflakedb/gosnowflake/blob/cdd89c1ada1bb58dfcfc48a6262328733971df58/wiremock_test.go#L32) are values that should be included in TOML. I think that you can even use [this](https://github.com/snowflakedb/gosnowflake/blob/cdd89c1ada1bb58dfcfc48a6262328733971df58/wiremock_test.go#L83) function to generate all the parameters you need and create a toml file on the fly in the test.

FYI: sql.Open and sql.OpenDB both works the same, just the first takes DSN and the second takes a connector (that takes config object instead of DSN).

sfc-gh-ext-simba-jy: It seems I may not have conveyed my point clearly. My main issue is that creating the DB using the TOML configuration works fine. The problem occurs when testing the DB in Go, since the test sends a SELECT 1 query, which appears to require the rootCA value in the transport. That’s why I mentioned the previous context.

I was wondering if there’s any other way to verify that the DB came from the TOML configuration without running such a query. For now, I’ve only added an error test after opening the DB, but please let me know if you think additional tests are necessary.

sfc-gh-pfus: But you don't have to use HTTPS wiremock, use HTTP one :)

sfc-gh-ext-simba-jy: I got it. I updated the testing. Please review it. Thank you!

sfc-gh-pfus (approved):

PR #1531, opened 2025-08-18
NO-SNOW Split transport config per usage

sfc-gh-pfus:
### Description

NO-SNOW Splits connection config (like revocation checking algorithm, TLS config etc) from the target service specific characteristic (like timeouts, pool size, connection expiry etc).

### Checklist

### DIFF
--- a/s3_storage_client.go
+++ b/s3_storage_client.go
@@ -6,17 +6,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"net/http"
-	"os"
-	"strings"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/logging"
+	"io"
+	"net/http"
+	"os"
+	"strings"
 )
 
 const (
@@ -51,7 +50,7 @@ func (util *snowflakeS3Client) createClient(info *execResponseStageInfo, useAcce
 	s3Logger := logging.LoggerFunc(s3LoggingFunc)
 	endPoint := getS3CustomEndpoint(info)
 
-	transport, err := newTransportFactory(util.cfg, telemetry).createTransport()
+	transport, err := newTransportFactory(util.cfg, telemetry).createTransport(util.cfg.transportConfigFor(transportTypeCloudProvider))
 	if err != nil {
 		return nil, err
 	}
--- a/transport.go
+++ b/transport.go
@@ -9,10 +9,28 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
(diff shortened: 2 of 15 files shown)


s3_storage_client.go
```diff
@@ -6,17 +6,16 @@ import (
...
-	"os"
-	"strings"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/logging"
+	"io"
```
sfc-gh-snoonan, line 15: nit: I think this is wrong because we generally keep the gh.com imports in a separate block.

transport.go
```diff
@@ -78,19 +57,20 @@ func (tf *transportFactory) createOCSPTransport() *http.Transport {
...
 		}
 	}
-	return tf.createBaseTransport(defaultTransportConfig(), tlsConfig)
+	return tf.createBaseTransport(transportConfig, tlsConfig)
 }
 
 // createNoRevocationTransport creates a transport without certificate revocation checking
-func (tf *transportFactory) createNoRevocationTransport() *http.Transport {
-	return tf.createBaseTransport(defaultTransportConfig(), nil)
+func (tf *transportFactory) createNoRevocationTransport(transportConfig *transportConfig) *http.Transport {
```
sfc-gh-snoonan: Do you want to delete this and just call createBaseTransport directly, or do we leave it since it is clear about the intent? We ignore the TLS config here since this gets used for the basic CRL and OCSP connections and those should be HTTP only, but it might be just as easy to see with the `nil` as the argument.

sfc-gh-snoonan (approved):
I like it. This makes it much clearer what we're trying to accomplish with each type of transport, and your plan for managing retries in a similar pattern sgtm.

transport.go
sfc-gh-pfus: I think this name is meaningful and I like having it. I would prefer leaving it as is. It is initialized for CRL/OCSP directly, so there's no point in creating a transport with TLS then.

s3_storage_client.go
sfc-gh-pfus: To be honest, we use default settings from Intellij Idea in our team and it uses `gofmt` ordering.

transport.go
sfc-gh-snoonan: SGTM

s3_storage_client.go
sfc-gh-snoonan: Yeah, gofmt will order by blocks separated by blank lines. If you get rid of the blank line, it treats them all the same. Generally, we do the core libraries in a separate block before the gh.com ones. If you don't want to change it, I can live with that.

sfc-gh-pfus: Yeah, I would prefer leaving it as is, as this is a default from the tool we use. Unless there are other global recommendations (like golang's itself or something) worth following.

sfc-gh-snoonan: https://google.github.io/styleguide/go/best-practices.html#import-ordering

sfc-gh-rsavenok (approved):

PR #1535, opened 2025-08-19
NO-SNOW Fix structured types tests on Jenkins

sfc-gh-pfus:
### Description

NO-SNOW Fixed tests that failed on inconsistent order of returned rows.

### Checklist

### DIFF
--- a/ci/container/test_authentication.sh
+++ b/ci/container/test_authentication.sh
@@ -14,9 +14,9 @@ export AUTHENTICATION_TESTS_ENV="docker"
 export RUN_AUTH_TESTS=true
 export AUTHENTICATION_TESTS_ENV="docker"
 
-go test -run TestExternalBrowser*
-go test -run TestClientStoreCredentials
-go test -run TestOkta*
-go test -run TestOauth*
-go test -run TestKeypair*
-go test -run TestEndToEndPat*
+go test -v -run TestExternalBrowser*
+go test -v -run TestClientStoreCredentials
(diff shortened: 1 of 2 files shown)


ci/container/test_authentication.sh
```diff
@@ -14,9 +14,9 @@ export AUTHENTICATION_TESTS_ENV="docker"
 export RUN_AUTH_TESTS=true
 export AUTHENTICATION_TESTS_ENV="docker"
 
-go test -run TestExternalBrowser*
-go test -run TestClientStoreCredentials
-go test -run TestOkta*
-go test -run TestOauth*
-go test -run TestKeypair*
-go test -run TestEndToEndPat*
+go test -v -run TestExternalBrowser*
```
sfc-gh-fpawlowski, line 17: Nit: is adding verbose output for all tests long-term change, or is it a left-over from debugging?

sfc-gh-fpawlowski (approved):
Left small nit, apart from that looks good

ci/container/test_authentication.sh
sfc-gh-pfus: It is - it gives us printing of test names. Without this we are blind - we just see some failed assertions without knowing which tests actually failed. And it prints only test names and status (PASS/FAIL), so not so verbose after all. We use it in our regular builds.

sfc-gh-fpawlowski: Okay, then I was mislead by the name. All seems good 👍

PR #1536, opened 2025-08-19
SNOW-2269982 update doc: quoted identifier, effect of validateDefaultParameters

sfc-gh-dszmolka:
### Description

This is a doc-only change, trying to address the challenges one of the users faced with a mixed-case database. 

### Checklist

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -69,12 +69,12 @@ The following connection parameters are supported:
     account parameter. For details, see the description of the account
     parameter.
 
-  - database: Specifies the database to use by default in the client session
-    (can be changed after login).
-	
-	--> Important note: for the database object and other objects (schema etc), please always adhere to the rules for Snowflake Object Identifiers; especially https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers.
+  - --> Important note: for the database object and other objects (schema, role, etc), please always adhere to the rules for Snowflake Object Identifiers; especially https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers.
     As mentioned in the docs, if you have e.g. a database with mIxEDcAsE naming, as you needed to create it with enclosing it in double quotes, similarly you'll need to reference it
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -71,6 +71,10 @@ The following connection parameters are supported:
 
   - database: Specifies the database to use by default in the client session
     (can be changed after login).
+	
+	--> Important note: for the database object and other objects (schema etc), please always adhere to the rules for Snowflake Object Identifiers; especially https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers.
```
sfc-gh-pfus: It's nice, but the same applies for schema and role too. Do you think it is the best location for this note?

doc.go
```diff
@@ -141,6 +145,11 @@ The following connection parameters are supported:
   - validateDefaultParameters: true by default. Set to false to disable checks on existence and privileges check for
     Database, Schema, Warehouse and Role when setting up the connection
 
+    --> Important note: with the default true value, the connection will fail as the validation fails, if you specify a non-existent database/schema/etc name.
+	This is particularly important when you have a miXedCaSE-named object (e.g. database) and you forgot to properly double quote it.
+    This behaviour is still preferable as it provides a very clear, fail-fast  indication of the configuration error. If you would still like to forego this validation,
+	which ensures that the driver always connects with proper database, schema etc. and creates a proper context for it, you can set this configuration to false to allow connection with invalid object identifiers.
```
sfc-gh-pfus, line 151: Maybe worth adding something like "Driver doesn't guarantee the behaviour in this case" to make it more scary?

doc.go
sfc-gh-dszmolka: you're right; moved this section a bit above database

doc.go
sfc-gh-dszmolka: added a bit more spicy wording

sfc-gh-pbulawa (approved):

PR #1540, opened 2025-08-21
fix: handle context errors in isRetryableError

frbvianna:
### Description

While using the ADBC driver, when the context is canceled, we get a flood of logs such as the following:
```
WARN[0026]retry.go:330 gosnowflake.(*retryHTTP).execute failed http connection. err: Get "https://<host>/results/<...>/main/data_W_X_Y_Z?<query-parameters>": context canceled. retrying...
```
My understanding is, since the context has been canceled, it makes no sense to retry such requests. In fact, it shouldn't be retried anyway since it will return the `ctx.Err()` at the end of the function:
https://github.com/snowflakedb/gosnowflake/blob/cdd89c1ada1bb58dfcfc48a6262328733971df58/retry.go#L380-L383

Can you please evaluate if this is OK? 

### Checklist

### DIFF
--- a/driver_test.go
+++ b/driver_test.go
@@ -1826,7 +1826,7 @@ func TestLargeSetResultCancel(t *testing.T) {
 		time.Sleep(time.Second)
 		cancel()
 		ret := <-c
-		if !errors.Is(ret.Error(), context.Canceled) {
+		if !errors.Is(ret, context.Canceled) {
 			t.Fatalf("failed to cancel. err: %v", ret)
 		}
 		close(c)
--- a/driver_test.go
+++ b/driver_test.go
@@ -1826,7 +1826,7 @@ func TestLargeSetResultCancel(t *testing.T) {
 		time.Sleep(time.Second)
 		cancel()
 		ret := <-c
-		if ret.Error() != "context canceled" {
+		if !errors.Is(ret.Error(), context.Canceled) {
 			t.Fatalf("failed to cancel. err: %v", ret)
 		}
 		close(c)
--- a/driver_test.go
+++ b/driver_test.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rsa"
 	"database/sql"
 	"database/sql/driver"
+	"errors"
 	"flag"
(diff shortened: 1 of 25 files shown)


frbvianna:

sfc-gh-rsavenok (approved):

retry.go
```diff
@@ -386,6 +386,9 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
 
 func isRetryableError(req *http.Request, res *http.Response, err error) (bool, error) {
 	if err != nil && res == nil { // Failed http connection. Most probably client timeout.
+		if errors.Is(err, context.Canceled) {
```
sfc-gh-pfus: Can you check also if the error may be `context.DeadlineExceeded`?

sfc-gh-pfus:
@frbvianna We are close, only one test failed `TestCancelQuery`. I suggest changing it to be more resilient like this:
```
		_, err := dbt.conn.QueryContext(ctx, "CALL SYSTEM$WAIT(10, 'SECONDS')")
		assertNotNilF(t, err)
		assertErrIsE(t, err, context.DeadlineExceeded)
```

frbvianna:
@sfc-gh-pfus, did some changes to resolve the test issue. Can you please check again?

driver_test.go
```diff
@@ -1825,7 +1826,7 @@ func TestLargeSetResultCancel(t *testing.T) {
 		time.Sleep(time.Second)
 		cancel()
 		ret := <-c
-		if ret.Error() != "context canceled" {
+		if !errors.Is(ret.Error(), context.Canceled) {
```
sfc-gh-pfus: Here we should use just `ret`, not `ret.Error()` (string representation of error).

frbvianna: Thanks, applying below.
```suggestion
		if !errors.Is(ret, context.Canceled) {
```

sfc-gh-pfus:
Hi @frbvianna ! I merged your change, thank you very much!

PR #1547, opened 2025-08-22
SNOW-1646792 Handle multistatements in QueryArrowStream

sfc-gh-pfus:
### Description

SNOW-1646792 Handles multistatements in QueryArrowStream

### Checklist


connection.go
```diff
@@ -679,21 +690,35 @@ func (asb *ArrowStreamBatch) GetStream(ctx context.Context) (io.ReadCloser, erro
...
+	// Location returns the location associated with the query response.
+	// If multistatement is used, this is the location of the initial query.
 	Location() *time.Location
+	// JSONData returns the data if JSON was returned instead of Arrow.
+	// If multistatement is used, this is the data for the current result set.
 	JSONData() [][]*string
 }
 
 type snowflakeArrowStreamChunkDownloader struct {
+	ctx         context.Context
```
zeroshade: Rather than storing the context in the downloader, should we instead have the `NextResultSet` function take a context as an argument? That would be the more idiomatic way of handling things (in general you want to avoid storing a context when possible)

sfc-gh-pfus: I was going back and forth on this. I wanted to preserve similar API to what Go offers (`NextResultSet` on neither `sql.Rows` nor `driver.Rows` takes context as a parameter), but I agree with your point.

zeroshade: I think my preference would be to use the context as an argument to `NextResultSet` because it's a new method so you wouldn't be breaking any interfaces that users are already using.

connection.go
```diff
@@ -781,6 +808,48 @@ func (scd *snowflakeArrowStreamChunkDownloader) GetBatches() (out []ArrowStreamB
...
+	resultID := scd.resultIDs[0]
+	logger.Debugf("downloading next result set, resultID: %v", resultID)
+	scd.resultIDs = scd.resultIDs[1:]
+	resultPath := fmt.Sprintf(urlQueriesResultFmt, resultID)
+	resp, err := scd.sc.getQueryResultResp(ctx, resultPath)
+	if err != nil {
+		return err
+	}
+	if !resp.Success {
+		logger.WithContext(ctx).Errorf("error: %v", err)
```
sfc-gh-snoonan: err will be nil here though, because we are checking that above on line 821 and returning. What's the expectation with resp.Success vs err?

sfc-gh-pfus: Good catch!

PR #1549, opened 2025-08-25
SNOW-2294562: Change GET to do multi-part parallel download for S3, Azure, and GCP

sfc-gh-gren:
### Description

SNOW-2294562
This PR contains the following changes:
1. Pass the `parallel` parameter to S3/Azure's download API to enable multi-part download on each file.
2. Implement multi-part parallel download for GCP which utilizes the same parameters as S3 and Azure.
3. Use `meta.options.MultiPartThreshold` as the part size when calling S3/Azure's download API. This matches the behavior of the PUT command for S3 ([link](https://github.com/snowflakedb/gosnowflake/blob/858561dce1e02cd839bc2ad2e57a2c396ace14b8/s3_storage_client.go#L213)).
4. Updated tests.

Tested [put_get_test.go](https://github.com/snowflakedb/gosnowflake/pull/1549/files#diff-a3eaa8612e5469f8b99e45b5d4a7c02fc11925993c07205b069fb6a6032c10df) against both AWS and Azure deployments.

### Checklist

### DIFF
--- a/azure_storage_client.go
+++ b/azure_storage_client.go
@@ -265,7 +265,8 @@ func (util *snowflakeAzureClient) uploadFile(
 func (util *snowflakeAzureClient) nativeDownloadFile(
 	meta *fileMetadata,
 	fullDstFileName string,
-	maxConcurrency int64) error {
+	maxConcurrency int64,
+	partSize int64) error {
 	azureLoc, err := util.extractContainerNameAndPath(meta.stageInfo.Location)
 	if err != nil {
 		return err
@@ -319,7 +320,9 @@ func (util *snowflakeAzureClient) nativeDownloadFile(
 		_, err = withCloudStorageTimeout(util.cfg, func(ctx context.Context) (any, error) {
 			return blobClient.DownloadFile(
 				ctx, f, &azblob.DownloadFileOptions{
-					Concurrency: uint16(maxConcurrency)})
+					Concurrency: uint16(maxConcurrency),
+					BlockSize:   int64Max(partSize, blob.DefaultDownloadBlockSize),
+				})
 		})
 		if err != nil {
 			return err
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -32,10 +32,11 @@ type (
 )
 
 const (
-	fileProtocol              = "file://"
-	dataSizeThreshold int64   = 64 * 1024 * 1024
-	isWindows                 = runtime.GOOS == "windows"
-	mb                float64 = 1024.0 * 1024.0
+	fileProtocol                      = "file://"
+	dataSizeThreshold         int64   = 64 * 1024 * 1024
+	streamingDownloadPartSize int64   = 8 * 1024 * 1024
+	isWindows                         = runtime.GOOS == "windows"
+	mb                        float64 = 1024.0 * 1024.0
 )
 
 const (
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -178,12 +178,19 @@ func (sfa *snowflakeFileTransferAgent) execute() error {
 		if sfa.stageLocationType != local {
 			sizeThreshold := sfa.options.MultiPartThreshold
 			meta.options.MultiPartThreshold = sizeThreshold
-			if meta.srcFileSize > sizeThreshold && sfa.commandType == uploadCommand {
+			if sfa.commandType == uploadCommand {
+				if meta.srcFileSize > sizeThreshold {
+					meta.parallel = sfa.parallel
+					largeFileMetas = append(largeFileMetas, meta)
+				} else {
+					meta.parallel = 1
+					smallFileMetas = append(smallFileMetas, meta)
+				}
+			} else {
+				// Enable multi-part download for all files to improve performance.
+				// The MultiPartThreshold will be passed to the Cloud Storage Provider to determine the part size.
 				meta.parallel = sfa.parallel
 				largeFileMetas = append(largeFileMetas, meta)
-			} else {
-				meta.parallel = 1
-				smallFileMetas = append(smallFileMetas, meta)
 			}
 		} else {
 			meta.parallel = 1
@@ -196,7 +203,7 @@ func (sfa *snowflakeFileTransferAgent) execute() error {
 			return err
 		}
 	} else {
-		if err = sfa.download(smallFileMetas); err != nil {
+		if err = sfa.download(largeFileMetas); err != nil {
 			return err
 		}
 	}
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -32,11 +32,11 @@ type (
 )
 
 const (
-	fileProtocol                      = "file://"
-	dataSizeThreshold         int64   = 64 * 1024 * 1024
-	streamingDownloadPartSize int64   = 8 * 1024 * 1024
-	isWindows                         = runtime.GOOS == "windows"
-	mb                        float64 = 1024.0 * 1024.0
+	fileProtocol                        = "file://"
+	multiPartThreshold          int64   = 64 * 1024 * 1024
+	streamingMultiPartThreshold int64   = 8 * 1024 * 1024
+	isWindows                           = runtime.GOOS == "windows"
+	mb                          float64 = 1024.0 * 1024.0
 )
 
 const (
--- a/gcs_storage_client.go
+++ b/gcs_storage_client.go
@@ -373,9 +373,11 @@ func (util *snowflakeGcsClient) getFileHeaderForDownload(downloadURL *url.URL, g
 	if err != nil {
 		return nil, err
 	}
-	if resp != nil && resp.Body != nil {
-		defer resp.Body.Close()
-	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			logger.Errorf("Failed to close response body: %v", err)
+		}
+	}()
 
 	if resp.StatusCode != http.StatusOK {
 		return nil, util.handleHTTPError(resp, meta, accessToken)
@@ -384,13 +386,14 @@ func (util *snowflakeGcsClient) getFileHeaderForDownload(downloadURL *url.URL, g
 	return resp, nil
 }
 
-// downloadFileInParts downloads a file using parallel range requests
+// downloadPart is a struct for downloading a part of a file in memory
 type downloadPart struct {
 	data  []byte
 	index int64
 	err   error
 }
 
+// downloadPartStream is a struct for downloading a part of a file in a stream
 type downloadPartStream struct {
 	stream io.ReadCloser
 	index  int64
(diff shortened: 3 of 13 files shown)


sfc-gh-gren:
* **#1549** <a href="https://app.graphite.dev/github/pr/snowflakedb/gosnowflake/1549?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/gosnowflake/1549?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

gcs_storage_client.go
```diff
@@ -282,7 +282,8 @@ func (util *snowflakeGcsClient) uploadFile(
 func (util *snowflakeGcsClient) nativeDownloadFile(
 	meta *fileMetadata,
 	fullDstFileName string,
-	maxConcurrency int64) error {
+	maxConcurrency int64,
+	partSize int64) error {
```
sfc-gh-pfus, line 287: You added a new parameter to satisfy an interface, but it is not used. I think GCP supports similar (https://cloud.google.com/storage/docs/sliced-object-downloads#rest-apis). It would be worth adding here as well (we should have file size from previous HEAD request, but it may be absent in `fileMetadata` for now).

put_get_test.go
```diff
@@ -489,7 +489,7 @@ func testPutGet(t *testing.T, isStream bool) {
 			ctx = WithFileTransferOptions(ctx, &SnowflakeFileTransferOptions{GetFileToStream: true})
 			ctx = WithFileGetStream(ctx, &streamBuf)
 		}
-		sql = fmt.Sprintf("get @%%%v 'file://%v'", tableName, tmpDir)
+		sql = fmt.Sprintf("get @%%%v 'file://%v' parallel=10", tableName, tmpDir)
```
sfc-gh-pfus, line 492: So basically client declares that it wants to use 10 parallel downloads. Do I get it correctly that it is the backend that is responsible for calculating `partSize`?

s3_storage_client.go
```diff
@@ -287,6 +288,7 @@ func (util *snowflakeS3Client) nativeDownloadFile(
 	var downloader s3DownloadAPI
 	downloader = manager.NewDownloader(client, func(u *manager.Downloader) {
 		u.Concurrency = int(maxConcurrency)
+		u.PartSize = int64Max(partSize, manager.DefaultDownloadPartSize)
```
sfc-gh-pfus, line 295: Is taking max safe? Maybe we should just set this value if `partSize` is defined, no matter what the value is? (`u.PartSize = cmp.Or(partSize, manager.DefaultDownloadPartSize)`

sfc-gh-gren: Thanks for the question. I was following the uploader logic at line 213 above.

```
uploader = manager.NewUploader(client, func(u *manager.Uploader) {
		u.Concurrency = maxConcurrency
		u.PartSize = int64Max(multiPartThreshold, manager.DefaultUploadPartSize)
	})
```

The `multiPartThreshold` (in our case `partSize`), is initialized to 64MB [here](https://github.com/snowflakedb/gosnowflake/blob/f76304431bff16e2d1c542bf35738db0e2420b49/connection_util.go#L117) if not provided. I think the purpose of int64Max() is that we want to make sure we send the value that's >= the minimum part size (`DefaultUploadPartSize`) required by the SDK. 

I added some comments to indicate that the partSize is set to a default value if not specified. Let me know if it makes sense. Thanks!

put_get_test.go
sfc-gh-gren: We are reusing the [MultiPartThreshold](https://github.com/snowflakedb/gosnowflake/blob/f76304431bff16e2d1c542bf35738db0e2420b49/file_transfer_agent.go#L86) option in SnowflakeFileTransferOptions as the partSize. This is the same behavior for the S3 uploader as well. This parameter will be set to 64MB by default if not specified.
```
// User example:

// Set custom MultiPartThreshold (e.g., 32MB instead of default 64MB)
ctx := gosnowflake.WithFileTransferOptions(context.Background(), &gosnowflake.SnowflakeFileTransferOptions{
    MultiPartThreshold: 32 * 1024 * 1024, // 32MB in bytes
})

// Use the context with your PUT/GET commands
db.ExecContext(ctx, "PUT file:///path/to/local/file @my_stage")
db.QueryContext(ctx, "GET @my_stage/file file:///path/to/download/")
```

```
uploader = manager.NewUploader(client, func(u *manager.Uploader) {
		u.Concurrency = maxConcurrency
		u.PartSize = int64Max(multiPartThreshold, manager.DefaultUploadPartSize).  <-- multiPartThreshold used as partSize
	})
```

gcs_storage_client.go
sfc-gh-gren: Thanks for the pointer. I took a look at the documentation, unlike their SDKs, we will need to implement the chunking and parallel GET logics by ourselves in order to leverage their Range GET REST api. The implementation and testing will be non-trivial. Is it possible to leave it to future improvements?  The `maxConcurrency` parameter is also ignored as of now.

azure_storage_client.go
```diff
@@ -305,9 +306,12 @@ func (util *snowflakeAzureClient) nativeDownloadFile(
 		}
 		defer f.Close()
 		_, err = withCloudStorageTimeout(util.cfg, func(ctx context.Context) (any, error) {
+			// partSize has been initialized to dataSizeThreshold (64MB) by default.
```
sfc-gh-pfus: I'm not a fan of such comments. It comes from the different part of code and it can be easily forgotten in case of updates.

gcs_storage_client.go
sfc-gh-pfus: How will be responsible for implementing it in the future? :) I'm not a fan of such, every now and then we have cases like "why my process on one OS/Cloud/whatever is slower than on the other?" and those are usually marked as "critical" (SEV-2) issues. I'm not a fan of saying "you know, we have a driver that is much faster on AWS than on GCP".

sfc-gh-gren: Sounds good. Added the change for GCP. I've ran put_get_test.go against our GCP deployment and it passed. Please let me know if there are more tests that I can run. Thanks!

azure_storage_client.go
sfc-gh-gren: Removed

gcs_storage_client.go
sfc-gh-pfus: I downsized threshold to 8MB and observed req/resp in intercepting proxy (MITM) - looks good. Thank you!

gcs_storage_client.go
```diff
@@ -303,6 +306,234 @@ func (util *snowflakeGcsClient) nativeDownloadFile(
...
+	close(jobs) // Signal no more jobs
+
+	// Collect results
+	parts := make([][]byte, numParts)
+	for i := int64(0); i < numParts; i++ {
+		result := <-results
+		if result.err != nil {
+			return result.err
+		}
+		parts[result.index] = result.data
```
sfc-gh-pfus, line 582: Well, it doesn't work well with streaming, right? If stream (`WithFileGetStream`) is used, we shouldn't buffer all data in memory. The first thing that comes to my mind is to not start all jobs at once, but start them in `maxConcurrency` batches, emit all (sorted) parts into output stream and only then start another round of batches. Worth considering, if we want to still use 64MB per batch with default parallel = 10 ([source](https://docs.snowflake.com/en/sql-reference/sql/get#optional-parameters) ), or maybe in case of streaming we should downsize this batch threshold?

gcs_storage_client.go
```diff
@@ -303,6 +306,234 @@ func (util *snowflakeGcsClient) nativeDownloadFile(
...
+		}
+		
+		// Add range header for partial content
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start, end))
+		
+		for k, v := range gcsHeaders {
+			req.Header.Add(k, v)
+		}
+		
+		client, err := newGcsClient(util.cfg, util.telemetry)
```
sfc-gh-pfus: Can we reuse the client, at least for downloading a single file (with multiple parts)? It's worth not reopening new connections because each new HTTP connection has its drawbacks (TCP binding, TLS handshake, OCSP checking, etc.).

gcs_storage_client.go
```diff
@@ -303,6 +306,234 @@ func (util *snowflakeGcsClient) nativeDownloadFile(
...
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	// Accept both 200 (full content) and 206 (partial content) status codes
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+		return nil, util.handleHTTPError(resp, meta, accessToken)
+	}
+
+	data, err := io.ReadAll(resp.Body)
```
sfc-gh-pfus: In case of streaming it is worth not reading the whole body. It's better to just keep a stream and then copy from this stream to target stream when needed.

gcs_storage_client.go
sfc-gh-gren: Good call. Added the function to do batched download for the streaming case. I also set a smaller default part size (8MB) for streaming download.

gcs_storage_client.go
sfc-gh-gren: Good catch. Changed to reuse the client.

gcs_storage_client.go
sfc-gh-gren: My understanding is that if we just return the stream handler from the go routine and write them into the buffer sequentially later, we are deferring the actual download to the write time, which is sequential instead of parallel. WDYT?

sfc-gh-pfus: Well, a client may give us the actual `io.Writer` with `WithFileGetStream`. So as long as we can write to this writer in correct order, when we have the actual chunk (still as stream, not byte array), it should be fine. Is it doable?

sfc-gh-gren: Changed to return stream from the go routine for the Streaming case. Please take a look. Thanks!

sfc-gh-gren:

file_transfer_agent.go
```diff
@@ -32,10 +32,11 @@ type (
 )
 
 const (
-	fileProtocol              = "file://"
-	dataSizeThreshold int64   = 64 * 1024 * 1024
-	isWindows                 = runtime.GOOS == "windows"
-	mb                float64 = 1024.0 * 1024.0
+	fileProtocol                      = "file://"
+	dataSizeThreshold         int64   = 64 * 1024 * 1024
+	streamingDownloadPartSize int64   = 8 * 1024 * 1024
```
sfc-gh-pfus: WDYT about aligning these two names? Something like `(streaming)multiPartThreshold`?

gcs_storage_client.go
```diff
@@ -303,6 +306,377 @@ func (util *snowflakeGcsClient) nativeDownloadFile(
...
+
+		// Collect results for this batch
+		batchResults := make([]downloadPartStream, batchSize)
+		for i := int64(0); i < batchSize; i++ {
+			result := <-results
+			if result.err != nil {
+				// Close any successful streams before returning error
+				for j := int64(0); j < i; j++ {
+					if batchResults[j].stream != nil {
+						_ = batchResults[j].stream.Close()
```
sfc-gh-pfus: Any error to log?

sfc-gh-pfus (approved):

file_transfer_agent.go
sfc-gh-gren: Makes sense. Done.

gcs_storage_client.go
sfc-gh-gren: Sg. Added logging.

sfc-gh-rsavenok (approved):

PR #1564, opened 2025-09-05
NO-SNOW Limit big.Float parsing memory

sfc-gh-pfus:
### Description

NO-SNOW Fixes memory allocation used when parsing numbers to big.Float (for JSON response, when `WithHigherPrecision` was used). Previously every big.Float occupied ~1GB of memory, now it is a manitude of bytes.

### Checklist

### DIFF
--- a/crl_test.go
+++ b/crl_test.go
@@ -1003,12 +1003,12 @@ func TestCrlE2E(t *testing.T) {
 	t.Run("Successful flow", func(t *testing.T) {
 		_ = logger.SetLogLevel("debug")
 		defer func() {
-			_ = logger.SetLogLevel("error")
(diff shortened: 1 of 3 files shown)


crl_test.go
sfc-gh-snoonan, line 1: Unrelated changes, right?

sfc-gh-snoonan (approved):

crl_test.go
sfc-gh-pfus: Actually pretty related. This CRL test was extremely slow due to memory consumption from the one that tested `big.Float`, and because of that, these timeouts here were very large. I want to reduce them to reasonable values now, when there is enough memory.

sfc-gh-rkowalski (approved):

Issue #1563, opened 2025-09-04
SNOW-2318982: Undocumented breaking change in SnowflakeTransport TLS config handling


saligrama:
Please answer these questions before submitting your issue. 
In order to accurately debug the issue this information is required. Thanks!

1. What version of GO driver are you using?

1.16.0
   
2. What operating system and processor architecture are you using?

MacOS 15.5 on M2 Max (though this issue also shows up in CI on Ubuntu 24.04 AMD64)
  
3. What version of GO are you using?
run `go version` in your console

1.25.0
 
4. Server version

This is connecting via a [Formal](https://joinformal.com) proxy.

5. What did you do?

Previously in v1.15.0 something like the following worked for specifying a custom TLS config (e.g., with custom cert pool) to connect:

```go
tlsConfig := &tls.Config{
    // custom fields here
} 

gosnowflake.SnowflakeTransport = &http.Transport{
    TLSClientConfig: tlsConfig,
}
```

However, on upgrading to v1.16.0, I get errors from the proxy of this nature:

```
2025/09/04 22:59:52 http: TLS handshake error from 192.168.107.1:43899: remote error: tls: bad certificate
```

which suggests that the custom certs in the cert pool from the client are not properly being passed.

The following *does* work:

```go
tlsConfigName := uuid.New().String()
gosnowflake.RegisterTLSConfig(tlsConfigName, tlsConfig)
config := &gosnowflake.Config{
    TLSConfigName: tlsConfigName
    // ...rest of the config
}
```

However, the changelog only mentions that v1.16.0 only "Added the `RegisterTLSConfig` function that lets you pass your own `TLSConfig` for the driver to use" and does not document that `gosnowflake.SnowflakeTransport.TLSClientConfig` is now broken.

6. What did you expect to see?

The previous behavior should work, or at least v1.16.0 should document the breaking change.

7. Can you set logging to DEBUG and collect the logs?

```
time="2025-09-04T16:27:25-07:00" level=info msg="Query String: account=account&database=test&ocspFailOpen=true&schema=test&validateDefaultParameters=false\n" func="gosnowflake.(*defaultLogger).Infof" file="log.go:192"
time="2025-09-04T16:27:25-07:00" level=info msg=OpenWithConfig func=gosnowflake.SnowflakeDriver.OpenWithConfig file="driver.go:57"
[...]

PR #1568, opened 2025-09-09
SNOW-2318982 Document custom TLS config usage in doc.go

sfc-gh-dszmolka:
### Description
This is a doc-only change. Added instructions for custom TLS configuration in the Go Snowflake Driver, same as described in https://github.com/snowflakedb/gosnowflake/issues/1563#issuecomment-3265469440

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -234,6 +234,21 @@ or Snowflake doc: https://docs.snowflake.com/en/developer-guide/snowflake-cli-v2
 
 If the connection.toml file is readable by others, a warning will be logged. To disable it you need to set the environment variable `SF_SKIP_WARNING_FOR_READ_PERMISSIONS_ON_CONFIG_FILE` to true.
 
+It you wish to specify a custom TLS config (e.g. to use the driver with a custom truststore) pass it through the `NewConnector`. Example:
+
+    tlsConfig := &tls.Config{
+        // your custom fields here
+    } 
+
+    config := &gosnowflake.Config{
+        Transporter: &http.Transport{
+            TLSClientConfig: tlsConfig,
+        },
+    }
+
+    connector := NewConnector(SnowflakeDriver{}, *cfg)
+    db := sql.OpenDB(connector)
+
 # Proxy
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -234,6 +234,21 @@ or Snowflake doc: https://docs.snowflake.com/en/developer-guide/snowflake-cli-v2
 
 If the connection.toml file is readable by others, a warning will be logged. To disable it you need to set the environment variable `SF_SKIP_WARNING_FOR_READ_PERMISSIONS_ON_CONFIG_FILE` to true.
 
+It you wish to specify a custom TLS config (e.g. to use the driver with a custom truststore) pass it through the `NewConnector`. Example:
```
sfc-gh-pfus: A couple of random thoughts:
1. I would rephrase it a bit to express that you can use your own transporter (e.g. to provide custom TLS config), as this is broader and not focused on one customer use case. WDYT?
2. Moreover, if you are documenting TLS configs, maybe you can also mention `RegisterTLSConfig`/`DeregisterTLSConfig`?

doc.go
```diff
@@ -234,6 +234,21 @@ or Snowflake doc: https://docs.snowflake.com/en/developer-guide/snowflake-cli-v2
...
+        // your custom fields here
+    } 
+
+    config := &gosnowflake.Config{
+        Transporter: &http.Transport{
+            TLSClientConfig: tlsConfig,
+        },
+    }
+
+    connector := NewConnector(SnowflakeDriver{}, *cfg)
```
sfc-gh-pfus: You prefixed `Config` with package, but not `SnowflakeDriver`. Can we align to use the same syntax in both?

doc.go
sfc-gh-dszmolka: 1. agreed, changed it

2. it wasn't the goal originally, but why not :) added

doc.go
sfc-gh-dszmolka: my bad, fixed

PR #1573, opened 2025-09-10
SNOW-2331330: use consistent log level for secure storage manager

### LINKED ISSUE #1572: Each Vault command print a warning-level message for credentials cache
Similar situation than in #1182 / SNOW-1552759

At each Vault command, a warning-level message documents credentials cache isn't supported on OS others than Linux, Darwin and Windows.

In the other hand, if it fails on Linux, only a debug message is printed, with the same choice in both case: switch to no op cache.

1.16.0
   
FreeBSD 14.3 (amd64)
   
go version go1.25.1 freebsd/amd64

n/a

5. What did you do?

$ vault status
`WARN[0000]log.go:204 gosnowflake.(*defaultLogger).Warnf OS freebsd does not support credentials cache`

Debug information, same level than a failure on Linux

dereckson:
### Description

SNOW-2331330. On Linux, when the secure storage manager can't be initialized, a debug level message is emitted.

Meanwhile, when the OS isn't supported, a warning was raised. This has the side effect to print on every CLI command using this library the warning, for example on Vault, when it's not the intent of the user to debug Snowflake connection.

Fixes #1572.

### Checklist
- [x] Created tests which fail without the change (if possible) - not applicable
- [x] Extended the README / documentation, if necessary - not applicable

### DIFF
--- a/secure_storage_manager.go
+++ b/secure_storage_manager.go
@@ -105,7 +105,7 @@ func newSecureStorageManager() secureStorageManager {
 	case "darwin", "windows":
 		return &threadSafeSecureStorageManager{&sync.Mutex{}, newKeyringBasedSecureStorageManager()}
 	default:
-		logger.Warnf("OS %v does not support credentials cache", runtime.GOOS)
+		logger.Debugf("OS %v does not support credentials cache", runtime.GOOS)
 		return newNoopSecureStorageManager()
 	}
 }


dereckson:

sfc-gh-pfus:
Thankyou @dereckson for your input, I merged your PR!

PR #1574, opened 2025-09-11
SNOW-1879120 Update OAuth token comment to include URL-escaping

sfc-gh-dszmolka:
Clarify that the OAuth token needs to be URL-escaped.

### Description

No code was changed, this is an edition to an existing comment.
Per the jira, relevant documentation was missing a bit which is already available elsewhere (e.g. [oauth.go](https://github.com/snowflakedb/gosnowflake/blob/8257f91fef2e7927cb5bf23f26a8ba09d02d8c2f/cmd/oauth/oauth.go#L34-L35)) so adding it to this relevant place too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Marked fields as deprecated (snowflakedb/gosnowflake#1556)
 - Exposed `QueryStatus` from `SnowflakeResult` and `SnowflakeRows` in `GetStatus()` function (snowflakedb/gosnowflake#1556)
 - Split timeout settings into separate groups based on target service types (snowflakedb/gosnowflake#1531)
+- Added small clarification in oauth.go example on token escaping (snowflakedb/gosnowflake#1574)
 
 ## 1.17.0
 


dsn.go
```diff
@@ -94,7 +94,7 @@ type Config struct {
 	InsecureMode bool             // driver doesn't check certificate revocation status
 	OCSPFailOpen OCSPFailOpenMode // OCSP Fail Open
 
-	Token            string        // Token to use for OAuth other forms of token based auth
+	Token            string        // Token to use for OAuth other forms of token based auth. Needs to be url-escaped. (e.g. url.QueryEscape(token))
```
sfc-gh-pfus: Have you tested it, by any chance? This field is used in so many places (old oauth, PAT, wif), I'm wondering if it behaves the same everywhere. Especially that we don't unescape it, but we pack it into a request body where I wouldn't guess escaping would be required...

sfc-gh-dszmolka: this was a request from @sfc-gh-jkasten , maybe he can provide more insight

sfc-gh-jkasten: Thanks David and Piotr! I originally filed this as I was trying to experiment with External OAuth a month into joining Snowflake. I was having difficulty getting everything to work (Generic "Invalid OAuth token" errors [[1](https://snowflakecomputing.atlassian.net/browse/SNOW-1867026)], DSN dropping the token field silently[[2](https://snowflakecomputing.atlassian.net/browse/SNOW-1879115)], and as I was searching for other examples in getting this to work, I came across the documentation David cited in oauth.go.

I know the thing I ended up getting working query escaped the OAuth token. Knowing what I now know about the internals, I would assume that this is an unnecessary step (the data is transmitted via JSON rather than the URL itself).

The bug may be in oauth.go's documentation.

sfc-gh-dszmolka: thanks @sfc-gh-jkasten so this PR is not necessary. But maybe another one is? 
> The bug may be in oauth.go's documentation.

do you have any specifics in mind?

sfc-gh-jkasten: Ahh... sorry. I took a closer look.

I initially took the comment 
```
Tokens must be escaped before being used in the DSN!
```
As a requirement on the dsn.go's Config token value rather than uniquely the DSN string.

I guess I might personally try to clarify the example [oauth.go](https://github.com/snowflakedb/gosnowflake/blob/8257f91fef2e7927cb5bf23f26a8ba09d02d8c2f/cmd/oauth/oauth.go#L34-L35) file to say...
```
All query parameters must be QueryEscaped before being used in a DSN string.
```

Yes, any string of query parameters should indicate that the person should URL encode them, but it might be helpful to be explicit if we intend for customers to construct their own DSN string. 

There is some documentation on an internal function about the need to URL encode, but nothing on the [exported function](https://github.com/snowflakedb/gosnowflake/blob/d7b781a3406f3c9875935a44dda6fd9bb8ed42b9/dsn.go#L401-L402)(s). The relation between Config/DSN and final internal login request might not be clear to people.

If we don't intend for customers to create their own DSN then we should just remove it from the example and plug the values directly into the Config.

Finally, I am also OK with chalking this up to user error (me). The only reason I got over to this example was due to the other issues referenced earlier.

sfc-gh-dszmolka: thank you for all the details and the suggestion , i added it exactly like this, to the oauth.go example. 
We did not yet have any other complaints (or i don't know about them) but this extra detail on the example will help hopefully others too. 

Anyhow, pls review and see if it's fitting the requirements now

sfc-gh-pfus (approved):

PR #1576, opened 2025-09-23
SNOW-2334860: Add support for CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY

sfc-gh-gmerticariu:
### Description

SNOW-2334860 

Add support for the CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY parameter which enables us to configure how ofter a heartbeat requests is emited by the client. 

In the existing implementation, a default of 60 min is used, which is too high for SPCS.  The refresh rate there is 30 min.

### Checklist

### DIFF
--- a/heartbeat.go
+++ b/heartbeat.go
@@ -11,19 +11,42 @@ import (
 )
 
 const (
-	// One hour interval should be good enough to renew tokens for four hours master token validity
-	heartBeatInterval = 3600 * time.Second
+	minHeartBeatInterval     = 900 * time.Second
+	maxHeartBeatInterval     = 3600 * time.Second
+	defaultHeartBeatInterval = 3600 * time.Second
 )
+
+func newDefaultHeartBeat(restful *snowflakeRestful) *heartbeat {
+	return newHeartBeat(restful, defaultHeartBeatInterval)
+}
+
+func newHeartBeat(restful *snowflakeRestful, heartbeatInterval time.Duration) *heartbeat {
+	logger.Debugf("Using heartbeat with custom interval: %v", heartbeatInterval)
+	if heartbeatInterval < minHeartBeatInterval {
+		logger.Warnf("Heartbeat interval %v is less than minimum %v, using minimum", heartbeatInterval, minHeartBeatInterval)
+		heartbeatInterval = minHeartBeatInterval
+	} else if heartbeatInterval > maxHeartBeatInterval {
+		logger.Warnf("Heartbeat interval %v is greater than maximum %v, using maximum", heartbeatInterval, maxHeartBeatInterval)
+		heartbeatInterval = maxHeartBeatInterval
+	}
+
+	return &heartbeat{
+		restful:           restful,
+		heartbeatInterval: heartbeatInterval,
+	}
+}
 
 type heartbeat struct {
 	restful      *snowflakeRestful
 	shutdownChan chan bool
+
+	heartbeatInterval time.Duration
 }
 
 func (hc *heartbeat) run() {
 	_, _, sessionID := safeGetTokens(hc.restful)
 	ctx := context.WithValue(context.Background(), SFSessionIDKey, sessionID)
-	hbTicker := time.NewTicker(heartBeatInterval)
+	hbTicker := time.NewTicker(hc.heartbeatInterval)
 	defer hbTicker.Stop()
 	for {
 		select {
--- a/connection.go
+++ b/connection.go
@@ -46,10 +46,11 @@ const (
 )
 
 const (
-	sessionClientSessionKeepAlive          = "client_session_keep_alive"
-	sessionClientValidateDefaultParameters = "CLIENT_VALIDATE_DEFAULT_PARAMETERS"
-	sessionArrayBindStageThreshold         = "client_stage_array_binding_threshold"
-	serviceName                            = "service_name"
+	sessionClientSessionKeepAlive                   = "client_session_keep_alive"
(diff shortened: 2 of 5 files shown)


heartbeat.go
```diff
@@ -11,19 +11,36 @@ import (
 )
 
 const (
-	// One hour interval should be good enough to renew tokens for four hours master token validity
-	heartBeatInterval = 3600 * time.Second
+	minHeartBeatInterval = 600 * time.Second
```
sfc-gh-dszmolka: in other drivers, and also snowflake documentation (https://docs.snowflake.com/en/sql-reference/parameters#client-session-keep-alive-heartbeat-frequency) this is 900 seconds

**however** it makes sense (and sometimes even a requirement) to send a heartbeat even more frequently , like to prevent a proxy's idle session teardown , and such. 

so we should consider either match it with other drivers (900) or allowing setting it even lower like 2 or 5 minutes. Perhaps Drivers team can weigh in and provide insight.

heartbeat.go
```diff
@@ -11,19 +11,36 @@ import (
...
+	maxHeartBeatInterval = 3600 * time.Second
 )
 
+func newDefaultHeartBeat(restful *snowflakeRestful) *heartbeat {
+	return newHeartBeat(restful, maxHeartBeatInterval)
+}
+
+func newHeartBeat(restful *snowflakeRestful, heartbeatInterval time.Duration) *heartbeat {
```
sfc-gh-dszmolka, line 23: in Python, if this is less than min, we set it to min. If its bigger than max, we set it to max. ([ref](https://github.com/snowflakedb/snowflake-connector-python/blob/v3.17.4/src/snowflake/connector/connection.py#L2015-L2033))

sfc-gh-gmerticariu: After digging a bit on how this is set in Golang I noticed that the parameter will be returned after the "authenticate" call is performed. In other words, when I set the `CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY`, it will get appended to the DSN and send to the Snowflake server. The snowflake server will validate and will return a value between 900 and 3600 if the value is valid, according to the documentation (https://docs.snowflake.com/en/sql-reference/parameters#client-session-keep-alive-heartbeat-frequency). 

I tried setting it to a different value (e.g. 400) and every time I receive an error: `failed to connect to Snowflake: 001008 (08004): SQL compilation error:\ninvalid value [400] for parameter 'CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY'"`

heartbeat.go
sfc-gh-gmerticariu: I updated the code and removed the check all together as we don't have any control over the value in the driver. We receive the param value from the first authentication call and it is validated server side. Any value outside of 900-3600 interval will result into an error.

heartbeat.go
sfc-gh-dszmolka: I mean technically it should be possible to send `/session/heartbeat` calls on an arbitrary frequency based on the config value (without having SF to validate it - i tested it in python and it works) but probably makes most sense to keep parity with other drivers and the SF controlled values for now, 900 as the min ?

sfc-gh-gmerticariu: That's my take as well, but I will wait for imput from the drivers team.

heartbeat.go
```diff
@@ -11,19 +11,39 @@ import (
...
+	maxHeartBeatInterval     = 3600 * time.Second
+	defaultHeartBeatInterval = 3600 * time.Second
 )
 
+func newDefaultHeartBeat(restful *snowflakeRestful) *heartbeat {
+	return newHeartBeat(restful, defaultHeartBeatInterval)
+}
+
+func newHeartBeat(restful *snowflakeRestful, heartbeatInterval time.Duration) *heartbeat {
+	if heartbeatInterval < minHeartBeatInterval {
```
sfc-gh-rsavenok, line 25: The same if/else is present in other drivers, but it feels weird to me that we change values provided by the developer without any information message or error. 

Since this is a constant, I'd suggest to put min+max values to doc.go

heartbeat.go
sfc-gh-pfus: I'm not sure this is the best idea. This is a backend parameter. If there should be min/max validation, it should be during parameter setting on the backend, not in clients.

heartbeat.go
sfc-gh-pfus: As above - I don't think we should document it in driver. This is not a driver parameter, this is a backend parameter. Backend parameter and its validation can evolve independently from drivers. It should be document on backend side, not here.

heartbeat.go
sfc-gh-rsavenok: 3 arguments for driver validation:
* this is already implemented in other drivers
* min max ranges are documented on https://docs.snowflake.com/en/sql-reference/parameters#client-session-keep-alive-heartbeat-frequency

heartbeat.go
sfc-gh-gmerticariu: @sfc-gh-pfus, @sfc-gh-rsavenok  since the parameter is always comming from the Authorziation, request it's guaranteed (except if there is a bug server side) that the value will be between 900 and 3600. Therefore this code won't be hit, but imo it's good to have a guard regardless.

Should the common ground be: keep the guards, but remove the mention in the docs and the warning messages? Again, the if should never be hit, as long as the server doesn't have any bugs.

sfc-gh-rsavenok: I think it should be kept as this is how it's implemented in other drivers, and most likely that's how it will be migrated to the universal driver

PR #1557, opened 2025-08-29
SNOW-2306340: Add pki-oversight as an additional owner for PKI-related files

sfc-gh-jkasten:
### Description

[SNOW-2306340](https://snowflakecomputing.atlassian.net/browse/SNOW-2306340)

Add an additional owner group for transport.go, ocsp.go, and crl.go. Please let me know if I got the mapping wrong. I am happy to adjust this or scope it down.

If this is too onerous, we can also look at just adding the group as an additional reviewer.

### Checklist
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

[SNOW-2306340]: https://snowflakecomputing.atlassian.net/browse/SNOW-2306340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1581, opened 2025-09-29
SNOW-2306340: Add pki-oversight as an additional owner for PKI-related files

sfc-gh-jkasten:
### Description

[SNOW-2306340](https://snowflakecomputing.atlassian.net/browse/SNOW-2306340)

Add an additional owner group for transport.go, ocsp.go, and crl.go. This is a direct copy of the previously approved https://github.com/snowflakedb/gosnowflake/pull/1557.

### Checklist
- [x] Created tests which fail without the change (if possible) - (N/A)
- [x] Extended the README / documentation, if necessary - (N/A)


.github/CODEOWNERS
```diff
@@ -1 +1,4 @@
 * @snowflakedb/Client
+/transport.go @snowflakedb/pki-oversight
```
sfc-gh-snoonan: The problem with this approach is that we'll need two folks from pki-oversight to approve any changes here due to repo settings. Given the group is 3 people, I'm a bit worried about the failure modes. What do you think?

sfc-gh-jkasten: I was unaware of that requirement. I think having 1 review from the group for every change is already a high bar. I would love to just be cc'd, but when I originally looked that up, it seemed as though we would be owning a custom hook. I guess I will spend more time researching it when I eventually get a chance.

sfc-gh-snoonan: We could temporarily do this as joint ownership. pki-oversight would get added by default as a reviewer, and the Client team would as well. Yes, they could bypass it with just two reviews, but it's still an improvement over just community knowledge and we can follow up with tighter ownership in the future.

Concretely, just make this `@snowflakedb/pki-oversight,@snowflakedb/Client` instead?

sfc-gh-jkasten: Thanks. That sounds reasonable. I have pushed the associated changes (using a space as a delimiter following this [documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners))

sfc-gh-snoonan (approved):

sfc-gh-rsavenok (approved):

Issue #1292, opened 2025-01-24
SNOW-2298165 SNOW-1896153 Non deterministic errors with concurrent context aware queries in v1.12.1


niger-prequel:
> 1. What version of GO driver are you using?

`1.12.1` and `1.12.0`
   
> 2. What operating system and processor architecture are you using?

Debian Bullseye x86
   
> 3. What version of GO are you using?

`go1.23.3`

> 4. Server version:* E.g. 1.90.1

`9.1.0`

> 5. What did you do?

We are loading data into snowflake via the golang data. Our strategy is to use `PUT...` SQL commands to upload parquet files to an internal stage. then we use a `COPY INTO...` statement to publish the data. We leverage the `database/sql` golang abstraction . So generally our code will look like 

```go
package example

import (
    "context"
    "database/sql"
    "fmt"
    "log"
    "sync"
    "time"

    _ "github.com/snowflakedb/gosnowflake" // Snowflake driver
)

func UploadFilesAndCopy(db *sql.DB, stageName, tableName string, files []string) error {
    // Create a context that will be used for all operations
    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
    defer cancel()

    // A channel that will hold the file paths we want to upload
    fileCh := make(chan string)

    // We'll use a WaitGroup to ensure all goroutines finish
    var wg sync.WaitGroup

    // Start a fixed number of worker goroutines to process the file uploads
    workerCount := 5
    for i := 0; i < workerCount; i++ {
        wg.Add(1)
        go func() {
            defer wg.Done()
            for file := range fileCh {
                putQuery := fmt.Sprintf(
                    "PUT file://%s @%s AUTO_COMPRESS=TRUE OVERWRITE=TRUE",
                    file, stageName,
                )
                if _, err := db.ExecContext(ctx, putQuery); err != nil {
                    log.Printf("failed to PUT file %q to stage %q: %v", file, stageName, err)
                }
            }
        }()
    }

    // Send file names into the channel for workers to pick up
    go func() {
        defer close(fileCh) // Close once we're done sending
        for _, file := range files {
            fileCh <- file
        }
    }()

    // Wait for all workers to finish uploading
    wg.Wait()

    // Now that all files are in the stage, run the COPY INTO command
    copyQuery := fmt.Sprintf(`COPY INTO %s FROM @%s FILE_FORMAT = (TYPE = PARQUET)`, tableName, stageName)
[...]

PR #1584, opened 2025-10-08
SNOW-2298165 Fix wrong context being used in updateFileMetadataWithPresignedURL

sfc-gh-rsavenok:
### Description

Addresses https://github.com/snowflakedb/gosnowflake/issues/1292 

The `updateFileMetadataWithPresignedURL` function for GCP was incorrectly using the database connection's context instead of the file transfer agent's context. This caused a bug where cancelling the connection context would also cancel any ongoing/future PUT/GET operations

### Checklist

### DIFF
--- a/restful.go
+++ b/restful.go
@@ -234,6 +234,8 @@ func postRestfulQueryHelper(
 
 	var resp *http.Response
 	fullURL := sr.getFullURL(queryRequestPath, params)
+
+	logger.WithContext(ctx).Infof("postQuery: make a request to Host: %v, Path: %v", fullURL.Host, fullURL.Path)
 	resp, err = sr.FuncPost(ctx, sr, fullURL, headers, body, timeout, defaultTimeProvider, cfg)
 	if err != nil {
 		return nil, err
@@ -245,7 +247,6 @@ func postRestfulQueryHelper(
 	}(resp, fullURL.String())
 
 	if resp.StatusCode == http.StatusOK {
-		logger.WithContext(ctx).Infof("postQuery: resp: %v", resp)
 		var respd execResponse
 		if err = json.NewDecoder(resp.Body).Decode(&respd); err != nil {
 			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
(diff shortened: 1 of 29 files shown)


driver_test.go
```diff
@@ -1885,6 +1886,31 @@ func TestCancelQuery(t *testing.T) {
 	})
 }
 
+func TestCancelQueryWithImplicitConnectionContext(t *testing.T) {
```
sfc-gh-pfus: Maybe it's worth having a test for explicit case too (`db.Conn(cancellableCtx)`)?

restful.go
```diff
@@ -202,7 +203,7 @@ func postRestfulQuery(
 
 	data, err = sr.FuncPostQueryHelper(ctx, sr, params, headers, body, timeout, requestID, cfg)
 
-	if err == context.Canceled || err == context.DeadlineExceeded {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		// For context cancel/timeout cases, a special cancel request needs to be sent.
 		if cancelErr := sr.FuncCancelQuery(context.Background(), sr, requestID, timeout); cancelErr != nil {
 			// Wrap the original error with the cancel error.
```
sfc-gh-pfus, line 209: I think there are a couple of places where we have the same problem:
1. `Begin` of transaction
2. `Prepare` of statement
3. `Exec`/`Query` without context
4. `renewExpiredClient`
WDYT?

driver_test.go
sfc-gh-rsavenok: Added 2 test cases

driver_test.go
```diff
@@ -1888,6 +1889,53 @@ func TestCancelQuery(t *testing.T) {
 	})
 }
 
+func TestCancelQueryWithConnectionContext(t *testing.T) {
+	testCases := []struct {
+		name           string
+		setupConnection func(db *sql.DB, ctx context.Context) error
```
sfc-gh-pfus: Good practice is to have context as a first parameter.

sfc-gh-pfus (approved):

PR #1588, opened 2025-10-15
SNOW-2395141 Add limiting to CRL download size and ensure better permissions for CRLs

sfc-gh-pfus:
### Description

SNOW-2395141 This PR brings a new property that allows limiting CRL download to a particular size and ensures more strict permissions on CRL cache directory.

### Checklist

### DIFF
--- a/crl.go
+++ b/crl.go
@@ -431,7 +431,8 @@ func (cv *crlValidator) updateCache(crlURL string, crl *x509.RevocationList, dow
 		if !errors.Is(err, os.ErrExist) {
 			logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
 			return
-		} else if err := os.Chmod(crlDirPath, 0700); err != nil {
+		}
+		if err = os.Chmod(crlDirPath, 0700); err != nil {
 			logger.Warnf("failed to chmod existing directory for CRL file %v: %v", crlFilePath, err)
 			return
 		}


crl.go
```diff
@@ -418,10 +420,23 @@ func (cv *crlValidator) updateCache(crlURL string, crl *x509.RevocationList, dow
...
 	crlFilePath := cv.crlURLToPath(crlURL)
-	if err := os.MkdirAll(filepath.Dir(crlFilePath), 0755); err != nil {
+	crlDirPath := filepath.Dir(crlFilePath)
+	crlDirParentPath := filepath.Dir(crlDirPath)
+	if err := os.MkdirAll(crlDirParentPath, 0755); err != nil {
 		logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
 		return
 	}
+	if err := os.Mkdir(crlDirPath, 0700); err != nil {
+		if errors.Is(err, os.ErrExist) {
```
sfc-gh-snoonan: I think this is easier to follow with a ! here.
```
if !errors.Is(err, os.ErrExist) {
    warn and return
}
// path exists, try chmod
if err := os.Chmod(crlDirPath, 0700); err != nil {
    warn and return
}
```

That said, I think we have the same logic for other directories we manage? Should we split this out into a writeSecureFile(path) helper - or at least the directory creation part?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 - Added temporary download files cleanup (snowflakedb/gosnowflake#1577)
 - Marked fields as deprecated (snowflakedb/gosnowflake#1556)
 - Exposed `QueryStatus` from `SnowflakeResult` and `SnowflakeRows` in `GetStatus()` function (snowflakedb/gosnowflake#1556)
+- Ensured proper permissions for CRL cache directory (snowflakedb/gosnowflake#1588)
+- Added `CrlDownloadMaxSize` to limit the size of CRL downloads (snowflakedb/gosnowflake#1588)
```
sfc-gh-snoonan, line 13: I understand this came out of the security review, but it seems relatively silly, especially if we're defaulting to unlimited (and I can't think of an absolutely safe value we could put here that would make a difference - maybe 1GB :-)?).
If a CRL exceeds this limit, then a client in fail closed mode will fail to connect anyway.

If we're leaving this unlimited by default, it because another configuration knob we have to maintain, and I think it is the wrong tradeoff.

crl.go
sfc-gh-pfus: re code reorg - ok, done.
re unifying - it's more complex, I think it's not correct now in other places. We have multiple files that we handle (OCSP cache, CRL cache, token cache, logging config and so on), and everywhere it's a bit different regarding permissions of files and dirs... It would be good to unify it, but I think we can leave it as-is for now, keeping in mind our long-term roadmap.

CHANGELOG.md
sfc-gh-pfus: I'm also not a fan of this. I can set the default to something reasonable like 200MB. We need to remember, that it takes ~3x in memory after parsing. Is it fine for you?

crl_test.go
```diff
@@ -73,6 +77,7 @@ func TestCrlCheckModeDisabledNoHttpCall(t *testing.T) {
 }
 
 func TestCrlModes(t *testing.T) {
+	logger.SetLogLevel("debug")
```
sfc-gh-rsavenok: Is this intentional? If so, I think it's not reset after the test

sfc-gh-rsavenok (approved):

crl.go
```diff
@@ -418,10 +421,21 @@ func (cv *crlValidator) updateCache(crlURL string, crl *x509.RevocationList, dow
...
+	crlDirParentPath := filepath.Dir(crlDirPath)
+	if err := os.MkdirAll(crlDirParentPath, 0755); err != nil {
 		logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
 		return
 	}
+	if err := os.Mkdir(crlDirPath, 0700); err != nil {
+		if !errors.Is(err, os.ErrExist) {
+			logger.Warnf("failed to create directory for CRL file %v: %v", crlFilePath, err)
+			return
+		} else if err := os.Chmod(crlDirPath, 0700); err != nil {
```
sfc-gh-snoonan: nit: you don't need the else; it's a separate block as you always return above.

sfc-gh-snoonan (approved):

Issue #1590, opened 2025-10-17
Nil pointer dereference in restful.go SNOW-2439606


obs-gh-felixbowing:
As described in this [comment](https://github.com/snowflakedb/gosnowflake/commit/11d879235177a2a9c2906214db79f7e122575f77#r168136170), there is an issue in restful.go that can lead to a nil pointer panic. The problem is that a regular call to body.close() was replaced with a deferred call.

I will propose a fix in the next hours.

sfc-gh-dszmolka:
hey @obs-gh-felixbowing  - we're already started investigating it due to https://github.com/snowflakedb/terraform-provider-snowflake/issues/4092 (and consequently https://github.com/snowflakedb/gosnowflake/pull/1587), but if you have a reproduction which you could share on pure gosnowflake, that would be highly appreciated !

obs-gh-felixbowing:
I don't have a reproduction, but I could reconstruct what happened from the stack trace. The issue lies in the deferred resp.body.Close() statement that is run inside a loop. Now if in one iteration the http call returns an error and the loop ends, we still call resp.body.Close() from previous iterations. But now resp is nil as the http call returned an error and we get a nil pointer dereference panic. 

The clean solution is to do exactly one http call + body reading in a separate function. The loop or any complex function should just call this function.

Side note: There is also a much less likely problem caused by the fact that there are two deferred calls to close() in the same function. The mentioned clean solution from above solves this as well. 

I will prepare a PR to show what I mean.

sfc-gh-dszmolka:
Thank you for your contribution on https://github.com/snowflakedb/gosnowflake/pull/1591 ! The team will look.

sfc-gh-pfus:
Hi Felix! We identified this bug and potential repro on Friday evening, today I finished working on this. I think we should merge my PR, because it includes repro and tests of the bug and also some test reorg. However, I'd like to still express gratitude for your work here, I know this is not your first PR and I hope not the last!

sfc-gh-dszmolka:
(ref  https://github.com/snowflakedb/gosnowflake/pull/1592 and https://github.com/snowflakedb/gosnowflake/pull/1596)

obs-gh-felixbowing:
Hi @sfc-gh-pfus, thanks for fixing this bug. 
I saw that you have kept the defer statement in the loop. Why? I think it has two disadvantages:
- It closed the bodies unnecessarily late. Hence it needs more memory.
[...]

PR #1591, opened 2025-10-18
Fix nil pointer panic in issue #1590

obs-gh-felixbowing:
The issue lies in the deferred resp.body.Close() statement that is run inside a loop. Now if in one iteration the http call returns an error and the loop ends, we still call resp.body.Close() from previous iterations. But now resp is nil as the http call returned an error and we get a nil pointer dereference panic.

The clean solution is to do exactly one http call + body reading in a separate function. The loop or any complex function should just call this function.

### DIFF
--- a/restful.go
+++ b/restful.go
@@ -213,6 +213,32 @@ func postRestfulQuery(
 	return data, err
 }
 
+// fetchAndDecodeExecResponse fetches a response using the provided request function and decodes it into an execResponse.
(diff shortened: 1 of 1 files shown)


sfc-gh-pfus:
Closing this one in favour of https://github.com/snowflakedb/gosnowflake/pull/1592 Thanks anyway, community PRs are always welcome!

Issue #1182, opened 2024-07-23
SNOW-1552759: Super annoying warning message from gosnowflake every invocation of Hashicorp vault cli under sudo


DrDaveD:
This message is displayed every time the Hashicorp vault-1.17.2 cli is invoked under sudo to root, at least on RHEL8:

WARN[0000]log.go:244 gosnowflake.(*defaultLogger).Warn DBUS_SESSION_BUS_ADDRESS envvar looks to be not set, this can lead to runaway dbus-daemon processes. To avoid this, set envvar DBUS_SESSION_BUS_ADDRESS=$XDG_RUNTIME_DIR/bus (if it exists) or DBUS_SESSION_BUS_ADDRESS=/dev/null.

This message comes from gosnowflake and was reported in vault at hashicorp/vault#22560 but that issue was closed saying to report here instead.  Most of the questions in the standard gosnowflake issue template aren't relevant, but this was compiled with go version 1.22.5, and the go.mod for vault-1.17.2 says it is using github.com/snowflakedb/gosnowflake v1.10.0.

When logged in with ssh to a RHEL8 linux server, DBUS_SESSION_BUS_ADDRESS is set so the warning is not emitted, but the variable does not carry over to `sudo -s` so it is emitted there.  However I have never seen any evidence of extra dbus-daemon processes when running vault from `sudo -s` on Rocky8 or Alma8.  There is a dbus-daemon process running on the system as the `dbus` user and that gets detected by the code added in #1058 so the message is emitted.    I tried to reproduce the original problem by unsetting DBUS_SESSION_BUS_ADDRESS and removing $XDG_RUNTIME_DIR/bus but I did not get any extra dbus-daemon processes with either vault or by running select1.go as suggested in #773.  Maybe I'm not running the latter test right; it required setting values for `SNOWFLAKE_TEST_{ACCOUNT|USER|PASSWORD}` which I set to arbitrary values and it just seemed to hang then when I did `go run cmd/select1/select1.go` in a clone of this git repository.

[...]

Issue #773, opened 2023-04-13
Zombie dbus processes generated by github.com/99designs/keyring


mihaitodor:
I noticed that `gosnowflake` added https://github.com/99designs/keyring as a dependency and, unfortunately, this library has an [outstanding bug](https://github.com/99designs/keyring/issues/103) which can lead some Linux systems to create zombie dbus processes. This issue was observed initially when importing the Apache Pulsar client library into Benthos as described [here](https://github.com/benthosdev/benthos/issues/1184) and it's triggered in one of the `init()` functions from `github.com/99designs/keyring`. One workaround that I identified is to set `DBUS_SESSION_BUS_ADDRESS=/dev/null` on process startup.

I tried to reproduce it in a Docker container, but didn't succeed yet, so I opened this issue to raise awareness and I'll share detailed reproduction steps at a later time if I get more time to poke around with it.

sfc-gh-dszmolka:
hi @mihaitodor (or anyone else stumbling into this issue) - first of all, thank you so much for raising this with us and for providing the details about the issue and even the workaround to tackle it. This is highly appreciated.

I would recommend marking this as closed until a reproduction is available. The issue can be still commented to, even in Closed state, and I can reopen it when it becomes more actionable with the repro. Thank you in advance !

crflanigan:
This issue appears to still exist:
https://github.com/influxdata/telegraf/issues/13481

sfc-gh-dszmolka:
hi folks - I see sometimes people stumble into this issue; so apparently it's still here. We're taking a look.

mihaitodor:
It hasn't been resolved, unfortunately... We're forcing a fork of `github.com/99designs/keyring` in Benthos: https://github.com/benthosdev/benthos/blob/bde063a467bd5e783c525365cb13ac239c0eb4e5/go.mod#L3, but that breaks `go install`, which is quite annoying:

```shell
> go install github.com/benthosdev/benthos/v4/cmd/benthos@latest
go: downloading github.com/benthosdev/benthos/v4 v4.20.0
go: github.com/benthosdev/benthos/v4/cmd/benthos@latest (in github.com/benthosdev/benthos/v4@v4.20.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

Related issues:
- https://github.com/99designs/keyring/issues/103
- https://github.com/apache/pulsar-client-go/issues/836
[...]

PR #1595, opened 2025-10-22
SNOW-1553756 split secure store manager into keyring and non-keyring parts 

sfc-gh-dszmolka:
### Description

This is for
* https://github.com/snowflakedb/gosnowflake/issues/1183
* https://github.com/snowflakedb/gosnowflake/issues/773
* https://github.com/snowflakedb/gosnowflake/issues/1182

Issue is, due to a long unfixed problem (see above issues) in `99design/keyring` , when the library is initialized, on certain Linux it can generate a runaway orphaned dbus process which remains even there when the gosnowflake process exists. Multiple invocations of gosnowflake (or any dependant library) leads to multiple, even hundreds of such processes lingering around. 

Sadly the keyring is currently initialized sooner than the driver itself , so the generation of the orphaned dbus process can not be avoided this way. Which is even more sad that for Linux, we don't even use keyring, so the issue is triggered for nothing.

Workaround for a long time is to set `DBUS_SESSION_BUS_ADDRESS` envvar on the affected distros.

This PR, as discussed with the owner team here, aims to split the functionality into multiple files, where we only initialize keyring for the OS for which we actually use it: Windows and Mac. Storing these credentials on Linux is file-based, no keyring needed there. 

Work done here is no more than reorganizing the functionality into separate files, and adding 2 loglines. 

#### Manual tests 
Re-reproduced issue on RHEL 7.9 , then validated fix using this branch on same OS. With this branch, no more zombie `dbus-daemon` processes are spawn. Step by step validation details in internal jira.

### DIFF
--- a/secure_storage_manager_linux.go
+++ b/secure_storage_manager_linux.go
@@ -1,0 +1,18 @@
+//go:build linux
+
+package gosnowflake
+
+import (
+	"runtime"
+	"sync"
+)
+
+func defaultOsSpecificSecureStorageManager() secureStorageManager {
+	logger.Debugf("OS is %v, using file based secure storage manager.", runtime.GOOS)
+	ssm, err := newFileBasedSecureStorageManager()
+	if err != nil {
+		logger.Debugf("failed to create credentials cache dir: %v. Not storing credentials locally.", err)
+		return newNoopSecureStorageManager()
+	}
+	return &threadSafeSecureStorageManager{&sync.Mutex{}, ssm}
+}
--- a/secure_storage_manager_notlinux.go
+++ b/secure_storage_manager_notlinux.go
@@ -6,7 +6,19 @@ import (
 	"github.com/99designs/keyring"
 	"runtime"
 	"strings"
+	"sync"
 )
+
+func defaultOsSpecificSecureStorageManager() secureStorageManager {
+	switch runtime.GOOS {
+	case "darwin", "windows":
+		logger.Debugf("OS is %v, using keyring based secure storage manager.", runtime.GOOS)
+		return &threadSafeSecureStorageManager{&sync.Mutex{}, newKeyringBasedSecureStorageManager()}
+	default:
+		logger.Debugf("OS %v does not support credentials cache", runtime.GOOS)
+		return newNoopSecureStorageManager()
+	}
+}
 
 type keyringSecureStorageManager struct {
 }
--- a/secure_storage_manager.go
+++ b/secure_storage_manager.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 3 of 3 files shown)


secure_storage_manager_linux_test.go
```diff
@@ -0,0 +1,226 @@
...
+		assertNilF(t, err)
+		assertEqualE(t, path, filepath.Join(testRoot2, "sub1", "sub2"))
+		stat, err := os.Stat(testRoot2)
+		assertNilF(t, err)
+		assertEqualE(t, stat.Mode(), 0700|os.ModeDir)
+	})
+}
+
+func TestSnowflakeFileBasedSecureStorageManager(t *testing.T) {
+	skipOnWindows(t, "file system permission is different")
```
sfc-gh-pfus: Should never happen, right?

secure_storage_manager_linux.go
```diff
@@ -0,0 +1,370 @@
...
+	pathSegments []string
+}
+
+var defaultLinuxCacheDirConf = []cacheDirConf{
+	{envVar: credCacheDirEnv, pathSegments: []string{}},
+	{envVar: "XDG_CACHE_DIR", pathSegments: []string{"snowflake"}},
+	{envVar: "HOME", pathSegments: []string{".cache", "snowflake"}},
+}
+
+func osSpecificSecureStorageManager() secureStorageManager {
```
sfc-gh-pfus: WDYT about leaving the whole implementation in the common file (as it was previously), moving only this function to the separate files? Reasoning - now, on Mac we won't be able to test this functionality. I know that keyring/keychain is problematic, but I'd like to have the possibility to run file-based impl tests on my Mac too. So my overall suggestion:
1. Move file-based implementation to `secure_storage_manager.go`.
2. Move keyring-based implementation t- `..._notlinux.go`.
3. Leave `osSpecificSecureStorageManager` to files - as it is now. We can even rename it `defaultOsSpecificSecureStorageManager`.

WDYT?

secure_storage_manager_linux_test.go
sfc-gh-dszmolka: separate linux test file has been deleted during addressing the review

secure_storage_manager_linux.go
sfc-gh-dszmolka: sounds good, restructured per advice

secure_storage_manager_notlinux.go
```diff
@@ -0,0 +1,131 @@
...
+			ring, _ := keyring.Open(keyring.Config{
+				WinCredPrefix: strings.ToUpper(tokenSpec.host),
+				ServiceName:   strings.ToUpper(tokenSpec.user),
+			})
+			item := keyring.Item{
+				Key:  credentialsKey,
+				Data: []byte(value),
+			}
+			if err := ring.Set(item); err != nil {
+				logger.Debugf("Failed to write to Windows credential manager. Err: %v", err)
```
sfc-gh-mhofman, line 50: nit: Error instead of Debug

sfc-gh-dszmolka: I didn't change any part of this code so it's the same as originally implemented ([reference](https://github.com/snowflakedb/gosnowflake/blame/99e518b9f74971d4059558f8759ef5963621491e/secure_storage_manager.go#L496)) and the same as other 'failed to <do something with keychain>' which are also Debug.

@sfc-gh-pfus what is your preference, should we change this to a stronger, error level message or leave it like this?
For context , since the caching was implemented 3 years ago, I don't know of user complaints around the current loglevels. They usually prefer to see stuff on DEBUG level if it's not a blocker error.

PR #1597, opened 2025-10-27
SNOW-2208212: RHEL support in Go

sfc-gh-pcyrek:
### Description

SNOW-2208212: RHEL support in Go

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,11 @@
 - 
 - 
 - 
-- Fixed nil pointer derereference while calling long-running queries (snowflakedb/gosnowflake#1592) (snowflakedb/gosnowflake#1596)
+- Fixed nil pointer dereference while calling long-running queries (snowflakedb/gosnowflake#1592) (snowflakedb/gosnowflake#1596)
 - 
 - 
 - 
-- 
+- Moved keyring-based secure storage manager into separate file to avoid the need to initialize keyring on Linux. (snowflakedb/gosnowflake#1595)
 - 
 - 
 
--- a/secure_storage_manager.go
+++ b/secure_storage_manager.go
@@ -10,13 +10,10 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/99designs/keyring"
 )
 
 type tokenType string
@@ -94,20 +91,7 @@ type secureStorageManager interface {
 var credentialsStorage = newSecureStorageManager()
 
 func newSecureStorageManager() secureStorageManager {
-	switch runtime.GOOS {
-	case "linux":
-		ssm, err := newFileBasedSecureStorageManager()
-		if err != nil {
-			logger.Debugf("failed to create credentials cache dir. %v", err)
-			return newNoopSecureStorageManager()
-		}
-		return &threadSafeSecureStorageManager{&sync.Mutex{}, ssm}
-	case "darwin", "windows":
-		return &threadSafeSecureStorageManager{&sync.Mutex{}, newKeyringBasedSecureStorageManager()}
-	default:
-		logger.Debugf("OS %v does not support credentials cache", runtime.GOOS)
-		return newNoopSecureStorageManager()
-	}
+	return defaultOsSpecificSecureStorageManager()
 }
 
 type fileBasedSecureStorageManager struct {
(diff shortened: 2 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -1,5 +1,8 @@
 name: Build and Test
 
+permissions:
+  contents: read
```
sfc-gh-pfus, line 4: What is this for? Can we have a short comment?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+FROM $BASE_IMAGE
+
+# Get target architecture for multi-arch builds
+ARG TARGETARCH
+
+# Install minimal required packages + gcc for CGO (race detection)
+RUN dnf install -y --allowerasing \
+    gcc \
+    glibc-devel \
+    java-11-openjdk \
```
sfc-gh-pfus, line 10: What is this for?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    jq \
+    tar \
+    gzip \
+    && dnf clean all
+
+# Set Java 11 as the default using environment variables
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+ENV PATH="${JAVA_HOME}/bin:${PATH}"
+
+# Install Go versions 1.23, 1.24, and 1.25 (multi-arch aware)
```
sfc-gh-pfus: Do we need this? Maybe we can run a smoke test on 1.25? I don't want to update it every half a year.

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+ENV GO_VERSIONS="1.23.4 1.24.2 1.25.0"
+
+# Download and install Go versions for the target architecture
```
sfc-gh-pfus: Isn't there a version available via yum?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+# Get target architecture for multi-arch builds
+ARG TARGETARCH
+
+# Install minimal required packages + gcc for CGO (race detection)
+RUN dnf install -y --allowerasing \
+    gcc \
+    glibc-devel \
+    java-11-openjdk \
+    python3 \
```
sfc-gh-pfus, line 11: And this?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    for version in $GO_VERSIONS; do \
+        echo "Installing Go $version for ${GOARCH}..."; \
+        wget -q https://golang.org/dl/go${version}.linux-${GOARCH}.tar.gz -O /tmp/go${version}.tar.gz; \
+        mkdir -p /usr/local/go${version}; \
+        tar -C /usr/local/go${version} --strip-components=1 -xzf /tmp/go${version}.tar.gz; \
+        rm /tmp/go${version}.tar.gz; \
+    done
+
+# Create wrapper scripts for Go versions (simplified)
+RUN for version in 1.23.4 1.24.2 1.25.0; do \
```
sfc-gh-pfus: Don't repeat yourself.

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+        echo "#!/bin/bash" > /usr/local/bin/go${short_version}; \
+        echo "export GOROOT=/usr/local/go${version}" >> /usr/local/bin/go${short_version}; \
+        echo 'exec $GOROOT/bin/go "$@"' >> /usr/local/bin/go${short_version}; \
+        chmod +x /usr/local/bin/go${short_version}; \
+    done
+
+# Ensure /usr/local/bin is in PATH (should be by default, but making sure)
+ENV PATH="/usr/local/bin:${PATH}"
+
+# This is to solve permission issue, read https://denibertovic.com/posts/handling-permissions-with-docker-volumes/
```
sfc-gh-pfus: This is for entrypoint, right?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+ARG GOSU_URL=https://github.com/tianon/gosu/releases/download/1.14/gosu-amd64
+ENV GOSU_PATH=${GOSU_URL}
+RUN GOARCH=${TARGETARCH:-amd64} && \
+    GOSU_ARCH=$([ "$GOARCH" = "amd64" ] && echo "amd64" || echo "arm64") && \
+    curl -o /usr/local/bin/gosu -SL https://github.com/tianon/gosu/releases/download/1.14/gosu-${GOSU_ARCH}
+RUN chmod +x /usr/local/bin/gosu
+
+# Copy entrypoint script
+COPY scripts/entrypoint.sh /usr/local/bin/entrypoint.sh
```
sfc-gh-pfus: Why do we need this?

ci/test_rockylinux9.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash -e
+#
+# Test GoSnowflake driver in Rocky Linux 9
+# NOTES:
+#   - Go version to be tested should be passed in as the first argument, e.g: "1.24". If omitted 1.24 will be assumed.
```
sfc-gh-pfus: I don't think we should assume 1.24. It's going to be deprecated soon. Let's fail fast.

ci/test_rockylinux9.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "[Info] Using Go ${GO_VERSION}"
+
+if ! command -v go${GO_VERSION} &> /dev/null; then
+    echo "[ERROR] Go ${GO_VERSION} not found!"
+    exit 1
+fi
+
+# Make the specified Go version the default 'go' command for make test
+case "$GO_VERSION" in
+    "1.23") export GOROOT="/usr/local/go1.23.4" ;;
```
sfc-gh-pfus: Let's not do this. Name directories in a way that is easy to calculate. (in worst case - use symlinks)

ci/test_rockylinux9_docker.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -e
+# Test GoSnowflake driver in Rocky Linux 9 Docker
+# NOTES:
+#   - By default this script runs Go 1.24 tests
```
sfc-gh-pfus: Same.

.github/workflows/build-test.yml
```diff
@@ -232,3 +235,24 @@ jobs:
             WIREMOCK_HTTPS_PORT: 13567
             WIREMOCK_ENABLE_ECDSA: true
           run: ./ci/test.sh
+    build-test-rockylinux9:
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+            matrix:
+                cloud: [ 'AWS', 'AZURE', 'GCP' ]
+                go: [ '1.23', '1.24', '1.25' ]
```
sfc-gh-pfus: Do we need the full matrix? Maybe we can just use one go version to one cloud?

ci/docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+# Set Java 11 as the default using environment variables
+ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk
+ENV PATH="${JAVA_HOME}/bin:${PATH}"
+
+# Install Go versions 1.23, 1.24, and 1.25 (multi-arch aware)
+ENV GO_VERSIONS="1.23.4 1.24.2 1.25.0"
+
+# Download and install Go versions for the target architecture
+RUN GOARCH=${TARGETARCH:-amd64} && \
+    for version in $GO_VERSIONS; do \
```
sfc-gh-pfus: We do we need all, even if only one is used? It doesn't make sense and takes time.

.github/workflows/build-test.yml
sfc-gh-pcyrek: It's CodeQL, security advice, you can find full description above in comments 

Code scanning
/ CodeQL

Workflow does not contain permissions
Medium

Actions job or workflow does not limit the permissions of the GITHUB_TOKEN. Consider setting an explicit permissions block, using the following as a minimal starting point: {contents: read}

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: DNF is the package manager used in Red Hat Enterprise Linux based distributions

the rest is different libs, java for wiremock etc.

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: I think so. It's good to test all Go versions to support all Go versions. Testing once every half-year doesn't seem like a big problem. I can't see a better option to avoid it, yet keep the coverage matrix.

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: It's under DNF, but newest version is 1.24.6

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: If it's not a problem, I would like to stay with native package manager

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: True True True

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: yes

ci/test_rockylinux9_docker.sh
sfc-gh-pcyrek: ok

ci/test_rockylinux9.sh
sfc-gh-pcyrek: ok

.github/workflows/build-test.yml
sfc-gh-pcyrek: Parallelisation costs us nothing. This is a product decision. I follow an established pattern. IMHO, it's good to give them all.

sfc-gh-pfus: It costs us retries in case of flaky tests :) The more redundant builds we have, the higher chance it is that we will have to retry.

ci/docker/rockylinux9/Dockerfile
sfc-gh-pfus: Yeah, I forgot about wiremock.

ci/docker/rockylinux9/Dockerfile
sfc-gh-pfus: I meant python3 precisely :)

ci/docker/rockylinux9/Dockerfile
sfc-gh-pfus: I rather meant - why do we have to install it manually instead of using DNF package?

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: that's very good catch, implemented different logic here

.github/workflows/build-test.yml
sfc-gh-pcyrek: At the same time, the more retries and combinations, the higher the chances we catch some rare real bugs. ;)

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: It's for Wiremock stuff i believe

ci/docker/rockylinux9/Dockerfile
sfc-gh-pcyrek: I get rid of entrypoints.sh as test_rockylinux.sh serves  more or less the same needs

ci/test_rockylinux9.sh
sfc-gh-pcyrek: discussed offline

.github/workflows/build-test.yml
sfc-gh-pcyrek: discussed offline

PR #1598, opened 2025-10-27
Release memory earlier

obs-gh-felixbowing:
Problem:
The current approach of calling close on the
body is correct, but has two disadvantages:
- it only releases memory after finishing all iterations have finished
- The current construction is not trivial to understand and hence future refactorings might break this code again.

Solution:
Factor it out into a separate function that
does call close after the response is not used
anymore.

Use the same logic also for the async path.

No semantic changes.

### DIFF
--- a/restful.go
+++ b/restful.go
@@ -247,8 +247,8 @@ func postRestfulQueryHelper(
 	}(resp, fullURL.String())
 
 	if resp.StatusCode == http.StatusOK {
-		var respd execResponse
-		if err = json.NewDecoder(resp.Body).Decode(&respd); err != nil {
+		respd := &execResponse{}
+		if err = json.NewDecoder(resp.Body).Decode(respd); err != nil {
 			logger.WithContext(ctx).Errorf("failed to decode JSON. err: %v", err)
 			return nil, err
 		}
@@ -269,7 +269,7 @@ func postRestfulQueryHelper(
 
 		// if asynchronous query in progress, kick off retrieval but return object
 		if respd.Code == queryInProgressAsyncCode && isAsyncMode(ctx) {
-			return sr.processAsync(ctx, &respd, headers, timeout, cfg)
+			return sr.processAsync(ctx, respd, headers, timeout, cfg)
 		}
 		for isSessionRenewed || respd.Code == queryInProgressCode ||
 			respd.Code == queryInProgressAsyncCode {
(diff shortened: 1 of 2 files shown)


obs-gh-felixbowing:
I see that nearly all tests here fail. I ran `make test` locally and didn't get any unexpected failures. 

Could it be that the test suite here requires approval of owner?

restful.go
```diff
@@ -542,3 +530,33 @@ func getQueryIDChan(ctx context.Context) chan<- string {
 	}
 	return c
 }
+
+// getAndDecodeResponse fetches a response using FuncGet and decodes it into the argument
```
sfc-gh-pfus: Why do you prefer passing the response as an argument? Is it more natural to just return it from this function?

restful.go
```diff
@@ -542,3 +530,33 @@ func getQueryIDChan(ctx context.Context) chan<- string {
...
+// called respd. It overwrites any existing data in respd.
+func getAndDecodeResponse(
```
sfc-gh-pfus: Maybe just `getExecResponse`? Especially if respd is returned, not passed as an argument.

sfc-gh-pfus:
> I see that nearly all tests here fail. I ran `make test` locally and didn't get any unexpected failures.
> 
> Could it be that the test suite here requires approval of owner?

Yes, it requires approval. We don't want to start builds automatically, because tests connect to the real Snowflake. Take a look at my comments, and when we have final commit I'll start the tests!

restful.go
obs-gh-felixbowing: The problem is that respd is not a pointer-to-execResponse, but an execResponse in `postRestfulQueryHelper()`. Hence golang would create a copy of the execResponse when returning it to the caller `postRestfulQueryHelper()`. From what I can see execResponse is pretty large. Hence I decided to just fill it, which is similar to the decode function.

The other options are:
- changing respd to be a pointer type in `postRestfulQueryHelper()`. But I didn't want to touch this as I am not sure what other things might break.
- as suggested by you: Creating a new execResponse (causing mem allocations) and copying this into respd when returning the result. 

Does this make sense?

restful.go
obs-gh-felixbowing: `getExecResponse` is fine too I think.

restful.go
sfc-gh-pfus: I think we can safely change it to a pointer. `postRestfulQueryHelper` returns a pointer after all, so we can "pointerise" it one level deeper.

restful.go
obs-gh-felixbowing: Done

restful.go
obs-gh-felixbowing: Nice! I have adapted the PR. The variable respd is now a pointer type.

obs-gh-felixbowing:
@sfc-gh-pfus , I have incorporated your feedback. Could you have a look at the latest state of this PR?

sfc-gh-pfus:
Thanks @obs-gh-felixbowing ! I asked in our team for the second approval. In the meantime - can you rebase your change?

sfc-gh-rsavenok (approved):

obs-gh-felixbowing:
Thanks for the approvals. I have rebased my change.

sfc-gh-pfus:
Thanks again, @obs-gh-felixbowing ! Your change is merged!

PR #1600, opened 2025-10-28
NO-SNOW fix JIRA ticket creation workflow

sfc-gh-dszmolka:
### Description
This is a workflow only change, no code was touched. 

seems to be broken since https://github.com/snowflakedb/gosnowflake/pull/1501 and now jira creation request is refused with 'Bad Request'

Example: https://github.com/snowflakedb/gosnowflake/actions/runs/18857613863/job/53809145629

Thus jira creation workflow does not create jiras.

Using the JDBC repo's workflow here (with a different component), which works today.


.github/workflows/jira_issue.yml
```diff
@@ -13,62 +13,38 @@ jobs:
       issues: write
     if: ((github.event_name == 'issue_comment' && github.event.comment.body == 'recreate jira' && github.event.comment.user.login == 'sfc-gh-mkeller') || (github.event_name == 'issues' && github.event.pull_request.user.login != 'whitesource-for-github-com[bot]'))
     steps:
-      - name: Create JIRA Ticket
-        id: create
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: snowflakedb/gh-actions
```
sfc-gh-hpathak, line 19: we dont want this dependency. i had tested the PR that was merged earlier. not sure why it fails now. maybe jira api changed? in any case we cant have a public repo depend on the above private repo

sfc-gh-dszmolka: @sfc-gh-hpathak could you please suggest a fix then, which is acceptable for you, and also results in a working jira creation? We don't have it currently.

sfc-gh-dheyman: @sfc-gh-hpathak +1 to what @sfc-gh-dszmolka said. Given that this feature was working previously and not having it right now is causing a lot of pain, we would need a working replacement. We'd appreciate if you could revise your previous solution.

sfc-gh-dszmolka:
obsoleted by https://github.com/snowflakedb/gosnowflake/pull/1634

PR #1601, opened 2025-10-28
SNOW-2236558 platform detection

sfc-gh-rsavenok:
### Description

Implements platform detection telemetry. Based on python implementation https://github.com/snowflakedb/snowflake-connector-python/pull/2387/files

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Ensured proper permissions for CRL cache directory (snowflakedb/gosnowflake#1588)
 - Added `CrlDownloadMaxSize` to limit the size of CRL downloads (snowflakedb/gosnowflake#1588)
 - Added platform telemetry to login requests. Can be disabled with `SNOWFLAKE_DISABLE_PLATFORM_DETECTION` environment variable (snowflakedb/gosnowflake#1601)
+- Bypassed proxy settings for WIF metadata requests (snowflakedb/gosnowflake#1593)
 - Fixed a bug where GCP PUT/GET operations would fail when the connection context was cancelled (snowflakedb/gosnowflake#1584)
 -
 -


platform_detection.go
```diff
@@ -0,0 +1,336 @@
...
+
+func init() {
+	platformDetectionDone = make(chan struct{})
+	go func() {
+		initializePlatformDetection()
+	}()
+}
+
+func initializePlatformDetection() {
+	logger.Debugf("initializePlatformDetection: starting")
```
sfc-gh-rsavenok: This log doesn't work because the log level is set after this code is executed. 

Do you have ideas where this could be moved @sfc-gh-pfus? Maybe in buildSnowflakeConn?

CHANGELOG.md
```diff
@@ -9,18 +9,18 @@
...
 - Ensured proper permissions for CRL cache directory (snowflakedb/gosnowflake#1588)
 - Added `CrlDownloadMaxSize` to limit the size of CRL downloads (snowflakedb/gosnowflake#1588)
-- 
+- Added platform telemetry to login requests. Can be disabled with `SNOWFLAKE_DISABLE_PLATFORM_DETECTION` environment variable. (snowflakedb/gosnowflake#1601)
 - Bypassed proxy settings for WIF metadata requests (snowflakedb/gosnowflake#1593)
-- 
-- 
-- 
+-
+-
```
sfc-gh-rsavenok, line 16: My editor has `trimTrailingWhitespace: true` on saves :)

platform_detection.go
```diff
@@ -0,0 +1,308 @@
...
+		{name: "has_gcp_identity", fn: detectGcpIdentity},
+	}
+
+	detectionStates := make(map[string]platformDetectionState, len(detectors))
+	var waitGroup sync.WaitGroup
+	var mutex sync.Mutex
+	waitGroup.Add(len(detectors))
+
+	for _, detector := range detectors {
+		detector := detector // capture loop variable
```
sfc-gh-pfus: What does this mean? Is it just to have correct detector in goroutines, instead of the last one? If yes, you can pass it to `go func(detector)` instead.

platform_detection.go
```diff
@@ -0,0 +1,308 @@
...
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return platformDetectionTimeout
+		}
+		return platformNotDetected
+	}
+	if out == nil || out.Arn == nil || *out.Arn == "" {
+		return platformNotDetected
+	}
+	if isValidArnForWif(*out.Arn) {
```
sfc-gh-pfus, line 199: Why only for WIF? This platform detection is something more than WIF, right?

sfc-gh-pfus (approved):

platform_detection.go
sfc-gh-rsavenok: Maybe that's just a bad name for `has_aws_identity`. Every detector we made is to track a platform that can run WIF (not every aws identity works for wif), so the DBSEC can compare WIF adoption and overall users on those platforms.

platform_detection.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode == http.StatusOK {
+		return platformDetected
+	}
+	return platformNotDetected
+}
+
+func isValidArnForWif(arn string) bool {
+	patterns := []string{
```
sfc-gh-dheyman, line 296: any tests for these regexps?

sfc-gh-rsavenok: Added tests

PR #1616, opened 2025-11-04
SNOW-1997074 Handle HTTP307 & 308 in drivers to achieve better resiliency to backend errors

sfc-gh-ext-simba-jy:
### Description

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,6 @@ New features:
 Bug fixes:
 
 - Handle HTTP307 & 308 in drivers to achieve better resiliency to backend errors (snowflakedb/gosnowflake#1616).
-- Fixed a data race error in tests caused by platform_detection init() function (snowflakedb/gosnowflake#1618)
-- Make secrets detector initialization thread safe and more maintainable (snowflakedb/gosnowflake#1621)
 -
 -
 -


restful.go
```diff
@@ -248,7 +248,12 @@ func postRestfulQueryHelper(
 
 	if resp.StatusCode == http.StatusOK {
 		respd := &execResponse{}
-		if err = json.NewDecoder(resp.Body).Decode(respd); err != nil {
+		if err = json.NewDecoder(resp.Body).Decode(&respd); err != nil {
+			if redirectResponse := resp.Request.Response; redirectResponse != nil {
```
sfc-gh-pfus: Why do we need to do it in this layer? We have `retry` layer and this redirect response is present without decoding body, right?

If not, we should probably add the same to other functions as well.

retry_test.go
```diff
@@ -673,3 +677,57 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
 		})
 	}
 }
+
+func TestRetry307(t *testing.T) {
+	wiremockHTTPS.registerMappings(t,
```
sfc-gh-pfus: Why do you use HTTPS and not HTTP?

retry_test.go
```diff
@@ -673,3 +677,57 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
...
+		wiremockMapping{filePath: "retry/http_307_retry.json"},
```
sfc-gh-pfus: We have `newWiremockMapping` function that can be helpful.

retry_test.go
```diff
@@ -673,3 +677,57 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
...
+	assertNilF(t, err)
+	testCertPool.AddCert(certificate)
+	cfg.Transporter = &http.Transport{
+		TLSClientConfig: &tls.Config{
+			RootCAs: testCertPool,
+		},
+	}
+	connector := NewConnector(SnowflakeDriver{}, *cfg)
+	db := sql.OpenDB(connector)
+	rows, err := db.Query("SELECT 1")
```
sfc-gh-pfus: We have `runSmokeQuery` function that can be helpful.

retry_test.go
```diff
@@ -673,3 +677,57 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
...
+	defer rows.Close()
+	var v int
+	assertTrueF(t, rows.Next())
+	assertNilF(t, rows.Scan(&v))
+	assertEqualE(t, v, 1)
+}
+
+func TestRetry308(t *testing.T) {
+	wiremockHTTPS.registerMappings(t,
+		wiremockMapping{filePath: "retry/http_308_retry.json"},
```
sfc-gh-pfus: It's basically the same, can we combine these tests using `t.Run`?

test_data/wiremock/mappings/retry/http_308_retry.json
```diff
@@ -0,0 +1,436 @@
+{
+  "mappings": [
+    {
+      "scenarioName": "HTTP 308 retry",
```
sfc-gh-pfus: It looks the same - can we parameterize just one file?

retry_test.go
sfc-gh-ext-simba-jy: Changed to HTTP.

retry_test.go
sfc-gh-ext-simba-jy: Applied

retry_test.go
sfc-gh-ext-simba-jy: Done.

retry_test.go
sfc-gh-ext-simba-jy: I refactored them.

test_data/wiremock/mappings/retry/http_308_retry.json
sfc-gh-ext-simba-jy: Done.

restful.go
sfc-gh-ext-simba-jy: I tried this at the retry layer, but I faced an issue the first time. I found what the root cause was and moved this into the retry layer.

retry.go
```diff
@@ -401,6 +402,13 @@ func isRetryableError(req *http.Request, res *http.Response, err error) (bool, e
 	if res == nil || req == nil {
 		return false, err
 	}
+
+	// handle redirect
+	if res.Request != nil && res.Request.Response != nil {
+		if parsedErr := json.NewDecoder(res.Body).Decode(&execResponse{}); parsedErr != nil {
```
sfc-gh-pfus: Why do we need to parse the response here? Shouldn't we focus here on `res.StatusCode` being 307/308? It's already handled by `isRetryableStatus`, right?

test_data/wiremock/mappings/retry/http_307_retry.json
```diff
@@ -0,0 +1,19 @@
+{
```
sfc-gh-pfus: This is not what I wanted. I wanted one file for the whole scenario, which use `params` of `wiremockMapping` to substitute status code.

retry.go
sfc-gh-ext-simba-jy: Hello @sfc-gh-pfus . No, it’s not actually handled by isRetryableStatus.
The redirection HTTP status codes (3xx) are handled internally by client.go in net/http.
The main purpose of this test is to verify how the client handles errors during redirection.

As a result, even if a redirection fails (for example, due to a timeout), the client may still return an HTTP status code 200 with an empty body or an incorrect response, which causes the execution to fail.
In short, we cannot rely on the status code alone to handle redirection errors on our client side.

Here’s what this code does:

In the first if statement, it checks whether the previous request was a redirection. According to the net/http documentation, the Request.Response field will not be nil if the previous request was redirected.

Even if a redirection occurs, it can still return a valid response. Therefore, the driver checks whether the response body exists, and if it’s empty, it retries the request.

test_data/wiremock/mappings/retry/http_307_retry.json
sfc-gh-ext-simba-jy: Done.

retry.go
sfc-gh-pfus: Hmmm, I don't get it. Why in case of timeout we get 200? I would expect to not receive response at all in that case. I think you're test is incorrect - your mock waits 65s and just returns 200 with no body. But this is not a timeout case.

What happens if we read the body, but there is no error? Can we have such a test case? I'm afraid we can't read it again in the upper layer. Also, retry layer handles various types of response bodies, not only execResponse.

retry_test.go
```diff
@@ -673,3 +674,20 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
...
+
+func TestRedirectRetry(t *testing.T) {
+	httpCodes := []string{
+		"307",
+		"308",
+	}
+
+	for _, httpCode := range httpCodes {
+		t.Run("retry with http code "+httpCode, func(t *testing.T) {
+			wiremock.registerMappings(t, newWiremockMappingWithParam("retry/redirection_retry_workflow.json", map[string]string{"HTTP_STATUS_CODE": httpCode}))
```
sfc-gh-pfus: Can we use `%key%` marker as in other tests?

test_data/wiremock/mappings/retry/redirection_retry_workflow.json
```diff
@@ -0,0 +1,309 @@
...
+        {
+            "scenarioName": "wiremock retry strategy",
+            "requiredScenarioState": "Query attempt with HTTP 3xx response",
+            "newScenarioState": "3xx redirect followed and times out",
+            "request": {
+                "urlPathPattern": "/temp-redirect-1",
+                "method": "POST"
+            },
+            "response": {
+                "fixedDelayMilliseconds": 65000
```
sfc-gh-pfus: Definitely wrong. Reduce it to something reasonable and decrease the timeout in driver. Our tests are already to long.

test_data/wiremock/mappings/retry/redirection_retry_workflow.json
```diff
@@ -0,0 +1,309 @@
...
+            }
+        },
+        {
+            "scenarioName": "wiremock retry strategy",
+            "requiredScenarioState": "3xx redirect followed and times out",
```
sfc-gh-pfus, line 92: I'm wondering if this is the root scenario that we want to check?
1. Query request ends with redirect.
2. Redirect leads to some unavailable address and causes timeout.
3. Retry to query request ends with no redirect this time?

That seems unlikely, I would expect that the second query request ends with the same redirect?

test_data/wiremock/mappings/retry/redirection_retry_workflow.json
sfc-gh-ext-simba-jy: This mocking file is made in the JDBC PR. Not made by my decision.

sfc-gh-ext-simba-jy: https://github.com/snowflakedb/snowflake-jdbc/pull/2344

sfc-gh-pfus: It doesn't mean it's correct :)

retry.go
sfc-gh-pfus: Ok, I dug a bit deeper and this is my thinking:
1. Default Golang timeout for a single request is 15 min 🤦 This is why you got 200 and no body - it was just an empty response returned after 65s, not timeout. I think for the sake of this task, we can leave the default timeout as is, but in your test, if you want to simulate a timeout, just use a lower value for `cfg.ClientTimeout` like 500ms or something. Then decrease this timeout in the scenario file as well.
2. I think that we incorrectly handle retries. Currently they are not retried ([problematic PR](https://github.com/snowflakedb/gosnowflake/pull/1540/files)). I think that what we should do is to distinguish two cases:
a) Context deadlined due to http client timeout
b) Context deadlined due to context deadline
I think we should check and retry only for case a). Now we retry never in case of timeout.

test_data/wiremock/mappings/retry/redirection_retry_workflow.json
sfc-gh-ext-simba-jy: Okay, no problem — I agree with you. I just wanted to confirm that we can modify the testing values. I initially thought we needed to unify all testing scenarios. I’ve already shared your questions with the team, and we’ll make sure the values are reasonable and include a timeout scenario as well. Thanks for bringing up these points!

retry_test.go
sfc-gh-ext-simba-jy: Hello @sfc-gh-pfus, I added the reply to your question https://github.com/snowflakedb/gosnowflake/pull/1616#discussion_r2501916156. We do not have the testing case currently and @sfc-gh-ext-simba-hx also raised a similar question (about the retry timeout) yesterday. I will add the testing scenario you mentioned in the PR and investigate the issue.

test_data/wiremock/mappings/retry/redirection_retry_workflow.json
sfc-gh-ext-simba-jy: According to the original mapping, it seems the second query request (retry) is not redirected like the first query attempt.

retry.go
```diff
@@ -393,14 +394,25 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
 
 func isRetryableError(req *http.Request, res *http.Response, err error) (bool, error) {
 	if err != nil && res == nil { // Failed http connection. Most probably client timeout.
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		if errors.Is(err, context.Canceled) {
 			return false, err
 		}
+
+		// the timeout happened, retry it first
```
sfc-gh-pfus: Is it a correct assumption? I would expect, that when timeout happens, `context.DeadlineExceeded` is returned. `canceled` is returned when the context is cancelled by a caller, manually.

Moreover, as I mentioned in the comment below - we have to distinguish `context.DeadlineExceeded` for two scenarios:
1. HTTP timeout happened - we should retry
2. Client context (passed to `sql.QueryContext`) deadlined - we shouldn't retry.

In both cases we probably receive the same `context.DeadlineExceeded` from `errors.Is` perspective. I have two ideas:
1. Check `ctx.Err()` if it is already deadlined.
2. Check if `err` is `net.Error` and verify `Timeout()` function

I prefer the first, as it is our source of ultimate truth - whether the context is definitely not valid anymore. But you have to test it.

retry.go
sfc-gh-pfus: I still don't think we can read the body here. It would be useless in the upper layer (we can't read the body twice).

retry_test.go
```diff
@@ -673,3 +674,29 @@ func TestCalculateRetryWaitForNonAuthRequests(t *testing.T) {
...
+	for _, httpCode := range httpCodes {
+		t.Run("retry with the redirection http code: "+httpCode, func(t *testing.T) {
+			wiremock.registerMappings(t, newWiremockMappingWithParam("retry/redirection_retry_workflow.json", map[string]string{"%HTTP_STATUS_CODE%": httpCode}))
+			cfg := wiremock.connectionConfig()
+			connector := NewConnector(SnowflakeDriver{}, *cfg)
+			db := sql.OpenDB(connector)
+			runSmokeQuery(t, db)
+		})
+
+		t.Run("retry when redirection exceeds client Timeout: "+httpCode, func(t *testing.T) {
```
sfc-gh-pfus: I don't think we should run this for each HTTP code. It adds unnecessary slowness to our builds. It's not even related to 307/308 IIUC.

sfc-gh-ext-simba-jy: Sure. I removed the loop and only tested one redirection HTTP code (307).

retry.go
sfc-gh-ext-simba-jy: I added 1 scenario and testing.

retry.go
sfc-gh-ext-simba-jy: Right. I added a validation function. Please review it and let me know if this works.

retry.go
```diff
@@ -391,16 +392,26 @@ func (r *retryHTTP) execute() (res *http.Response, err error) {
...
 
-func isRetryableError(req *http.Request, res *http.Response, err error) (bool, error) {
+func isRetryableError(ctx context.Context, req *http.Request, res *http.Response, err error) (bool, error) {
+	if ctx.Err() != nil {
+		return false, ctx.Err()
+	}
+
 	if err != nil && res == nil { // Failed http connection. Most probably client timeout.
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		if errors.Is(err, context.Canceled) {
```
sfc-gh-pfus: I think this is already covered by the first if here, right? Can you add a test case (if missing) and check without this if?

retry.go
```diff
@@ -411,3 +422,18 @@ func isRetryableStatus(statusCode int) bool {
 func isLoginRequest(req *http.Request) bool {
 	return contains(authEndpoints, req.URL.Path)
 }
+
+func isRetryableRedirectionResponse(res *http.Response) (bool, error) {
+	data, err := io.ReadAll(res.Body)
```
sfc-gh-pfus: Again, I'm repeating myself, but you shouldn't read body here. I think that handling context timeouts/deadlines already covers the timeout situation. What's the point of doing this parsing check?

retry.go
```diff
@@ -411,3 +422,18 @@ func isRetryableStatus(statusCode int) bool {
...
+	if err != nil {
+		return true, err
+	}
+	if len(data) == 0 {
+		return true, err
+	}
+	if parsedErr := json.NewDecoder(bytes.NewBuffer(data)).Decode(&execResponse{}); parsedErr != nil {
+		return true, parsedErr
+	}
+	res.Body = io.NopCloser(bytes.NewBuffer(data))
```
sfc-gh-pfus: This is some workaround, but increases the memory needed to buffer the whole response.

retry_test.go
```diff
@@ -486,70 +487,107 @@ func TestLoginRetry429(t *testing.T) {
...
 			req:      &http.Request{URL: &url.URL{Path: loginRequestPath}},
 			res:      nil,
 			err:      &url.Error{Err: context.Canceled},
 			expected: false,
 		},
 		{
+			ctx:      context.Background(),
+			req:      &http.Request{URL: &url.URL{Path: queryRequestPath}},
+			res:      nil,
+			err:      &url.Error{Err: context.Canceled},
```
sfc-gh-pfus: I think it would be good to have a e2e scenario too:
1. create a cancellable context
2. start a request to wiremock that hangs for 1s in a separate goroutine
3. cancel the context after 100ms
And then check if we can remove handling this dedicated for cancellation if block.

retry_test.go
```diff
@@ -486,70 +487,107 @@ func TestLoginRetry429(t *testing.T) {
...
 			res:      &http.Response{StatusCode: http.StatusTooManyRequests},
 			err:      nil,
 			expected: true,
 		},
+
+		{
+			ctx:      deadLineCtx,
+			req:      &http.Request{URL: &url.URL{Path: loginRequestPath}},
+			res:      nil,
+			err:      &url.Error{Err: context.Canceled},
```
sfc-gh-pfus: It's not cancelled, it's context.DeadlineExceeded if your context is deadlined.

retry.go
```diff
@@ -411,3 +422,18 @@ func isRetryableStatus(statusCode int) bool {
 func isLoginRequest(req *http.Request) bool {
 	return contains(authEndpoints, req.URL.Path)
 }
+
+func isRetryableRedirectionResponse(res *http.Response) (bool, error) {
```
sfc-gh-pfus: This function works, but only for one redirect, right? Maybe we should make it recursive to follow all redirects?

sfc-gh-ext-simba-jy: We do not need the this function anymore. I removed it.

retry_test.go
sfc-gh-ext-simba-jy: Removed.

retry_test.go
sfc-gh-ext-simba-jy: It seems this test covers your scenario already, in my opinion: https://github.com/snowflakedb/gosnowflake/blob/a686da1d5c0dd66984bc93834135f613dfd7c396/restful_test.go#L663

retry.go
sfc-gh-ext-simba-jy: No more need to check.

retry.go
sfc-gh-ext-simba-jy: I double-checked it, and this one is not necessary anymore.

retry.go
sfc-gh-ext-simba-jy: Right, I misunderstood the main point of this PR.
This WireMock test is actually meant to verify the behavior when the client timeout is shorter than the server’s redirection wait time. I initially thought I needed to handle the case where the client timeout is longer than the server timeout. That is why I tried to validate the body information, but it turns out that this isn’t necessary. In short, we do not need to validate the response anymore in the retry level.

retry.go
sfc-gh-pfus: So can you remove it? ;)

sfc-gh-pfus (approved):

PR #1617, opened 2025-11-05
SNOW-2395141 Fix CRL download checking

sfc-gh-pfus:
### Description

### Checklist

### DIFF
--- a/crl.go
+++ b/crl.go
@@ -295,6 +295,13 @@ func (cv *crlValidator) validateCrlAgainstCrlURL(cert *x509.Certificate, crlURL 
 		if err != nil {
 			logger.Warnf("failed to download CRL from %v: %v", crlURL, err)
 		}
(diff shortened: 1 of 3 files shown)


crl.go
```diff
@@ -295,6 +295,10 @@ func (cv *crlValidator) validateCrlAgainstCrlURL(cert *x509.Certificate, crlURL
 		if err != nil {
 			logger.Warnf("failed to download CRL from %v: %v", crlURL, err)
 		}
+		if newCrl != nil && newCrl.NextUpdate.Before(now) {
```
sfc-gh-snoonan, line 298: Because of downloadTime.Add(crlCacheCleaner.cacheValidityTime).Before(now) above, you might be returning a certError when we still technically have a valid CRL.

sfc-gh-pfus: You're right!

Issue #1615, opened 2025-11-04
SNOW-2677427: Data race hit in v1.17.1


crobert-1:
1. What version of GO driver are you using?
`v1.17.1`
   
2. What operating system and processor architecture are you using?
OS: Ubuntu 24.04
CPU: amd64
   
3. What version of GO are you using?
`go 1.24.9`

5. What did you do?

The OpenTelemetry Collector has a dependency on the `gosnowflake` driver, and [renovate opened a PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44032) today updating the dependency from `v1.17.0` to `v1.17.1`.

Our CI tests are not calling any functionality of the `gosnowflake` driver, it's simply importing the module.

6. What happened?

[Tests are failing](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/19075128969/job/54488781254?pr=44032#step:4:115) at the highest level, not inside of actual tests, due to a data race in `init` functionality of the `gosnowflake` driver.

Here's the test failure. I'm not very familiar with the innerworkings of `gosnowflake` and `aws-sdk-go-v2`, but the AWS SDK Go dependency was not updated as a part of this PR and test failure, so I don't believe it's the cause of the data race.

```
=== Failed
=== FAIL: .  (0.00s)
==================
WARNING: DATA RACE
Read at 0x0000062e23b0 by goroutine 19:
  github.com/aws/aws-sdk-go-v2/config.resolveDefaultAWSConfig()
      /home/runner/go/pkg/mod/github.com/aws/aws-sdk-go-v2/config@v1.27.11/resolve.go:30 +0x186
  github.com/aws/aws-sdk-go-v2/config.configs.ResolveAWSConfig()
      /home/runner/go/pkg/mod/github.com/aws/aws-sdk-go-v2/config@v1.27.11/config.go:147 +0x165
  github.com/aws/aws-sdk-go-v2/config.LoadDefaultConfig()
      /home/runner/go/pkg/mod/github.com/aws/aws-sdk-go-v2/config@v1.27.11/config.go:199 +0x2e4
  github.com/snowflakedb/gosnowflake.detectAwsIdentity()
      /home/runner/go/pkg/mod/github.com/snowflakedb/gosnowflake@v1.17.1/platform_detection.go:181 +0xee
  github.com/snowflakedb/gosnowflake.detectPlatforms.func1()
[...]

PR #1618, opened 2025-11-06
SNOW-2677427 Change where platfrom detection inits

sfc-gh-rsavenok:
### Description

Fixes https://github.com/snowflakedb/gosnowflake/issues/1615 + hides AWS SDK log message when metadata hit fails

The root cause is the Go testing framework that is reassigning `os.Stderr` while AWS SDK `resolveDefaultAWSConfig` tries to read it.

The solution is to move platform detection from `init()` to the connection creation function `buildSnowflakeConn` + ensuring that the function executes only once.

### Checklist

### DIFF
--- a/platform_detection_test.go
+++ b/platform_detection_test.go
@@ -238,7 +238,7 @@ func TestDetectPlatformsTimeout(t *testing.T) {
 	platforms := detectPlatforms(context.Background(), 200*time.Millisecond)
 	executionTime := time.Since(start)
 
-	assertTrueE(t, len(platforms) == 0, fmt.Sprintf("Expected empty platforms, got: %v", platforms))
+	assertEqualE(t, len(platforms), 0, fmt.Sprintf("Expected empty platforms, got: %v", platforms))
 	assertTrueE(t, executionTime >= 200*time.Millisecond && executionTime < 250*time.Millisecond,
 		fmt.Sprintf("Expected execution time around 200ms, got: %v", executionTime))
 }


connection.go
```diff
@@ -868,6 +868,7 @@ func buildSnowflakeConn(ctx context.Context, config Config) (*snowflakeConn, err
 		queryContextCache:   (&queryContextCache{}).init(),
 		currentTimeProvider: defaultTimeProvider,
 	}
+	initPlatformDetection()
```
sfc-gh-dheyman, line 871: So the same solution as in JDBC 👍

platform_detection_test.go
```diff
@@ -71,10 +72,31 @@ func setupWiremockMetadataEndpoints() func() {
...
+	initPlatformDetection()
+	platforms1 := getDetectedPlatforms()
+
+	// Verify caching works and AWS Lambda detected
+	assertTrueF(t, slices.Equal(platforms1, detectedPlatformsCache), "getDetectedPlatforms should return cached result")
+	assertTrueF(t, slices.Contains(platforms1, "is_aws_lambda"), "Should detect AWS Lambda")
+
+	// Change environment and test sync.Once behavior
+	cleanup()
+	os.Setenv("GITHUB_ACTIONS", "true")
```
sfc-gh-pfus, line 93: Why do we need this here? We should roll back this value later (see `envOverride`). Unless `cleanup` handles this.

platform_detection.go
```diff
@@ -33,16 +33,18 @@ var (
 )
 
 var (
-	detectedPlatformsCache []string
-	platformDetectionDone  chan struct{}
+	detectedPlatformsCache    []string
+	initPlatformDetectionOnce sync.Once
```
sfc-gh-dheyman, line 39: does this guarantee thread safety? Will the other threads wait or will they also proceed but the first result will be taken into account?

platform_detection_test.go
```diff
@@ -71,10 +72,31 @@ func setupWiremockMetadataEndpoints() func() {
...
+	originalDone, originalCache := platformDetectionDone, detectedPlatformsCache
+	initPlatformDetectionOnce, platformDetectionDone, detectedPlatformsCache = sync.Once{}, make(chan struct{}), nil
+	defer func() { platformDetectionDone, detectedPlatformsCache = originalDone, originalCache }()
+
+	os.Setenv("AWS_LAMBDA_TASK_ROOT", "/var/task")
+	initPlatformDetection()
+	platforms1 := getDetectedPlatforms()
+
+	// Verify caching works and AWS Lambda detected
+	assertTrueF(t, slices.Equal(platforms1, detectedPlatformsCache), "getDetectedPlatforms should return cached result")
```
sfc-gh-pfus: I think we could introduce new `assertSlicesEqualF/E` function. In this case it could be `E`, right?

platform_detection.go
sfc-gh-rsavenok: In Go (Golang), sync.Once is a synchronization primitive from the sync package that ensures a piece of code runs only once, no matter how many times it’s called — even across multiple goroutines.

platform_detection_test.go
sfc-gh-rsavenok: Cleaup will restore it, it's just a variable set to sumulate`is_github_action` in the detector.

The test does:
* simulate aws in first run
* explicitly sumulates github env in second run
* ensures that both runs actually return result of the first run

platform_detection_test.go
sfc-gh-rsavenok: We didn't need it for the past 7 years, so do you think it's really needed? 

Pls confirm if you think that we should and I'll add it :)

connection.go
sfc-gh-pfus: Sad but true :( And only because of how golang handles JSON test output :(

platform_detection.go
sfc-gh-pfus: Yes.

platform_detection_test.go
sfc-gh-pfus: It was added only in 1.21, so we could use it since not long ago :) I'd add it. Same for slices.Contains.

sfc-gh-rsavenok: Turns out we already have a helper for this - assertDeepEqualE :)

platform_detection_test.go
```diff
@@ -217,8 +238,8 @@ func TestDetectPlatformsTimeout(t *testing.T) {
 	platforms := detectPlatforms(context.Background(), 200*time.Millisecond)
 	executionTime := time.Since(start)
 
-	assertTrueF(t, len(platforms) == 0, fmt.Sprintf("Expected empty platforms, got: %v", platforms))
-	assertTrueF(t, executionTime >= 200*time.Millisecond && executionTime < 250*time.Millisecond,
+	assertTrueE(t, len(platforms) == 0, fmt.Sprintf("Expected empty platforms, got: %v", platforms))
```
sfc-gh-pfus: `assertEqualE`?

PR #1625, opened 2025-11-17
SNOW-2814644 Add logging of query text

sfc-gh-pfus:
### Description

SNOW-2814644 Add logging of query text

### Checklist

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -96,6 +96,16 @@ func (sc *snowflakeConn) exec(
 	describeOnly bool,
 	bindings []driver.NamedValue) (
 	*execResponse, error) {
+	if sc.cfg.LogQueryText || isLogQueryTextEnabled(ctx) {
+		if len(bindings) > 0 && (sc.cfg.LogQueryParameters || isLogQueryParametersEnabled(ctx)) {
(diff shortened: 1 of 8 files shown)


sfc-gh-rsavenok (approved):

connection.go
```diff
@@ -96,6 +96,16 @@ func (sc *snowflakeConn) exec(
...
 	bindings []driver.NamedValue) (
 	*execResponse, error) {
+	if sc.cfg.LogQueryText || isLogQueryTextEnabled(ctx) {
+		if len(bindings) > 0 && (sc.cfg.LogQueryParameters || isLogQueryParametersEnabled(ctx)) {
+			logger.WithContext(ctx).Infof("Executing query: %v with bindings: %v", query, bindings)
+		} else {
+			logger.WithContext(ctx).Infof("Executing query: %v", query)
+		}
+	} else {
+		logger.WithContext(ctx).Infof("Executing query")
```
sfc-gh-snoonan, line 106: Just a thought - it would be nice to log a hash of the query or an ID even if we don't log the query in this case, as we'd be able to tell there are different queries being executed.

sfc-gh-pfus: We have a session ID logged (thanks to `WithContext(ctx)`), is it good enough?

PR #1629, opened 2025-11-20
SNOW-2465270 Integrate minicore

sfc-gh-pfus:
### Description

SNOW-2465270 Integrated minicore into go driver.

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -5,11 +5,8 @@ labels: bug
 ---
 
 
-<!--
-If you need urgent assistance then file the issue using the support process: 
-https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge 
-otherwise continue here. 
--->
+:exclamation: If you need **urgent assistance** then [file a case with Snowflake Support](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge).
+Otherwise continue here.
 
 
 Please answer these questions before submitting your issue. 
--- a/.github/workflows/jira_issue.yml
+++ b/.github/workflows/jira_issue.yml
@@ -19,26 +19,37 @@ jobs:
           JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
           JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+          ISSUE_BODY: ${{ github.event.issue.body }}
+          ISSUE_URL: ${{ github.event.issue.html_url }}
         run: |
+          # debug
+          #set -x
+          TMP_BODY=$(mktemp)
+          trap "rm -f $TMP_BODY" EXIT
+
           # Escape special characters in title and body
-          TITLE=$(echo '${{ github.event.issue.title }}' | sed 's/"/\\"/g' | sed "s/'/\\\'/g")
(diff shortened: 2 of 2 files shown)


auth.go
```diff
@@ -451,11 +455,26 @@ func authenticate(
...
-		Os:        runtime.GOOS,
-		OsVersion: osVersion,
-		Isa:       runtime.GOARCH,
-		GoVersion: runtime.Version(),
+		Os:            runtime.GOOS,
+		OsVersion:     osVersion,
+		Isa:           runtime.GOARCH,
+		GoVersion:     runtime.Version(),
+		CoreVersion:   coreVersion,
+		CoreLoadError: cmp.Or(coreLoadErr, errors.New("")).Error(),
```
sfc-gh-mhofman: we need both to go through secret detector

sfc-gh-pfus: Good idea!

.github/workflows/build-test.yml
```diff
@@ -240,19 +240,113 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                cloud: [ 'AWS', 'AZURE', 'GCP' ]
```
sfc-gh-jszczerbinski, line 243: Why remove the cartesian product here?

sfc-gh-jszczerbinski: If you want to have less builds for a PR that's fine, but let's run them all on main

libsf_mini_core_linux_arm64_glibc.so
sfc-gh-jszczerbinski, line 1: Nit: since we have musl suffix, should we have gnu or glibc suffix here?

minicore_provider.go
```diff
@@ -0,0 +1,152 @@
...
+	_ "embed"
+
+	"bufio"
+	"os"
+	"runtime"
+	"strings"
+)
+
+//go:embed libsf_mini_core_aix_ppc64.a
+var coreLibAixPpc64 []byte
```
sfc-gh-jszczerbinski: Do we embed every library bytes into every build? Can we embed limit the amount using compile time checks? Aren't you worried about binary size?

ci/test.bat
```diff
@@ -61,7 +61,7 @@ echo [INFO] Role:      %SNOWFLAKE_TEST_ROLE%
 go install github.com/jstemmer/go-junit-report/v2@latest
 
 REM Run tests and save output to file
-go test %GO_TEST_PARAMS% --timeout 90m --tags=sfdebug -race -coverprofile=coverage.txt -covermode=atomic -v . > test-output.txt 2>&1
+go test %GO_TEST_PARAMS% --timeout 90m --tags=sfdebug -coverprofile=coverage.txt -covermode=atomic -v . > test-output.txt 2>&1
```
sfc-gh-rsavenok, line 65: Why -race is disabled for Windows?

minicore_provider.go
sfc-gh-rsavenok: +1 to this question: the prev design that compiled only 1 binary looked better and from what I read, it can also handle musl vs glib

.github/workflows/build-test.yml
sfc-gh-pfus: I don't think we need this. We have never had an issue on a particular cloud AND a particular Go version. We use the same approach in JDBC and it's fine.

libsf_mini_core_linux_arm64_glibc.so
sfc-gh-pfus: Added.

minicore_provider.go
sfc-gh-pfus: Yeah, your right.

ci/test.bat
sfc-gh-pfus: Because it doesn't work on ARM windows. But it's fine, one build testing for race is enough. And we have all mac/linux tests for this.

sfc-gh-rsavenok: Perhaps add a comment to this file explaining why the -race option is skipped?

sfc-gh-pfus: Added.

PR #1638, opened 2025-12-03
Reduce spammy "telemetry disabled" logs

obs-gh-felixbowing:
Problem:
As we return an error in
these two places, the caller
emits a warning level log message
each time this happens. This
is very verbose.

Solution:
Don't return an error. Turning
off telemetry is a publicly accessible
feature so it should not
trigger a warning. Instead
we emit a debug level
log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ Bug fixes:
 -
 
 Internal changes:
-- 
+- Remove spammy "telemetry disabled" log messages (snowflakedb/gosnowflake#1638).
 -
 -
 -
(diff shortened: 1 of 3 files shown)


sfc-gh-pfus:
@obs-gh-felixbowing there is one test failing, can you take a look? Also, can you add a line to our changelog.md file (internal changes section)?

obs-gh-felixbowing:
Nice! Yes I have updated the PR with the fixed test + a changelog entry.

sfc-gh-pfus:
Thanks @obs-gh-felixbowing for the PR, it's merged!

PR #1646, opened 2025-12-05
SNOW-2871406 Expand user home dir only when needed

sfc-gh-pfus:
### Description

SNOW-2871406 In environments where we can't look up the current user (we encountered it in PE), we shouldn't depend on it. In such cases, there's no point in expanding `~` to user home.

### Checklist

### DIFF
--- a/file_util.go
+++ b/file_util.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/url"
 	"os"
-	usr "os/user"
 	"path/filepath"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


file_util.go
```diff
@@ -215,6 +215,9 @@ func baseName(path string) string {
 
 // expandUser returns the argument with an initial component of ~
 func expandUser(path string) (string, error) {
+	if !strings.HasPrefix(path, "~") {
+		return path, nil
+	}
 	usr, err := usr.Current()
```
sfc-gh-snoonan: Can't we just use https://pkg.go.dev/os#UserHomeDir ? This is cross platform and handles environmental variables properly.

sfc-gh-pfus: We can, but it doesn't matter. The problem will be the same - if the user is unknown, we still won't be able to get HOME dir. But changed either way.

sfc-gh-snoonan (approved):

Issue #1658, opened 2025-12-12
SNOW-2897060: Compress test_data/largefile.txt


maciej:
The project uses `test_data/largefile.txt` for some of its tests. This file in turns gets included in the published go module and distributed to go module caches everywhere where `gosnowflake`  is used.
Given the module's widespread use there are probably terabytes of disk space wasted across all developer machines around the world.
Thankfully, the file compresses to a mere 15kiB with zstandard and decompresses in milliseconds, not impeding test runs significantly[1].

I suggest having it compressed in the repo and removing the 70-odd megs.

Wasted disk space:
```
⋊> ~/g/p/m/g/snowflakedb pwd
/Users/maciej/go/pkg/mod/github.com/snowflakedb
⋊> ~/g/p/m/g/snowflakedb find . -path '*/test_data/largefile.txt' -type f -print0 | xargs -0 stat -f '%z' | awk '{s+=$1} END {printf "%.2f MiB\n", s/1024/1024}'
334.97 MiB
```

Decompression time (with vanilla zstandard, Go implementation should not be significantly slower)
```
time zstd -d largefile.txt.zstd -o largefile.txt.2
largefile.txt.zstd  : 70248250 bytes

________________________________________________________
Executed in   45.33 millis    fish           external
   usr time   18.83 millis    0.43 millis   18.40 millis
   sys time   19.72 millis    2.06 millis   17.67 millis
```

sfc-gh-dszmolka:
hey @maciej fair observation, thank you for drawing our attention to it and also for the fix suggestion. The team will take a look how to best move forward with this.

sfc-gh-pfus:
To be honest, I think we could even remove this file and generate random content on demand.

PR #1659, opened 2025-12-12
Compress test_data/largefile.txt fixture

maciej:
An attempt at reducing disk space usage of the published module. See #1658 for issue description.

The code was written by ChatGPT 5.1 Codex, using OpenCode with the following prompt:
```
The project is currently using test_data/largefile.txt for some of its tests.
It's a large ~70MiB file that when compressed with zstd (see the untracked changes) takes up only 15kiB.
In order to reduce the working directory size and the size of the published module change every instance
of where the largefile.txt is used in tests into the zstd equiv. Few things you'll have to remember:
the compress package is currently used as an indirect dependency, we'll need to run go mod tidy
to bring it into the direct deps section, tests cannot be run without setting up Snowflake DB;
I think we won't be able to do it here. Try reading from the zstd file wrapping any readers
with transparent zstd decompression. Try not to decompress the file in place.
```

The _zstded_ file decompresses exactly to the `largefile.txt`, but I nonetheless suggest code reviewers verifying this claim.

I did not setup tests locally for this change.


put_get_test.go
```diff
@@ -1200,3 +1207,65 @@ func createCancelTestFile(t *testing.T, sourceDir string) (string, func()) {
...
+	t.Helper()
+
+	reader, err := openLargeTestFile(sourceDir)
+	assertNilF(t, err)
+
+	return reader
+}
+
+func openLargeTestFile(sourceDir string) (io.ReadCloser, error) {
+	filePath := filepath.Join(sourceDir, "test_data", "largefile.txt.zstd")
```
sfc-gh-pfus, line 1243: As mentioned, if we want to fix it, I would prefer having randomly generated content instead of compressed file. My opinion it would be best to:
1. remove largefile.txt
2. generate random content if file not exists
3. save it in test_data dir
4. add it to gitignore

maciej: @sfc-gh-pfus OK, fair enough. I'm happy to contribute, but I'd appreciate if you'd point me to how to run tests (Snowflake setup); I'm a tourist here, just caring for small Go mod caches. 

```
go mod why github.com/snowflakedb/gosnowflake
# github.com/snowflakedb/gosnowflake
[INTERNAL | REDACTED]
github.com/burningalchemist/sql_exporter/cmd/sql_exporter
github.com/burningalchemist/sql_exporter
github.com/snowflakedb/gosnowflake
```

We don't use the lib directly ourselves, it's a sql_exporter requirement. I'll fork / tweak sql_exporter if I need to.

maciej:

put_get_test.go
sfc-gh-pfus: Well, to run integration tests locally, you have to have a Snowflake account, [trial](https://signup.snowflake.com/) should be good enough. Our tests are disabled for external contributions by default, but when we have something close to final version, I'll trigger them manually.

Our tests create random schema in SF during the init, but it can be disabled by setting `SKIP_SETUP=true` env variable. Because your changes are mostly about what happens before a connection is made, my suggestion is (if you don't want to start a trial):
1. Set `SKIP_SETUP=true`
2. Add printing of a query text (to ensure file paths are correct) temporarily - remove it when your change is ready
3. Run a test
4. Observe manually if a correct file is created in a directory and it matches printed query text

sfc-gh-pfus:
I'm closing this PR, because it was already solved by @sfc-gh-boler . Thanks anyway!

PR #1662, opened 2025-12-18
DOC-11379 adding some more context to async query execution vs KeepSessionAlive

sfc-gh-dszmolka:
### Description
No code was changed, this is a doc-only change.

Added to have the gosnowflake equivalent of the effort which goes for Python and JDBC drivers too, to clarify how they interact with async query execution in context of other related driver-side settings and server-side parameter

(More details in the doc and Slack attached to the internal jira)

### DIFF
--- a/auth_test.go
+++ b/auth_test.go
@@ -716,6 +716,12 @@ func TestUnitAuthenticateWithConfigMFA(t *testing.T) {
 	}
 }
 
+// This test creates two groups of scenarios:
+// a) singleAuthenticationPrompt=true - in this case, we start authenticating threads at once,
+// but due to locking mechanism only one should reach wiremock without MFA token.
+// b) singleAuthenticationPrompt=false - in this case, there is no locking, so all threads should rush,
+// but on Wiremock only first will be served with correct response (simulating a user confirming MFA only once).
+// The remaining threads should return error.
 func TestMfaParallelLogin(t *testing.T) {
 	skipOnMissingHome(t)
 	skipOnMac(t, "interactive keyring access not available on macOS runners")
--- a/driver_test.go
+++ b/driver_test.go
@@ -2351,7 +2351,10 @@ func initPoolWithSizeAndReturnErrors(db *sql.DB, poolSize int) []error {
 	for i := 0; i < poolSize; i++ {
 		go func(wg *sync.WaitGroup) {
 			defer wg.Done()
-			time.Sleep(time.Duration(rand.Intn(1000)) * time.Millisecond)
+			// Wiremock handles incoming request in parallel, in non-atomic way.
+			// If two requests start at the same time, they both see the same scenario state,
+			// even if it should be changed after the request is matched to a particular scenario state.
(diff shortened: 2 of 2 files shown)


doc.go
```diff
@@ -1357,6 +1357,28 @@ and before retrieving the results. For a more elaborative example please see cmd
 			...
 		}
 
+==> Some considerations related to the KeepSessionAlive configuration option in context of asynchronous query execution
+
+When calling Close() on the connection object, it performs the following actions:
```
sfc-gh-pfus: It's worth noting that this happens not only when `conn.Close` is called, but also for all connections in a pool (aka sql.DB) when the pool is being closed. Maybe we can rephrase it to "when SQL Go connection is being closed..."?

doc.go
```diff
@@ -1357,6 +1357,28 @@ and before retrieving the results. For a more elaborative example please see cmd
...
+
+* stops the scheduled heartbeats (CLIENT_SESSION_KEEP_ALIVE), if it was enabled
+
+* cleans up all the http connections which are already idle - doesn't touch the ones which are in active use currently
+
+* if Config.KeepSessionAlive is false (default), then actively logs out the current Snowflake session.
+
+!! Caveat: If there are any queries which are currently executing in the same Snowflake session (e.g. async queries sent with WithAsyncMode()), then those queries are automatically cancelled from the client side a couple minutes later after the Close() call, as a Snowflake session which has been actively logged out from, cannot sustain any queries.
+
+You can govern this behaviour with setting Config.KeepSessionAlive to true; when the corresponding Snowflake session will be kept alive for a long time even after an explicit Connection.Close() call; by default 24 hours past the time when the last running query in the session finished executing.
```
sfc-gh-pfus: I don't like repeating the Snowflake backend behaviour/docs here (I'm referring to the default 24h). Maybe we can just mention, that Snowflake session is not immediately deleted, but expires according to backend policy rules?

doc.go
```diff
@@ -1357,6 +1357,28 @@ and before retrieving the results. For a more elaborative example please see cmd
...
+
+Please consider that this might have unexpected, billable effects (might leave queries running up to configured value of STATEMENT_TIMEOUT_IN_SECONDS (https://docs.snowflake.com/en/sql-reference/parameters#statement-timeout-in-seconds).
```
sfc-gh-pfus: I think the remaining part is too verbose. IMHO we should keep the docs concise. Otherwise:
1. The more text it contains, the less readable it becomes. Especially if we expand all configuration options in a similar manner.
2. It repeats the docs from somewhere else, so in case of changes we have to change it in more places.
3. It repeats the external docs, so when the backend behaviour changes, we are leaving with old driver docs that can't be changed and are inaccurate.

WDYT?

sfc-gh-dszmolka: cut some redundant material (before ABORT_DETACHED_QUERY) but the rest is necessary i think, as it's driver specific and both for Python and JDBC, it is part of the driver specific docs

sfc-gh-rsavenok (approved):

PR #1664, opened 2025-12-22
SNOW-2465270 Add CGO_ENABLED marker to telemetry

sfc-gh-pfus:
### Description

SNOW-2465270 Add CGO_ENABLED marker to telemetry

### Checklist

### DIFF
--- a/internal/cgo/cgo_enabled.go
+++ b/internal/cgo/cgo_enabled.go
@@ -1,0 +1,6 @@
+//go:build cgo
+
+package cgo
+
+// Enabled is set to true if CGO is enabled.
+var Enabled = true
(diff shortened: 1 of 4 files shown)


internal/cgo/cgo_enabled.go
```diff
@@ -0,0 +1,8 @@
+//go:build cgo
+
+package cgo
+
+var _ = func() any {
```
sfc-gh-boler: I think it would be slightly cleaner to have cgo_enabled.go and cgo_not_enabled.go with inverse tags, instead of depending on initialization order.

sfc-gh-pfus: Ok, done.

sfc-gh-merbel (approved):

PR #1665, opened 2025-12-23
NO-SNOW Improved error logs

sfc-gh-pfus:
### Description

NO-SNOW Added more logs for easier finding of problem root causes.

### Checklist

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/snowflakedb/gosnowflake/internal/cgo"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


sfc-gh-sbirmiwal (approved):
LGTM

please update the PR description before merging

auth_wif.go
```diff
@@ -192,7 +192,7 @@ func (c *awsIdentityAttestationCreator) createAttestation() (*wifAttestation, er
 
 	attestationService := c.attestationServiceFactory(c.ctx, c.cfg)
 	if attestationService == nil {
-		return nil, fmt.Errorf("AWS attestation service could not be created")
```
sfc-gh-rsavenok, line 195: What benefits does `errors.New` give us vs `fmt.Errorf`? From what I'm reading, it allows to do better error checks e.g.
```
var ErrNotFound = errors.New("not found")

if errors.Is(err, ErrNotFound) {
  ...
}
```

But since we don't export these errors to variables, it won't work like that?

sfc-gh-pfus: It doesn't matter in this case. `errors.New` just creates a simple string error. `fmt.Errorf` creates a new error with wrapped errors, but there is no error to be wrapped, so it is useless. It has a tiny impact on the performance (a microsecond worth fighting for! ;) ).

sfc-gh-dheyman (approved):

PR #1666, opened 2025-12-23
SNOW-2395738 Expose tokenFilePath in Config, read it for every new connection

sfc-gh-pfus:
### Description

SNOW-2395738 Expose tokenFilePath in Config, read it for every new connection

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ Internal changes:
 -
 -
 -
--
+- Improved some logs returned from cloud storage clients (snowflakedb/gosnowflake#1665).
 -
 -
 -
--- a/auth_wif.go
+++ b/auth_wif.go
@@ -192,7 +192,7 @@ func (c *awsIdentityAttestationCreator) createAttestation() (*wifAttestation, er
 
 	attestationService := c.attestationServiceFactory(c.ctx, c.cfg)
 	if attestationService == nil {
-		return nil, fmt.Errorf("AWS attestation service could not be created")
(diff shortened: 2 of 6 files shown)


auth.go
```diff
@@ -511,7 +511,10 @@ func createRequestBody(sc *snowflakeConn, sessionParameters map[string]interface
 	case AuthTypeOAuth:
 		requestMain.LoginName = sc.cfg.User
 		requestMain.Authenticator = AuthTypeOAuth.String()
-		requestMain.Token = sc.cfg.Token
+		var err error
+		if requestMain.Token, err = sc.cfg.getToken(); err != nil {
```
sfc-gh-merbel, line 518: nit: `sc.cfg.getToken()` or `sc.cfg.Token` ?

sfc-gh-merbel (approved):

auth.go
sfc-gh-pfus: This is not a nit, this is a fundamental change ;) This is what the whole change is about. If a token is provided via a file path (and not a static value), we should read the file again for each new connection. It is critical for OAuth, where tokens usually last only for an hour. If a connection pool is in use, it's common that connections come and go. And when they come, they need a new token present in the file path.

sfc-gh-dheyman (approved):

auth.go
sfc-gh-merbel: Okey, now I see that this is our custom method that is implemented further in dsn.go. Thanks @sfc-gh-pfus

PR #1670, opened 2026-01-06
SNOW-2924941: Fix OCSP cache clearer stop() race condition

### LINKED ISSUE #1669: SNOW-2924941: OCSP cache clearer stop() has race condition due to buffered channel
## Description

The `stopOCSPCacheClearing` channel is defined as a buffered channel with capacity 2:

https://github.com/snowflakedb/gosnowflake/blob/master/ocsp.go#L109

```go
var stopOCSPCacheClearing = make(chan struct{}, 2)
```

This causes a race condition in the `stop()` method:

```go
func (occ *ocspCacheClearerType) stop() {
	occ.mu.Lock()
	running := occ.running
	occ.mu.Unlock()
	if running {
		stopOCSPCacheClearing <- struct{}{}  // Non-blocking, goes into buffer
		<-stopOCSPCacheClearing              // Could immediately read its own value!
	}
}
```

## Problem

With a buffered channel:
1. The send in `stop()` puts a value into the buffer (non-blocking since buffer has space)
2. The receive in `stop()` could immediately read that same value back
3. `stop()` returns without the goroutine ever seeing the stop signal

The goroutine continues running, and the synchronization handshake is broken.

## Expected Behavior

The `stop()` function should block until the cache clearer goroutine has actually stopped.

## Suggested Fix

Change the channel to be unbuffered:

```go
var stopOCSPCacheClearing = make(chan struct{})
```

With an unbuffered channel:
- The send in `stop()` blocks until the goroutine receives
- The goroutine's acknowledgment send blocks until `stop()` receives

This ensures proper synchronization for the handshake pattern.

## Impact

This bug can cause:
- Goroutine leaks when `StopOCSPCacheClearer()` is called
- Resources (ticker) not being released
- Non-deterministic behavior in tests

obs-gh-felixbowing:
The stopOCSPCacheClearing channel was defined as a buffered channel with capacity 2, which caused a race condition in the stop() method. With a buffered channel, the send operation in stop() would put a value into the buffer (non-blocking), and the receive operation could immediately read that same value back, causing stop() to return without the goroutine ever seeing the stop signal.

This fix changes the channel to be unbuffered, ensuring proper synchronization:
- The send in stop() blocks until the goroutine receives
- The goroutine's acknowledgment send blocks until stop() receives

This ensures the handshake pattern works correctly and prevents:
- Goroutine leaks when StopOCSPCacheClearer() is called
- Resources (ticker) not being released
- Non-deterministic behavior in tests

Fixes #1669

### DIFF
--- a/ocsp.go
+++ b/ocsp.go
@@ -106,7 +106,7 @@ const (
 	maxClockSkew           = 900 * time.Second // buffer for clock skew
 )
 
-var stopOCSPCacheClearing = make(chan struct{}, 2)
+var stopOCSPCacheClearing = make(chan struct{})
 
 type ocspStatusCode int
 
--- a/ocsp_test.go
+++ b/ocsp_test.go
@@ -624,3 +624,24 @@ func syncUpdateOcspResponseCache(f func()) {
 	defer ocspResponseCacheLock.Unlock()
 	f()
 }
+
+func TestOcspCacheClearerNoRaceCondition(t *testing.T) {
+	defer func() {
+		StopOCSPCacheClearer()
+		initOCSPCache()
+		StartOCSPCacheClearer()
+	}()
+
+	for i := 0; i < 5; i++ {
+		StartOCSPCacheClearer()
+		StopOCSPCacheClearer()
+
+		ocspCacheClearer.mu.Lock()
+		running := ocspCacheClearer.running
+		ocspCacheClearer.mu.Unlock()
+
+		if running {
+			t.Fatalf("Iteration %d: Cache clearer should be stopped but is still running", i)
+		}
+	}
+}


ocsp_race_test.go
```diff
@@ -0,0 +1,61 @@
+package gosnowflake
```
sfc-gh-pfus: Can you move this test to ocsp_test.go?

ocsp_race_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"os"
+	"testing"
+	"time"
+)
+
+// TestOcspCacheClearerNoRaceCondition tests that the stop() method properly
+// synchronizes with the goroutine and doesn't have a race condition.
+// This test verifies the fix for issue #1669.
```
sfc-gh-pfus: I don't think this comment is needed. I think the function name is enough. We generally don't add such comments.

ocsp_race_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestOcspCacheClearerNoRaceCondition(t *testing.T) {
+	// Save original environment variable
+	origValue := os.Getenv(ocspResponseCacheClearingIntervalInSecondsEnv)
```
sfc-gh-pfus: We have `envOverride` for this, can you use it? Also, the var name could be a bit more descriptive what value it holds :)

ocsp_race_test.go
```diff
@@ -0,0 +1,61 @@
...
+	defer func() {
+		StopOCSPCacheClearer()
+		os.Setenv(ocspResponseCacheClearingIntervalInSecondsEnv, origValue)
+		initOCSPCache()
+		StartOCSPCacheClearer()
+	}()
+
+	// Set a short interval for testing
+	os.Setenv(ocspResponseCacheClearingIntervalInSecondsEnv, "10")
```
sfc-gh-pfus: Is the 10s interval needed? If I understand correctly, this value is not relevant for the test?

ocsp_race_test.go
```diff
@@ -0,0 +1,61 @@
...
+	ocspCacheClearer.mu.Unlock()
+
+	if running {
+		t.Fatal("Cache clearer should be stopped but is still running")
+	}
+
+	// Try to stop again - should be a no-op since it's not running
+	StopOCSPCacheClearer()
+
+	// Start and stop multiple times to ensure no goroutine leaks
```
sfc-gh-pfus: Actually it is the same what happens above? Do we need all these iterations or the code above?

obs-gh-felixbowing: Good point, consolidated into just the loop.

ocsp_race_test.go
obs-gh-felixbowing: Removed

sfc-gh-rsavenok (approved):

sfc-gh-pfus:
@obs-gh-felixbowing sorry, I missed your fixes! Can you rebase your change and add a word to changelog.md?

sfc-gh-pfus:
I'm closing this PR in favour of https://github.com/snowflakedb/gosnowflake/pull/1704

PR #1677, opened 2026-01-27
SNOW-3009271 Add more OS info gathering on Linux

sfc-gh-pfus:
### Description

SNOW-3009271 This commit introduces OS information gathering for Linux.

### Checklist

### DIFF
--- a/internal/os/os_details_linux.go
+++ b/internal/os/os_details_linux.go
@@ -1,0 +1,12 @@
+//go:build linux
+
+package os
+
+// GetOsDetails returns OS details from /etc/os-release on Linux.
+// The result is cached so it's only read once.
+func GetOsDetails() map[string]string {
+	osDetailsOnce.Do(func() {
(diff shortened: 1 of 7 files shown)


internal/os/os_details_linux.go
```diff
@@ -0,0 +1,86 @@
...
+func GetOsDetails() map[string]string {
+	osDetailsOnce.Do(func() {
+		osDetails = readOsRelease()
+	})
+	return osDetails
+}
+
+// readOsRelease reads and parses /etc/os-release file.
+// Returns nil on any error.
+func readOsRelease() map[string]string {
```
sfc-gh-rsavenok: Do you plan to add any tests to check that parsing works correctly? e.g. test with a dummy os-release contents containing values with `""` and without and checking that the produced map is what you expect?

sfc-gh-pfus: Added.

sfc-gh-dheyman (approved):

PR #1682, opened 2026-01-30
SNOW-0: Patch potential memory leaks in gosnowflake upload utilities

sfc-gh-abhmishra:
### Description
-  Patch potential memory leaks in gosnowflake upload utilities
- PEP (snowsight backend) was getting container restarts while executing this endpoint - [slack context and tracing here](https://snowflake.slack.com/archives/C06950CKXEJ/p1769802010535119?thread_ts=1769796203.973699&cid=C06950CKXEJ)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Bug fixes:
 -
 - Fixed minicore file name gathering in client environment (snowflakedb/gosnowflake#1661).
 -
--
+- Fixed file descriptor leaks in storage utilties (snowflakedb/gosnowflake#1682)
 -
 - Fixed path escaping for GCS urls (snowflakedb/gosnowflake#1678).
 


gcs_storage_client.go
```diff
@@ -92,6 +94,16 @@ func (util *snowflakeGcsClient) getFileHeader(meta *fileMetadata, filename strin
 		if err != nil {
 			return nil, err
 		}
+		if resp == nil {
```
sfc-gh-pfus: Based on `client.Do` docs, it is guaranteed to never return nil, if error is nil.

gcs_storage_client.go
```diff
@@ -215,10 +227,16 @@ func (util *snowflakeGcsClient) uploadFile(
 			uploadSrc = meta.realSrcStream
 		}
 	} else {
-		uploadSrc, err = os.Open(dataFile)
+		file, err := os.Open(dataFile)
```
sfc-gh-pfus: Do we need this separate variable? Maybe we can register defer and just pass `uploadSrc` as an argument?

gcs_storage_client.go
```diff
@@ -243,6 +261,16 @@ func (util *snowflakeGcsClient) uploadFile(
 	if err != nil {
 		return err
 	}
+	if resp == nil {
```
sfc-gh-pfus: Apply the mentioned suggestions to all your changes.

gcs_storage_client.go
```diff
@@ -215,10 +222,16 @@ func (util *snowflakeGcsClient) uploadFile(
...
+		var err error
 		uploadSrc, err = os.Open(dataFile)
 		if err != nil {
 			return err
 		}
+		defer func(src io.Closer) {
+			if err := src.Close(); err != nil {
+				logger.Warnf("failed to close %v file: %v", dataFile, err)
+			}
+		}(uploadSrc.(io.Closer))
```
sfc-gh-abhmishra, line 234: had to cast it here because **uploadSrc is defined as an ioReader** whereas the return type of **os.Open()** is different

gcs_storage_client.go
```diff
@@ -636,9 +658,6 @@ func (util *snowflakeGcsClient) downloadRangeStream(
 	if err != nil {
 		return nil, err
 	}
-	if resp == nil {
```
sfc-gh-abhmishra, line 639: removed this too - it was present in the code already, let me know if you want me to add back

gcs_storage_client.go
sfc-gh-abhmishra: fixed

gcs_storage_client.go
sfc-gh-abhmishra: It's because the uploadSrc declaration is a ioReader above - **but now we want to call close() on it** , left more comments on the patch

gcs_storage_client.go
sfc-gh-abhmishra: done

sfc-gh-pbulawa (approved):

PR #1687, opened 2026-02-04
[SNOW-3049473] Add recovery for file upload/download panics

sfc-gh-dramesha:
### Description

- Added recovery blocks to handle file uploads and downloads panics
- Added checks to return error where it used to panic before instead

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Upcoming Release
 
+Bug fixes:
+
+- Added panic recovery block for stage file uploads and downloads operation (snowflakedb/gosnowflake#1687).
+
 ## 1.19.0
 
 New features:


sfc-gh-dramesha:
recheck

file_transfer_agent.go
```diff
@@ -794,6 +794,12 @@ func (sfa *snowflakeFileTransferAgent) uploadFilesParallel(fileMetas []*fileMeta
 				wg.Add(1)
 				go func(k int, m *fileMetadata) {
 					defer wg.Done()
+					defer func() {                                                               
+						if r := recover(); r != nil {                                        
+                            errors[k] = fmt.Errorf("panic during file upload: %v", r)
```
sfc-gh-pfus, line 799: Can we use `%w` here, just for the sake of unwrapping?

file_transfer_agent.go
```diff
@@ -956,6 +962,12 @@ func (sfa *snowflakeFileTransferAgent) downloadFilesParallel(fileMetas []*fileMe
 				wg.Add(1)
 				go func(k int, m *fileMetadata) {
 					defer wg.Done()
+					defer func() {                                                               
+						if r := recover(); r != nil {                                        
+                            errors[k] = fmt.Errorf("panic during file download: %v", r)
```
sfc-gh-pfus, line 967: Same here.

storage_client.go
```diff
@@ -245,6 +245,10 @@ func (rsu *remoteStorageUtil) downloadOneFile(meta *fileMetadata) error {
 						logger.Errorf("Stream decryption failed for %s - temp file will be cleaned up to prevent corrupted data: %v", meta.srcFileName, err)
 						return err
 					}
+					logger.Debugf("Total file size: %d", totalFileSize)
+					if totalFileSize < 0 || totalFileSize > meta.sfa.streamBuffer.Len() {
```
sfc-gh-pfus, line 249: Why do we need this check?

sfc-gh-dramesha: So, when we were debugging, I saw [stacktraces for panics](https://docs.google.com/document/d/1vmXhpLFkMYgtxvU1wzpWQSbXrNF8HFZBTojEldyYhuY/edit?tab=t.0#bookmark=id.cdsqnr3b047g) from here - the bytes.Buffer.Truncate(n) panics if `n < 0 or n > len` (https://go.dev/src/bytes/buffer.go). Since we've added recovery block we don't necessarily need this check but I added just in case

file_transfer_agent.go
sfc-gh-dramesha: Since recover() returns any (interface{}), not error. The recovered value could be:                                          
  - An error (from panic(err))                                                                                           
  - A string (from panic("something went wrong"))                                                                        
  - Any other type (from panic(123))                                                                                     
                                                                                                                         
 %w requires the argument to be an error type and will fail at runtime if it's not. The simpler %v approach works for all types. So, kept it %v.
 
 I can also change it to something like below, let me know:
 ```
   defer func() {                                                                                                                      
      if r := recover(); r != nil {                                                                                                   
          var err error                                                                                                               
          if e, ok := r.(error); ok {                                                                                                 
              err = fmt.Errorf("panic during file upload: %w", e)                                                                   
          } else {                                                                                                                    
              err = fmt.Errorf("panic during file upload: %v", r)                                                                   
          }                                                                                                                           
          errors[k] = err                                                                        
          results[k] = nil                                                                                                            
      }                                                                                                                               
  }()
  ```

sfc-gh-pfus (approved):

sfc-gh-rsavenok (approved):

PR #1704, opened 2026-02-16
SNOW-2924941 Use cancellable context in OCSP cache clearer

sfc-gh-pfus:
### Description

SNOW-2924941 Use cancellable context in OCSP cache clearer.

### Checklist


ocsp.go
```diff
@@ -1131,6 +1131,7 @@ func (occ *ocspCacheClearerType) start() {
 	if occ.running {
 		return
 	}
+	occ.ctx, occ.cancel = context.WithCancel(context.Background())
```
sfc-gh-boler: No need to store the context `ocspCacheClearerType`. It's only referenced by the goroutine spawned by this function, so we can pass it as a closure.

ocsp.go
```diff
@@ -1121,6 +1119,8 @@ func initOcspModule() {
 }
 
 type ocspCacheClearerType struct {
+	ctx     context.Context
+	cancel  context.CancelFunc
 	running bool
 	mu      sync.Mutex
```
sfc-gh-boler, line 1122: Let's put the `mu` before any fields it protects, so that it is clearer what it is responsible for.

PR #1705, opened 2026-02-17
SNOW-1881542 Migrate context keys to private iota

sfc-gh-pfus:
### Description

SNOW-1881542 Migrate context keys to private iota

Migrated context keys from strings to private iota. Using a private iota fixes the problem with context key clashes.

### Checklist

### DIFF
--- a/util.go
+++ b/util.go
@@ -14,32 +14,50 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/memory"
 )
 
-type contextKey string
+type contextKey int
 
 const (
-	multiStatementCount              contextKey = "MULTI_STATEMENT_COUNT"
-	asyncMode                        contextKey = "ASYNC_MODE_QUERY"
-	queryIDChannel                   contextKey = "QUERY_ID_CHANNEL"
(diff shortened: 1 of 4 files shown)


util.go
```diff
@@ -14,33 +14,51 @@ import (
...
+	sfSessionIDKey:   "LOG_SESSION_ID",
+	sfSessionUserKey: "LOG_USER",
+}
+
+// String returns a human-readable name for context keys used in logging.
+func (k contextKey) String() string {
+	if name, ok := contextKeyNames[k]; ok {
+		return name
+	}
+	return fmt.Sprintf("contextKey(%d)", int(k))
```
sfc-gh-rsavenok, line 59: Do I understand correctly that this will print contextKey(N) in case the name is not in the map?

Is this some Go best practice? I find it wierd that we'll be printing number that are hard to reason about wihout manually counting the index. + the index might change between driver versions

sfc-gh-rsavenok (approved):

util.go
sfc-gh-pfus: If the map does not contain a key, yes. This is because only the two entries are printed in logs.

sfc-gh-dheyman (approved):

PR #1710, opened 2026-02-23
SNOW-1881542 Migrate from logrus to slog, move logger to internal package

sfc-gh-pfus:
### Description

SNOW-1881542 Migrate from logrus to slog, move logger to internal package

This change includes:
1. Simplified SFLogger interface
2. Migration from logrus to slog
3. Moving logging internals to internal package

### Checklist

### DIFF
--- a/easy_logging.go
+++ b/easy_logging.go
@@ -87,14 +87,18 @@ func easyLoggingInitError(err error) error {
 
 func reconfigureEasyLogging(logLevel string, logPath string) error {
 	// don't allow any change if a non-default logger is already being used.
-	if _, ok := logger.(loggerinternal.EasyLoggingSupport); ok {
+	currentLogger := GetLogger()
+	if !loggerinternal.IsEasyLoggingLogger(currentLogger) {
+		logger.Warnf("Cannot reconfigure easy logging: custom logger is in use")
 		return nil // cannot replace custom logger
 	}
+
 	newLogger := CreateDefaultLogger()
 	err := newLogger.SetLogLevel(logLevel)
 	if err != nil {
 		return err
 	}
+
 	var output io.Writer
 	var file *os.File
 	output, file, err = createLogWriter(logPath)
@@ -106,6 +110,13 @@ func reconfigureEasyLogging(logLevel string, logPath string) error {
 	if err != nil {
 		logger.Errorf("%s", err)
 	}
+
+	// Actually set the new logger as the global logger
+	if err := SetLogger(newLogger); err != nil {
+		logger.Errorf("Failed to set new logger: %s", err)
+		return err
+	}
+
 	return nil
 }
 
--- a/easy_logging_test.go
+++ b/easy_logging_test.go
@@ -16,6 +16,10 @@ import (
 func TestInitializeEasyLoggingOnlyOnceWhenConfigGivenAsAParameter(t *testing.T) {
 	skipOnWindows(t, "Doesn't work on Windows")
 	defer cleanUp()
+	origLogLevel := logger.GetLogLevel()
+	defer logger.SetLogLevel(origLogLevel)
+	logger.SetLogLevel("error")
+
 	logDir := t.TempDir()
 	logLevel := levelError
 	contents := createClientConfigContent(logLevel, logDir)
@@ -42,6 +46,10 @@ func TestInitializeEasyLoggingOnlyOnceWhenConfigGivenAsAParameter(t *testing.T) 
 func TestConfigureEasyLoggingOnlyOnceWhenInitializedWithoutConfigFilePath(t *testing.T) {
 	skipOnWindows(t, "Doesn't work on Windows")
 	skipOnMissingHome(t)
+	origLogLevel := logger.GetLogLevel()
+	defer logger.SetLogLevel(origLogLevel)
+	logger.SetLogLevel("error")
+
 	appExe, err := os.Executable()
 	assertNilF(t, err, "application exe not accessible")
 	userHome, err := os.UserHomeDir()
@@ -87,6 +95,10 @@ func TestReconfigureEasyLoggingIfConfigPathWasNotGivenForTheFirstTime(t *testing
 	skipOnWindows(t, "Doesn't work on Windows")
 	skipOnMissingHome(t)
 	defer cleanUp()
+	origLogLevel := logger.GetLogLevel()
+	defer logger.SetLogLevel(origLogLevel)
+	logger.SetLogLevel("error")
+
 	configDir, err := os.UserHomeDir()
 	logDir := t.TempDir()
 	assertNilF(t, err, "user home directory error")
@@ -147,6 +159,10 @@ func TestEasyLoggingFailOnNotExistingConfigFile(t *testing.T) {
 func TestLogToConfiguredFile(t *testing.T) {
 	skipOnWindows(t, "Doesn't work on Windows")
 	defer cleanUp()
+	origLogLevel := logger.GetLogLevel()
+	defer logger.SetLogLevel(origLogLevel)
+	logger.SetLogLevel("error")
+
 	dir := t.TempDir()
 	easyLoggingInitTrials.reset()
 	configContent := createClientConfigContent(levelWarn, dir)
--- a/internal/logger/easy_logging_support.go
+++ b/internal/logger/easy_logging_support.go
@@ -15,6 +15,13 @@ func CloseFileOnLoggerReplace(sflog interface{}, file *os.File) error {
 	return fmt.Errorf("logger does not support closeFileOnLoggerReplace")
 }
 
+// IsEasyLoggingLogger checks if the given logger is based on the default logger implementation.
+// This is used by easy logging to determine if reconfiguration is allowed.
+func IsEasyLoggingLogger(sflog interface{}) bool {
+	_, ok := unwrapToEasyLoggingLogger(sflog)
+	return ok
+}
+
 // unwrapToEasyLoggingLogger unwraps a logger to get to the underlying default logger if present
 func unwrapToEasyLoggingLogger(sflog interface{}) (EasyLoggingSupport, bool) {
 	current := sflog


loginterface/interfaces.go
```diff
@@ -0,0 +1,60 @@
+package loginterface
+
+import (
+	"context"
+	"io"
+	"log/slog"
+)
```
rickr-sigma: One of the goals of this interface'd approach was to allow callers to build implementations independently from slog.  I would imagine we want to avoid references to slog am much as possible.

loginterface/interfaces.go
```diff
@@ -0,0 +1,60 @@
...
+	Warnf(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
+
+	Trace(args ...interface{})
+	Debug(args ...interface{})
+	Info(args ...interface{})
+	Warn(args ...interface{})
+	Error(args ...interface{})
+	Fatal(args ...interface{})
```
rickr-sigma: looks like we are quite close to getting rid of this batch of interfaces (i.e. the non-f ones).  I don't see any uses and only implementations of the interface...am I missing something?

log.go
```diff
@@ -1,514 +1,113 @@
...
+// logger is a proxy that delegates all calls to the internal global logger
+// This ensures a single source of truth for the current logger
+var logger SFLogger = loggerinternal.NewLoggerProxy()
 
 // SetLogger set a new logger of SFLogger interface for gosnowflake
-// Deprecated: will be reorganized in the future releases.
+// The provided logger will automatically be wrapped with secret masking.
 func SetLogger(inLogger *SFLogger) {
-	logger = *inLogger //.(*defaultLogger)
+	_ = loggerinternal.SetLoggerWithMasking(*inLogger)
```
rickr-sigma: why would SetLogger() auto-wrap with a Masking logger?  Can we instead have a func that explicitly does just that. E.g. SetLoggerWithMasking(SFLogger)  ?  

My argument would be that the client is passing in all this (maskable) data, it is up to it to decide on the logging policy.

loginterface/interfaces.go
```diff
@@ -0,0 +1,60 @@
...
+// Example usage:
+//
+//	logger := gosnowflake.GetLogger()
+//	if slogLogger, ok := logger.(gosnowflake.SFSlogLogger); ok {
+//	    customHandler := slog.NewJSONHandler(os.Stdout, nil)
+//	    slogLogger.SetHandler(customHandler)
+//	}
+type SFSlogLogger interface {
+	SetHandler(handler slog.Handler) error
+}
```
rickr-sigma: should we at least put this interface in a different file?  Say loginterface/extensions.go ?

loginterface/interfaces.go
sfc-gh-pfus: Slog is not required, it is in a separate interface from the main `SFLogger`. But I moved it into a separate file to make it more clear.

loginterface/interfaces.go
sfc-gh-pfus: Well, you may be right. These non-f functions are for logging a plain string, so I can replace the args to be strings only.

log.go
sfc-gh-pfus: I can ask what our security thinks about this, we'll see.

loginterface/interfaces.go
sfc-gh-pfus: Done!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,5 +3,6 @@
 SNOW-XXX Please explain the changes you made here.
 
 ### Checklist
+- [ ] Added proper logging (if possible)
```
sfc-gh-rsavenok, line 6: Maybe be a bit more explicit about what "proper" means in the context of this driver?

sfc-gh-rsavenok (approved):

log_test.go
sfc-gh-boler: since you are cleaning up this area, this test looks wrong

log_test.go
```diff
@@ -123,32 +72,21 @@ func TestOffLogLevel(t *testing.T) {
...
-	logger.Warningln("warningln")
 	logger.Error("error")
 	logger.Errorf("error%v", "f")
-	logger.Errorln("errorln")
 
 	assertEqualE(t, buf.Len(), 0, "log messages count")
 	assertEqualE(t, logger.GetLogLevel(), "OFF", "log level")
 }
 
 func TestLogSetLevel(t *testing.T) {
```
sfc-gh-boler, line 88: I don't see a test verifying that lower level errors are suppressed. We can easily change this one to default to debug, and see it allows warnings, but filters traces.

internal/logger/secret_detector.go
```diff
@@ -0,0 +1,132 @@
...
+	privateKeyDataRegexp   *regexp.Regexp
+	privateKeyParamRegexp  *regexp.Regexp
+	connectionTokenRegexp  *regexp.Regexp
+	passwordRegexp         *regexp.Regexp
+	dsnPasswordRegexp      *regexp.Regexp
+	clientSecretRegexp     *regexp.Regexp
+	jwtTokenRegexp         *regexp.Regexp
+)
+
+func registerRegexps() {
```
sfc-gh-boler: Let's compile all regexes when the package is loaded instead of lazily. It's better to crash on startup when during an eventual logging event.

internal/logger/secret_detector.go
```diff
@@ -0,0 +1,132 @@
...
+	privateKeyDataRegexp = regexp.MustCompile(privateKeyDataPattern)
+	privateKeyParamRegexp = regexp.MustCompile(privateKeyParamPattern)
+	connectionTokenRegexp = regexp.MustCompile(connectionTokenPattern)
+	passwordRegexp = regexp.MustCompile(passwordPattern)
+	dsnPasswordRegexp = regexp.MustCompile(dsnPasswordPattern)
+	clientSecretRegexp = regexp.MustCompile(clientSecretPattern)
+	jwtTokenRegexp = regexp.MustCompile(jwtTokenPattern)
+}
+
+type secretmasker string
```
sfc-gh-boler: What's the purpose of this type vs just going over a list of regexes?

sflog/interface.go
sfc-gh-boler, line 1: Calling this file `main.go` is pretty confusing given the convention around main packages. How about "interface.go"?

loginterface/interface.go
```diff
@@ -0,0 +1,41 @@
+package loginterface
```
sfc-gh-boler: Please add a package comment.

internal/logger/accessor.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// Example usage:
+//   log := logger.GetLogger()
+//   log.Info("Message from internal package")
+//   log.WithField("key", "value").Info("Structured log")
+func GetLogger() loginterface.SFLogger {
+	loggerAccessorMu.RLock()
+	defer loggerAccessorMu.RUnlock()
+
+	if globalLogger == nil {
```
sfc-gh-boler: Why don't we can call Do() unconditionally? If there is some expectation that an existing `globalLogger` should not be replaced, then move that logic to the anonymous function in Do()

internal/logger/accessor.go
```diff
@@ -0,0 +1,113 @@
...
+// without importing the main gosnowflake package (avoiding circular dependencies)
+var (
+	globalLogger       loginterface.SFLogger
+	loggerAccessorOnce sync.Once
+	loggerAccessorMu   sync.RWMutex
+)
+
+// SetGlobalLoggerAccessor sets the logger accessor for internal packages
+// This should be called by the main package during initialization
+func SetGlobalLoggerAccessor(logger loginterface.SFLogger) {
```
sfc-gh-boler: `SetGlobalLoggerAccessor` and `SetLoggerWithMasking` do a similar thing in general, but their names have an odd asymmetry. Maybe `SetRawLogger` and `SetMaskedLogger` would be better names?

internal/logger/accessor.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/snowflakedb/gosnowflake/v2/loginterface"
+)
+
+// LoggerAccessor allows internal packages to access the global logger
+// without importing the main gosnowflake package (avoiding circular dependencies)
+var (
+	globalLogger       loginterface.SFLogger
+	loggerAccessorOnce sync.Once
+	loggerAccessorMu   sync.RWMutex
```
sfc-gh-boler: Put the mutex before the fields it protects, so that it's scope is a bit clearer.

sfc-gh-boler: In general RWMutex is not a good choice most of the time. Use sync.Mutex instead.

internal/logger/accessor.go
```diff
@@ -0,0 +1,113 @@
...
+func GetLogger() loginterface.SFLogger {
+	loggerAccessorMu.RLock()
+	defer loggerAccessorMu.RUnlock()
+
+	if globalLogger == nil {
+		// Return a default logger if not initialized yet
+		loggerAccessorOnce.Do(func() {
+			inner := NewDefaultLogger()
+			wrapped := NewSecretMaskingLogger(inner)
+			if sfLogger, ok := wrapped.(loginterface.SFLogger); ok {
```
sfc-gh-boler: NewSecretMaskingLogger() always returns `secretMaskingLogger` or panics. So it can't return any other interface. Why don't we make `NewSecretMaskingLogger` return a concrete type, so that the compiler can enforce correctness for us?

internal/logger/accessor.go
```diff
@@ -0,0 +1,113 @@
...
+	defer loggerAccessorMu.Unlock()
+
+	inner := NewDefaultLogger()
+	wrappedInterface := NewSecretMaskingLogger(inner)
+	wrapped, _ := wrappedInterface.(loginterface.SFLogger)
+	globalLogger = wrapped
+}
+
+// loggerError is a simple error type for logger-related errors
+type loggerError struct {
```
sfc-gh-boler: What's the purpose of this type? It doesn't seem to be used anywhere where a simple errors.New() wouldn't already work.

log.go
```diff
@@ -1,514 +1,110 @@
...
-}
+func init() {
+	// Set default log keys in internal package
+	SetLogKeys(SFSessionIDKey, SFSessionUserKey)
 
-func (log *defaultLogger) Logln(level rlog.Level, args ...interface{}) {
-	log.inner.Logln(level, args...)
-}
+	// Initialize the default logger in internal package
+	loggerinternal.CreateAndSetDefaultLogger()
```
sfc-gh-boler: Why don't we do the same from one of the internal package's init()?

internal/logger/context.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"log/slog"
+	"sync"
+)
+
+// Storage for log keys and hooks (single source of truth)
+var (
+	logKeys               []interface{}
+	clientLogContextHooks map[string]ClientLogContextHook
+	contextConfigMu       sync.RWMutex
```
sfc-gh-boler: Just like in the other file, put mutex ahead of the fields it protects and use Mutex instead of RWMutex

internal/logger/levels.go
```diff
@@ -0,0 +1,59 @@
...
+		return LevelDebug, nil
+	case "INFO":
+		return LevelInfo, nil
+	case "WARN":
+		return LevelWarn, nil
+	case "ERROR":
+		return LevelError, nil
+	case "FATAL":
+		return LevelFatal, nil
+	default:
```
sfc-gh-boler: "OFF" is missing

internal/logger/levels.go
```diff
@@ -0,0 +1,59 @@
...
+		return "debug"
+	case LevelInfo:
+		return "info"
+	case LevelWarn:
+		return "warn"
+	case LevelError:
+		return "error"
+	case LevelFatal:
+		return "fatal"
+	default:
```
sfc-gh-boler: "off"?

internal/logger/slog_logger.go
```diff
@@ -0,0 +1,457 @@
...
+						Line:     src.Line,
+						Function: src.Function,
+					}
+					_, location := callerPrettyfier(frame)
+					return slog.String(slog.SourceKey, location)
+				}
+			}
+			return a
+		},
+	}
```
sfc-gh-boler, line 150: NewDefaultLogger has pretty much the same setup. Can that be refactored to reduce duplicatoin?

internal/logger/slog_logger.go
```diff
@@ -0,0 +1,457 @@
...
+// isEnabled checks if logging is enabled (for OFF level)
+func (log *defaultLogger) isEnabled() bool {
+	log.mu.Lock()
+	defer log.mu.Unlock()
+	return log.enabled
+}
+
+// SetLogLevel sets the log level
+func (log *defaultLogger) SetLogLevel(level string) error {
+	upperLevel := strings.ToUpper(level)
```
sfc-gh-boler: parseLevel to enum, so that we can operate on the enum values.

internal/logger/secret_detector.go
sfc-gh-pfus: I think the biggest pro is chainability instead of this "lisp-style" (embedding a function, in a function, in a function...).

.github/PULL_REQUEST_TEMPLATE.md
sfc-gh-pfus: I don't know if there are short guidlines that could be included here 🤷 This is just a reminder to think about this.

log_test.go
sfc-gh-pfus: You mean the message? Fixed.

internal/logger/accessor.go
sfc-gh-pfus: Actually, there's no point in calling this in `GetLogger`. It should be set up in the init func.

internal/logger/accessor.go
sfc-gh-pfus: I reorganized and simplified this area.

internal/logger/accessor.go
sfc-gh-pfus: Why? I'd like to have very parallel read of logger and block everything only in case of setting a logger (which should be pretty never).

internal/logger/accessor.go
sfc-gh-pfus: Reorganized, thanks!

internal/logger/secret_detector.go
sfc-gh-boler: It doesn't have to be nested function call. A simple iteration over a list of regexes to apply would be perfectly fine. 

As it is, it's pretty hard to tell if all masks are being applied.

internal/logger/accessor.go
sfc-gh-boler: Overhead of RWMutex.RLock() is larger than Mutex.Lock(). So if the critical section is fast (here we are just returning a single value), RWMutex may be making the execution slower overall. 

As a rule of thumb, if we care about performance to that level, then we should have a microbenchmark. :)

internal/logger/level_filtering.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (l *levelFilteringLogger) Errorf(format string, args ...interface{}) {
+	if !l.shouldLog("error") {
+		return
+	}
+	l.inner.Errorf(format, args...)
+}
+
+func (l *levelFilteringLogger) Fatalf(format string, args ...interface{}) {
+	if !l.shouldLog("fatal") {
```
sfc-gh-boler: This one is interesting. Since "OFF" is a level higher than "FATAL", it means disabling logging will disable crashing. Maybe we should not support "OFF" or remove `shouldLog` check here and log "FATAL" regardless?

internal/logger/level_filtering.go
```diff
@@ -0,0 +1,268 @@
...
+var _ SFLogger = (*levelFilteringLogger)(nil)
+
+// Unwrap returns the inner logger (for introspection by easy_logging)
+func (l *levelFilteringLogger) Unwrap() interface{} {
+	return l.inner
+}
+
+// shouldLog determines if a message at messageLevel should be logged
+// given the current configured level
+func (l *levelFilteringLogger) shouldLog(messageLevel string) bool {
```
sfc-gh-boler: While `shouldLog` is supposed to give us some efficiency boost by not logging, current implementation is pretty inefficient. We have two string conversions (`ToUpper`) and comparisons.  It also eventually acquires a mutex (from `rawLogger`).

We could make `SetLevel` update an atomic value and then just compare that value with the desired `slog.Level`.

log.go
```diff
@@ -1,514 +1,124 @@
...
+//
+// You cannot bypass these protective layers. If you need to configure them, use the
+// returned logger's methods (SetLogLevel, etc.).
+//
+// Example:
+//
+//	customLogger := mylogger.New()
+//	gosnowflake.SetLogger(customLogger)
+func SetLogger(logger SFLogger) {
+	_ = loggerinternal.SetLogger(logger)
```
sfc-gh-boler: What if this fails? Should we make this SetLogger also return `error`?

driver.go
```diff
@@ -114,15 +114,15 @@ func skipRegistration() bool {
...
 
-var logger = CreateDefaultLogger()
-
 func init() {
 	if !skipRegistration() {
 		sql.Register("snowflake", &SnowflakeDriver{})
 	}
-	_ = logger.SetLogLevel("error")
+
+	// Set initial log level
```
sfc-gh-boler, line 122: Duplicated logic in log.go. Drop this?

internal/logger/secret_masking.go
```diff
@@ -0,0 +1,235 @@
...
+func (e *secretMaskingEntry) Tracef(format string, args ...interface{}) {
+	message := fmt.Sprintf(format, args...)
+	maskedMessage := MaskSecrets(message)
+	e.inner.Tracef("%s", maskedMessage)
+}
+
+func (e *secretMaskingEntry) Debugf(format string, args ...interface{}) {
+	message := fmt.Sprintf(format, args...)
+	maskedMessage := MaskSecrets(message)
+	e.inner.Debugf("%s", maskedMessage)
```
sfc-gh-boler: Let's use the non-formatting variants of the methods, i.e.: Debug/Print, etc.

internal/logger/level_filtering.go
sfc-gh-pfus: We have to support OFF, because it is needed for easy logging :/ But good catch about not shutting down the entire app!

internal/logger/level_filtering.go
sfc-gh-pfus: Ok, I refactored it.

log.go
sfc-gh-pfus: Right, propagated.

driver.go
sfc-gh-pfus: Ok, I removed it there. For me driver.go is a good starting point to look at the driver, so I prefer keeping it here.

internal/logger/secret_detector.go
sfc-gh-boler: Nice code and complexity reduction!

internal/logger/level_filtering.go
sfc-gh-boler: So... what are we going to do here? It's pretty important to respect Fatal's contract and crash.

internal/logger/slog_logger.go
```diff
@@ -0,0 +1,395 @@
...
+		log.enabled = false
+		log.mu.Unlock()
+		return nil
+	}
+
+	log.mu.Lock()
+	log.enabled = true
+	log.mu.Unlock()
+
+	log.level = upperLevel
```
sfc-gh-boler: BTW, this should be done under the mutex.

internal/logger/level_filtering.go
sfc-gh-pfus: I logged it still. But this is not critical, we use fatal only in diagnostic tool, which is fine to fatal if there are errors. In the main driver, we never call fail.

internal/logger/slog_logger.go
sfc-gh-pfus: Good catch!

PR #1711, opened 2026-02-24
NO-SNOW Add support for using private key auth in arrow batches tests

sfc-gh-pfus:
### Description

NO-SNOW Add support for using private key auth in arrow batches tests

### Checklist

### DIFF
--- a/arrowbatches/batches_test.go
+++ b/arrowbatches/batches_test.go
@@ -41,8 +41,11 @@ func repoRoot(t *testing.T) string {
 		t.Fatalf("failed to get working directory: %v", err)
 	}
 	for {
-		if _, err := os.Stat(filepath.Join(dir, "go.mod")); err == nil {
+		if _, err = os.Stat(filepath.Join(dir, "go.mod")); err == nil {
 			return dir
+		}
+		if !os.IsNotExist(err) {
+			t.Fatalf("failed to stat go.mod in %q: %v", dir, err)
 		}
 		parent := filepath.Dir(dir)
 		if parent == dir {
@@ -97,7 +100,7 @@ func testConfig(t *testing.T) *sf.Config {
 	}
 	cfg, err := sf.GetConfigFromEnv(configParams)
 	if err != nil {
-		t.Skipf("Snowflake config not available: %v", err)
+		t.Fatalf("failed to get config from environment: %v", err)
 	}
 	if isJWT {
 		privKeyPath := os.Getenv("SNOWFLAKE_TEST_PRIVATE_KEY")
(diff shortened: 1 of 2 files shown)


arrowbatches/batches_test.go
```diff
@@ -21,33 +26,98 @@ import (
...
+// repoRoot walks up from the current working directory to find the directory
+// containing go.mod, which is the repository root.
+func repoRoot(t *testing.T) string {
+	t.Helper()
+	dir, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("failed to get working directory: %v", err)
+	}
+	for {
+		if _, err := os.Stat(filepath.Join(dir, "go.mod")); err == nil {
```
sfc-gh-boler: We may be unnecessarily traversing up if we are hitting other errors than missing files. Let's add a IsNotExist guard. Any other error and we stop walking up the tree.

structured_type_arrow_batches_test.go
```diff
@@ -18,6 +23,48 @@ import (
 	"github.com/snowflakedb/gosnowflake/v2/arrowbatches"
 )
 
+func arrowTestRepoRoot(t *testing.T) string {
```
sfc-gh-boler, line 26: Let's extract this into a utility test function (along with the copy in repoRoot)

structured_type_arrow_batches_test.go
```diff
@@ -18,6 +23,48 @@ import (
...
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			t.Fatal("could not find repository root (no go.mod found)")
+		}
+		dir = parent
+	}
+}
+
+func arrowTestReadPrivateKey(t *testing.T, path string) *rsa.PrivateKey {
```
sfc-gh-boler, line 47: this one should be extracted too

structured_type_arrow_batches_test.go
sfc-gh-pfus: You mean to a common package? It is not possible in the current architecture, because it needs `Config` from the main package. My suggestion is leave it as is for now and in a separate commit I can create an internal `config` package, move there some stuff (dsn.go, client_configuration.go, tls_config.go), reexport for backward compatibility. But then I could create an internal `testconn` package that could contain all such test connection helpers there. WDYT?

sfc-gh-boler: Sorry, I suggested to move only repoRoot and readPrivateKey functions to a common internal package. They don't have dependencies on Config. I agree that it would be nice to see what we can do about testConn, but I agree that's a separate change.

sfc-gh-boler: (Resolving since you are already working on the refactor)

sfc-gh-boler (approved):
lgtm

PR #1712, opened 2026-02-25
SNOW-3161003 Fix isPrefixEqual to actually compare the second URL's port

sfc-gh-boler:
isPrefixEqual had a copy-paste bug where both p1 and p2 were derived from u1, meaning the port of u2 was never checked. This made the port validation in SAML authentication a no-op — URLs on the same hostname but different ports would incorrectly pass the prefix check.

Fix by reading p2 from u2 instead of u1, and add table-driven unit tests covering port mismatch scenarios that would have passed before the fix.

### Checklist

### DIFF
--- a/authokta_test.go
+++ b/authokta_test.go
@@ -36,6 +36,37 @@ func TestUnitPostBackURL(t *testing.T) {
 	}
 }
 
+func TestUnitIsPrefixEqual(t *testing.T) {
+	mustParse := func(raw string) *url.URL {
+		u, err := url.Parse(raw)
+		assertNilF(t, err, "parsing URL: "+raw)
+		return u
+	}
+	tests := []struct {
+		name     string
+		u1       string
+		u2       string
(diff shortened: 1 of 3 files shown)


authokta_test.go
```diff
@@ -36,6 +36,37 @@ func TestUnitPostBackURL(t *testing.T) {
 	}
 }
 
+func TestUnitIsPrefixEqual(t *testing.T) {
+	mustParse := func(raw string) *url.URL {
```
sfc-gh-pfus, line 40: For me this function may be a bit overkill, we could just inline it in the relevant t.Run, but I'm fine with keeping it too.

sfc-gh-boler: It's that just that we have to call it twice... and make sure we don't introduce a copy paste bug there. It makes sense it keep it.

sfc-gh-pbulawa (approved):

PR #1715, opened 2026-02-26
SNOW-3166926 Fix tokenFilePath DSN validation false positive

sfc-gh-boler:
The tokenFilePath DSN parameter was unusable due to a premature token file read during DSN parsing that triggered a false validation error.

When parsing the tokenFilePath DSN parameter, the code immediately read the token file and populated BOTH cfg.TokenFilePath and cfg.Token. Later validation checked if both fields were set and rejected the configuration with "token and tokenFilePath cannot be specified at the same time", even though the user only specified tokenFilePath in the DSN.

This PR removes the premature token file read during DSN parsing. In addition, remove redundant validation in parseDSNParams.

### Checklist

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -176,6 +176,8 @@ func (c *Config) getToken() (string, error) {
 	return c.Token, nil
 }
 
+var errTokenConfigConflict = errors.New("token and tokenFilePath cannot be specified at the same time")
+
 // Validate enables testing if config is correct.
 // A driver client may call it manually, but it is also called during opening first connection.
 func (c *Config) Validate() error {
@@ -188,7 +190,7 @@ func (c *Config) Validate() error {
 		return errors.New("WorkloadIdentityImpersonationPath is not supported for Azure")
 	}
 	if c.Token != "" && c.TokenFilePath != "" {
-		return errors.New("token and tokenFilePath cannot be specified at the same time")
+		return errTokenConfigConflict
 	}
 	return nil
 }
@@ -1013,10 +1015,6 @@ func parseDSNParams(cfg *Config, params string) (err error) {
 			cfg.Token = value
 		case "tokenFilePath":
 			cfg.TokenFilePath = value
-			cfg.Token, err = readToken(value)
-			if err != nil {
-				return
-			}
 		case "tlsConfigName":
 			cfg.TLSConfigName = value
 		case "workloadIdentityProvider":
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -1172,7 +1172,6 @@ func TestParseDSN(t *testing.T) {
 				CloudStorageTimeout:       defaultCloudStorageTimeout,
 				IncludeRetryReason:        ConfigBoolTrue,
 				DisableSamlURLCheck:       ConfigBoolFalse,
-				Token:                     "mock_token123456",
 				TokenFilePath:             "test_data/snowflake/session/token",
 			},
 			ocspMode: ocspModeFailOpen,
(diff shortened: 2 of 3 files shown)


dsn.go
```diff
@@ -197,9 +197,6 @@ func (c *Config) Validate() error {
 	if strings.EqualFold(c.WorkloadIdentityProvider, "azure") && len(c.WorkloadIdentityImpersonationPath) > 0 {
 		return errors.New("WorkloadIdentityImpersonationPath is not supported for Azure")
 	}
-	if c.Token != "" && c.TokenFilePath != "" {
```
sfc-gh-pfus, line 190: Why do you think it is not needed? This is a safeguard for someone providing both `token` and `tokenFilePath` in their config/DSN at the same time. Have you checked if other drivers have the similar check?

dsn_test.go
```diff
@@ -2479,3 +2478,47 @@ func TestDSNParsingWithTLSConfig(t *testing.T) {
 		})
 	}
 }
+
+// TestTokenFilePathValidationBug reproduces SNOW-3166926 where using tokenFilePath
+// in DSN triggers a false positive validation error claiming both token and
+// tokenFilePath are specified, when only tokenFilePath was actually provided.
+func TestTokenFilePathValidationBug(t *testing.T) {
```
sfc-gh-pfus: Do we really need this test? I think that the one you just fixed above is good enough.

dsn.go
sfc-gh-boler: Uh, there was a duplicate in parseDSNParams. But looks like Validate() will trigger on both paths, so I'm restoring this check

dsn_test.go
sfc-gh-boler: We don't have any tests for Validate.

dsn_test.go
```diff
@@ -2479,3 +2478,26 @@ func TestDSNParsingWithTLSConfig(t *testing.T) {
...
+	cfg := &Config{
+		Account:       "a",
+		User:          "u",
+		Password:      "p",
+		Token:         "direct-token",
+		TokenFilePath: "test_data/snowflake/session/token",
+	}
+	err := cfg.Validate()
+	if err == nil {
+		t.Error("Expected validation error when both Token and TokenFilePath are set")
```
sfc-gh-pfus: `assertNotNilE`?

dsn_test.go
```diff
@@ -2479,3 +2478,26 @@ func TestDSNParsingWithTLSConfig(t *testing.T) {
...
+	}
+
+	// Only Token — should pass
+	cfg.TokenFilePath = ""
+	assertNilE(t, cfg.Validate())
+
+	// Only TokenFilePath — should pass
+	cfg.Token = ""
+	cfg.TokenFilePath = "test_data/snowflake/session/token"
+	assertNilE(t, cfg.Validate())
```
sfc-gh-pfus: Maybe worth adding a case, that `Validate` returns nil for neither of the two is set?

sfc-gh-pfus (approved):

dsn_test.go
```diff
@@ -2479,3 +2478,26 @@ func TestDSNParsingWithTLSConfig(t *testing.T) {
...
+func TestTokenAndTokenFilePathValidation(t *testing.T) {
+	cfg := &Config{
+		Account:       "a",
+		User:          "u",
+		Password:      "p",
+		Token:         "direct-token",
+		TokenFilePath: "test_data/snowflake/session/token",
+	}
+	err := cfg.Validate()
+	if err == nil {
```
sfc-gh-pfus: It may be worth checking this error to ensure that the error was returned from the expected branch.

dsn_test.go
sfc-gh-boler: I don't see the point of those wrappers, but consistency is fine I guess.

dsn_test.go
sfc-gh-boler: We are looking for lack of error here. No matter where it comes from, it's not wanted.

dsn_test.go
sfc-gh-pfus: Yeah, me too 😂

dsn.go
sfc-gh-pfus, line 893: Just to double check - it is tested in `Validate` and it is just not needed here, right?

sfc-gh-boler: That's right. The Config produced by ParseDSN is passed to OpenWithConfig which calls Validate.

sfc-gh-boler:
### Merge activity

* **Mar 2, 1:13 PM UTC**: @sfc-gh-boler merged this pull request with [Graphite](https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1715).

PR #1716, opened 2026-02-26
SNOW-3169547 Update Jenkins job to use main branch instead of bptp-stable

sfc-gh-boler:
Changed svn_revision parameter from 'bptp-stable' to 'main' to match the updated Jenkins job configuration.

This prevents git checkout failures when the refspec doesn't include tags and ensures the job uses the main branch of snowflake-eng/snowflake for accessing test infrastructure.

Depends on: jenkins_utils/25056


sfc-gh-boler:
Closing, because this is a larger change that our SDET team is working to address.

PR #1717, opened 2026-02-26
Add headless SSO authentication

Margaev:
### Description

Make headless SSO authentication possible to enable SSO when running in the container, similar to how [snowflake-connector-python](https://github.com/snowflakedb/snowflake-connector-python) does it by prompting the user to open the URL and paste back the redirected URL with the token.

### Checklist

### DIFF
--- a/authexternalbrowser.go
+++ b/authexternalbrowser.go
@@ -1,6 +1,7 @@
 package gosnowflake
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/base64"
@@ -12,6 +13,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -29,6 +31,8 @@ You can close this window now and go back where you started from.
 
 	bufSize = 8192
 )
+
+var ErrBrowserOpen = errors.New("browser open error")
 
 // Builds a response to show to the user after successfully
 // getting a response from Snowflake.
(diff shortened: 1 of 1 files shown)


Margaev:

authexternalbrowser.go
```diff
@@ -99,6 +101,37 @@ func openBrowser(browserURL string) error {
 	return nil
 }
 
+func urlToRequestLine(input string) string {
+	u, err := url.Parse(input)
+	if err != nil {
+		return ""
```
sfc-gh-pfus: I don't think that returning an empty string is the best option here. It silently drops the error, but it is unusable later - we will try parsing HTTP on an empty line, which will obviously fail, but we won't know the original cause. Maybe we could propagate the error later?

authexternalbrowser.go
```diff
@@ -99,6 +101,37 @@ func openBrowser(browserURL string) error {
...
+	}
+	token := u.Query().Get("token")
+	if token == "" {
+		return ""
+	}
+	return "GET /?token=" + token + " HTTP/1.1"
+}
+
+func promptForRedirectedURL(browserURL string) ([]byte, error) {
```
sfc-gh-pfus: This name suggests it prompts only, but it also converts the URL into a request line and extracts a token.

authexternalbrowser.go
```diff
@@ -99,6 +101,37 @@ func openBrowser(browserURL string) error {
...
+	fmt.Println("We were unable to open a browser window for you, please open the url above manually then paste the URL you are redirected to into the terminal.")
+
+	reader := bufio.NewReader(os.Stdin)
+	input, err := reader.ReadString('\n')
+	if err != nil {
+		return nil, fmt.Errorf("Error reading input: %v", err)
+	}
+
+	requestLine := urlToRequestLine(strings.TrimSpace(input))
+	token, err := getTokenFromResponse(requestLine)
```
sfc-gh-pfus: I'm wondering, why we want to convert to a request line just to extract the token from it again? We already have easily parsable URL, maybe we could just extract a token from `input`?

authexternalbrowser.go
```diff
@@ -265,7 +298,12 @@ func doAuthenticateByExternalBrowser(ctx context.Context, sr *snowflakeRestful,
 	}
 
 	if err = defaultSamlResponseProvider().run(loginURL); err != nil {
-		return authenticateByExternalBrowserResult{nil, nil, err}
+		logger.WithContext(ctx).Infoln("falling back to headless SAML response provider")
+		headlessProvider := &headlessSamlResponseProvider{}
+		if err = headlessProvider.run(loginURL); err != nil {
+			return authenticateByExternalBrowserResult{nil, nil, err}
```
sfc-gh-pfus: We are losing and not logging anywhere the original `err` from `defaultSamlResponseProvider`. I think we should at least log it, or maybe return a wrapped error which contains errors for both cases?

authexternalbrowser.go
```diff
@@ -265,7 +298,12 @@ func doAuthenticateByExternalBrowser(ctx context.Context, sr *snowflakeRestful,
 	}
 
 	if err = defaultSamlResponseProvider().run(loginURL); err != nil {
-		return authenticateByExternalBrowserResult{nil, nil, err}
+		logger.WithContext(ctx).Infoln("falling back to headless SAML response provider")
```
sfc-gh-pfus: We'll fallback to headless in all cases, which is too wide (for instance what if the URL itself is not correct)? Maybe this is correct, maybe not. Do you think you can narrow errors to those related to opening a browser?

sfc-gh-pfus:
Hi @Margaev do you still think you need this functionality?

Margaev:
> Hi @Margaev do you still think you need this functionality?

Hi @sfc-gh-pfus, yes, will update the PR shortly

PR #1768, opened 2026-04-08
SNOW-3349897 Fix incorrect source logger when logged without WithContext

sfc-gh-pfus:


sfc-gh-pfus:
* **#1768** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1768?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1768?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Upcoming release
 
+Bug fixes:
+
+- Fix incorrect source location when logged without `LogEntry` (i.e. without `WithContext`) (snowflakedb/gosnowflake#1768).
```
sfc-gh-boler: With a clearer subject:

> - Fix logger reporting incorrect source location when called without `WithContext` (snowflakedb/gosnowflake#1768).

See the other comment about the performance impact. If it's not negligible, we should mention it here too.

internal/logger/slog_logger.go
```diff
@@ -12,13 +11,58 @@ import (
...
 )
 
 // formatSource formats caller information for logging
 func formatSource(frame *runtime.Frame) (string, string) {
 	return path.Base(frame.Function), fmt.Sprintf("%s:%d", path.Base(frame.File), frame.Line)
 }
 
+// findActualCaller walks the call stack to find the first caller outside internal/logger package.
+// This allows the logger to report the correct source location regardless of wrapper layers.
+func findActualCaller() uintptr {
```
sfc-gh-boler, line 25: Have you considered the performance impact of this? Each log entry how becomes somewhat more expensive. It may be useful to quantify this with a small benchmark. 

If it turns out to be a large overhead, the best course of action could be to move masking and filtering into the slog.Handler itself, so that we have predictable call depth.

sfc-gh-pfus: Yes. The impact is quite significant - 3x time. However, thanks to level filtering in the upper loggger layer, this happens only when a client actually enables a specific level. We don't log much on info, only things related to the query overall, not particular rows. So the actual slowdown is ~0.03ms per query. I think this is negligible having in mind that the network time even for the simplest queries is 100-200ms.

sfc-gh-boler (approved):
lgtm

PR #1777, opened 2026-04-17
SNOW-3328211: Deflake CI tests and retry on truncated query response

sfc-gh-boler:
## Summary
Follow-up to #1765. Fixes the root cause of intermittent LOB test failures and deflakes two other tests.

### Production fix: retry on truncated query response (`restful.go`)
When a large query result (e.g. 64MB LOB) is returned inline in the HTTP response, `json.NewDecoder(resp.Body).Decode()` can fail with `io.ErrUnexpectedEOF` if the ~100MB response body is truncated mid-stream by a network hiccup. Previously this error was returned directly to the caller with no retry. Now the driver:

1. Closes the truncated response body to release resources immediately
2. Re-issues the query via `FuncPostQuery` with the same request ID, so Snowflake returns the cached result without re-executing
3. Bounds the retry to **one attempt** via a context key (`truncatedResponseRetry`) to prevent infinite recursion on persistently truncated responses

This follows the existing pattern used for session-expired retries at line 260.

### Test fixes
- **TestMultipleAsyncQueries** (`async_test.go`): add 3-minute timeout with `time.NewTimer` and buffered channels to prevent indefinite hang and goroutine leak when timing-dependent assertion flakes (seen on Windows and FIPS mode).
- **TestExpiredCertificate** and **TestSelfSignedCertificate** (`driver_ocsp_test.go`): removed. These tested Go stdlib `crypto/x509` certificate validation, not driver logic, and depended on unreliable external `badssl.com` endpoints. `TestSelfSignedCertificate` was already disabled in a comment block.

### Changelog
Added entry under "Upcoming release" in `CHANGELOG.md`.

## Test plan
- [x] `TestUnitPostQueryHelperRetriesOnTruncatedResponse` — validates retry path and body closure
- [x] `TestUnitPostQueryHelperDoesNotRetryTruncatedResponseTwice` — validates retry is bounded to one attempt
- [ ] `TestLOBRetrievalWithArrow` and `TestLOBRetrievalWithJSON` pass without test-level retries
- [ ] `TestMultipleAsyncQueries` passes; timeout only fires on real flake
- [ ] Full CI passes with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bindings_test.go
+++ b/bindings_test.go
@@ -1435,10 +1435,10 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
 			dbt.mustExec(forceJSON)
 		}
 
-		var res string
 		testSizes := [2]int{smallSize, largeSize}
 		for _, testSize := range testSizes {
 			t.Run(fmt.Sprintf("testLOB_%v_useArrowFormat=%v", strconv.Itoa(testSize), strconv.FormatBool(useArrowFormat)), func(t *testing.T) {
+				var res string
 				rows, err := dbt.query(fmt.Sprintf("SELECT randstr(%v, 124)", testSize))
 				assertNilF(t, err)
 				defer func() {
--- a/bindings_test.go
+++ b/bindings_test.go
@@ -1434,19 +1434,36 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
 		testSizes := [2]int{smallSize, largeSize}
 		for _, testSize := range testSizes {
 			t.Run(fmt.Sprintf("testLOB_%v_useArrowFormat=%v", strconv.Itoa(testSize), strconv.FormatBool(useArrowFormat)), func(t *testing.T) {
-				rows, err := dbt.query(fmt.Sprintf("SELECT randstr(%v, 124)", testSize))
-				assertNilF(t, err)
-				defer func() {
+				const maxRetries = 3
+				for attempt := 1; attempt <= maxRetries; attempt++ {
+					rows, err := dbt.query(fmt.Sprintf("SELECT randstr(%v, 124)", testSize))
+					if err != nil {
+						if attempt < maxRetries && strings.Contains(err.Error(), "unexpected EOF") {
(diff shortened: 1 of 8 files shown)


bindings_test.go
```diff
@@ -1434,19 +1434,36 @@ func testLOBRetrieval(t *testing.T, useArrowFormat bool) {
...
+					if err != nil {
+						rows.Close()
+						if attempt < maxRetries && strings.Contains(err.Error(), "unexpected EOF") {
+							t.Logf("attempt %d/%d failed with transient error during scan: %v, retrying...", attempt, maxRetries, err)
+							time.Sleep(5 * time.Second)
+							continue
+						}
+						assertNilF(t, err)
+					}
+					assertNilF(t, rows.Close())
```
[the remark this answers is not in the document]
sfc-gh-pfus: Even worse - `sql.Rows` which are not closed prohibit `sql.DB` to close and wait infinitely.

driver_ocsp_test.go
```diff
@@ -531,32 +531,45 @@ func TestExpiredCertificate(t *testing.T) {
...
-		t.Fatalf("should fail to ping. %v", testURL)
-	}
-	urlErr, ok := err.(*url.Error)
-	if !ok {
-		t.Fatalf("failed to extract error URL Error: %v", err)
-	}
-	_, ok = urlErr.Err.(x509.CertificateInvalidError)
-
-	if !ok {
+	const maxRetries = 3
```
sfc-gh-pfus, line 539: I don't think we need retries. Maybe we can just accept the error that sometimes happen as an "expected" value, similarly to what we had before (expecting `timeout` or `connection attempt failed`? I don't see a point in making this logic complex with retries.
Eventually, it would be great to test it on wiremock. Or maybe just remove these tests, since this is implemented by golang itself, not our code.

sfc-gh-boler: Good catch. Removed this test and another comment out test.

sfc-gh-pfus (approved):

PR #1788, opened 2026-05-05
SNOW-3474718: don't drop legal files having names with trailing dot from PUT

sfc-gh-dszmolka:
### Description

This came from a Snowsight use-case, where the team noticed that files like  `Untitled 1.` got confirmed on the GUI to be uploaded, but actually wasn't `PUT` to the stage. 

Turns out [the validation we have in place](https://github.com/snowflakedb/gosnowflake/blob/v2.0.2/file_util.go#L209-L210) for years just silently drops even legal file names when they end with `.` char. 

This change aims  to address this issue and allow files legally ending with `.` to be uploaded to the stage.
Also added new tests. 

Manual tests done:
* upload `/path/to/Untitled 1.` to user stage, then list out user stage to confirm

### DIFF
--- a/file_util_test.go
+++ b/file_util_test.go
@@ -3,6 +3,7 @@ package gosnowflake
 import (
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


file_util_test.go
```diff
@@ -30,14 +31,45 @@ func TestBaseName(t *testing.T) {
 		{"/tmp", "tmp"},
 		{"/home/desktop/.", ""},
 		{"/home/desktop/..", ""},
+		{".", ""},
+		{"..", ""},
+		{"/", ""},
+		{"/home/desktop/", "desktop"},
```
sfc-gh-boler: Previous implementation would return "" in this case, it "desktop". While this will likely get caught later, it would be good to keep the old behaviour.

file_util_test.go
```diff
@@ -30,14 +31,45 @@ func TestBaseName(t *testing.T) {
...
 	}
 
 	for _, test := range testcases {
 		t.Run(test.in, func(t *testing.T) {
-			base := baseName(test.in)
-			if test.out != base {
-				t.Errorf("Failed to get base, input %v, expected: %v, got: %v", test.in, test.out, base)
-			}
+			actual := baseName(test.in)
+			assertEqualE(t, actual, test.out, "baseName(%q)", test.in)
```
sfc-gh-boler: assertEqualE doesn't take description as a formatted string, but as a list of space-separated strings. So the %q interpolation will never happen. Use: ```"baseName:", test.in)```

(This is one of may reasons why I don't like assert libraries: they are a DSL and full of traps. But we have what we have.)

file_util_test.go
```diff
@@ -30,14 +31,45 @@ func TestBaseName(t *testing.T) {
...
+		{`C:\Users\trailing-dot.txt.`, "trailing-dot.txt."},
+		{`C:\path\to\Untitled 1.`, "Untitled 1."},
+		{`C:\path\to\.`, ""},
+		{`C:\path\to\..`, ""},
+	}
+
+	for _, test := range testcases {
+		t.Run(test.in, func(t *testing.T) {
+			actual := baseName(test.in)
+			assertEqualE(t, actual, test.out, "baseName(%q)", test.in)
```
sfc-gh-boler: ditto

sfc-gh-boler (changes requested):

file_util_test.go
sfc-gh-dszmolka: ok, done

file_util_test.go
sfc-gh-dszmolka: ok, done

file_util_test.go
sfc-gh-dszmolka: while the `filepath.Base` returns `desktop` in this case but indeed we had a custom implementation of it which rejcted the path ending with `/`  (with absolutely no unit tests or explanation around it, which could have given us a clue **why** it was implmented in this custom way)

 but now also checked the Python driver now to compare and [it also has](https://github.com/snowflakedb/snowflake-connector-python/blob/v4.4.0/src/snowflake/connector/file_util.py#L76C9-L76C48) `os.path.basename` which for `/home/desktop/` would return empty string, i reinstated this part of the original logic (reject if the path ends with `/`
 
 thank you @sfc-gh-boler for catching this!

sfc-gh-boler (approved):
lgtm

PR #1790, opened 2026-05-07
SNOW-3176172: signed minicore for Windows and Mac

sfc-gh-dszmolka:
### Description

Similarly to the [node.js driver](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1401), sign the Windows and Mac minicore binaries here too. 

**Note**: the binaries themselves are the same as in previous versions; only difference is that they are now signed.


sfc-gh-boler (approved):
lgtm

PR #1787, opened 2026-05-04
minicore: skip -ldl on OpenBSD

atanas-vladimirov:
### Description

OpenBSD does not provide a separate libdl, so unconditional -ldl breaks linking. dlopen/dlsym are available via libc on OpenBSD, so this keeps POSIX loading portable while preserving behavior on other platforms.

### Checklist
- [ ] Added proper logging (if possible)
- [ ] Created tests which fail without the change (if possible)
- [ ] Extended the README / documentation, if necessary

PR #1791, opened 2026-05-12
minicore: skip -ldl on OpenBSD

sfc-gh-boler:
Mirror of #1787 by @atanas-vladimirov, re-pushed to a branch on this repo so the secrets-gated CI jobs in `build-test.yml` actually receive `PARAMETERS_SECRET` / `GOLANG_PRIVATE_KEY_SECRET`. Fork PRs get empty values for those, which makes `gpg` fail with "Bad session key" in `ci/scripts/setup_connection_parameters.sh` and blocks every integration test.

Same head commit as #1787 (`b70fc77`). All credit for the change goes to the original author; #1787 will be closed once this lands.

---

### Description (from #1787)

OpenBSD does not provide a separate libdl, so unconditional `-ldl` breaks linking. `dlopen`/`dlsym` are available via libc on OpenBSD, so this keeps POSIX loading portable while preserving behavior on other platforms.

Co-authored-by: Atanas Vladimirov <vladimirov.atanas@gmail.com>


sfc-gh-boler:
This PR is needed, because merging requires certain tests which don't run against PRs from forks. I'll merge this one and close the original PR. Authorship will be preserved.

sfc-gh-dheyman (approved):

Issue #1789, opened 2026-05-06
SNOW-3479780: S3 chunk download stall causes silent result truncation and permanent goroutine leak


frbvianna:
1. What version of GO driver are you using?                                                                                                                              
                                                                                                                                                                           
  github.com/snowflakedb/gosnowflake v1.19.1                                                                                                                               
                                                                                                                                                                           
  2. What operating system and processor architecture are you using?                                                                                                       
                                                                                                                                                                           
  Linux/amd64 (production deployment in Kubernetes)                                                                                                                        
                  
  3. What version of GO are you using?                                                                                                                                     
                  
  go1.26.1                                                                                                                                                                 
                  
  4. Server version:                                                                                                                                                       
   
 10.16.101                                                                                                                       
                  
  5. What did you do?                                                                                                                                                      
                  
  Executed queries against SNOWFLAKE_SAMPLE_DATA.TPCH_SF1000.PARTSUPP, which is publicly available in any Snowflake account and can be used to reproduce the issue at      
  scale. Each query retrieves a sub-range of the table up to ~7.2 GB, not the full table in a single query. Result chunks are fetched from S3 in background goroutines. We
  consumed the result set by calling RecordReader.Next() in a loop (via the ADBC Snowflake driver, which uses gosnowflake underneath).                                     
                  
  During the fetch, one S3 download stalled — Next() blocked for approximately 10 minutes with no error. When it unblocked, only a few more batches were returned before   
  Next() returned false with Err() == nil.
                                                                                                                                                                           
  Additionally, after the query completed and the reader was released (cancelFn() called), goroutines that had been blocked in Next() on the last result channel remained  
  permanently stuck. Inspecting pprof showed them blocked at the channel receive inside reader.Next() indefinitely:
                                                                                                                                                                           
  9 goroutines blocked at:
      github.com/snowflakedb/gosnowflake.(*snowflakeRows).Next                                                                                                             
          (or equivalent internal channel receive)                                                                                                                         
                                                                                                                                                                           
  The root cause appears to be that GetStream — the HTTP call that fetches each result batch from S3 — does not propagate context cancellation to the underlying HTTP      
  request promptly. When cancelFn() is called, download goroutines that are blocked inside GetStream do not exit. As a result, group.Wait() never returns,                 
  close(chs[lastChannelIndex]) never executes, and any goroutine waiting on that channel leaks permanently.                                                                
                  
  6. What did you expect to see?                                                                                                                                           
   
[...]

PR #1795, opened 2026-05-12
SNOW-3479780: unblock canceled Arrow chunk reads

sfc-gh-boler:
SNOW-3479780 / issue #1789 report that canceling an Arrow stream after the chunk response is already open can leave the read path stalled until the client timeout because nothing interrupts the live body read. In practice that also leaves the caller-facing read goroutine stuck and surfaces transport-specific close behavior instead of a clean `context.Canceled`.

This change wraps opened Arrow chunk streams in a cancel-aware reader that interrupts the raw response body when `ctx.Done()` fires, normalizes interrupted terminal read errors to `ctx.Err()`, and still preserves true successful completion and EOF semantics, including the final `n > 0, io.EOF` case. The tests now cover stalled-read unblock, non-EOF close-error normalization, already-canceled `GetStream()` behavior, close-then-cancel watcher shutdown, and completed plain and gzip-backed streams. The default streaming model is unchanged, and Reset/retry integration stays out of scope.

### Checklist

### DIFF
--- a/arrow_stream.go
+++ b/arrow_stream.go
@@ -11,6 +11,8 @@ import (
 	"maps"
 	"net/http"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/apache/arrow-go/v18/arrow/ipc"
@@ -76,7 +78,7 @@ func (asb *ArrowStreamBatch) GetStream(ctx context.Context) (io.ReadCloser, erro
 			return nil, err
 		}
 	}
-	return asb.rr, nil
+	return newCancelableStream(ctx, asb.rr), nil
 }
 
 // streamWrapReader wraps an io.Reader so that Close closes the underlying body.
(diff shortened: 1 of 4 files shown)


arrow_stream.go
```diff
@@ -94,6 +96,118 @@ func (w *streamWrapReader) Close() error {
...
+			return n, normalized
+		}
+		r.completed.Store(true)
+	} else if normalized := r.normalizeInterruptedReadErr(err); normalized != nil {
+		return n, normalized
+	}
+	return n, err
+}
+
+func (r *cancelableStream) normalizeInterruptedReadErr(err error) error {
```
sfc-gh-dheyman: I think we should only rewrite the error if we actually interrupted

sfc-gh-dheyman (approved):
one comment, but overall LGTM

arrow_stream.go
sfc-gh-boler: We will overwrite the error only if an interrupt happens, otherwise we will return the error from `r.inner.Read()`.

I've have refactored this a bit make it clearer.

PR #1797, opened 2026-05-18
SNOW-3351450: emit client_connection_identifier_shape in-band telemetry

sfc-gh-boler:
The Traffic team needs to understand how customers actually
supply connection-identifier inputs (`Account`, `Region`, `Host`) ahead of
the Northstar migration. This PR adds a single in-band telemetry event,
`client_connection_identifier_shape`, that captures which of those three
fields the user explicitly set and whether the raw account string carried
the dot ("account.region") or dash ("org-account") markers. The shape is
captured at the earliest parsing stage across DSN (`ParseDSN`), TOML
(`LoadConnectionConfig`), and programmatic `Config` (via
`InferInputShapeIfMissing` called from `Connector.Connect` and
`OpenWithConfig`), then emitted once at post-login time alongside the
existing `connectionTelemetry` hook. Provenance is recorded before
`FillMissingConfigParameters` mutates the `Config` so inferred values are
never misclassified as user-supplied.

The implementation mirrors how `snowflake-connector-python`
(`_log_minicore_import`) and `snowflake-jdbc` (`sendMinicoreTelemetry`)
route their analogous "minicore usage" telemetry: through
`sc.telemetry.addLog` to `/telemetry/send`, gated by the server-supplied
`CLIENT_TELEMETRY_ENABLED` session parameter that arrives in the login
response. A local `SF_TELEMETRY_DISABLE_CONNECTION_SHAPE` env-var kill
switch is wired in addition. The `Config.inputShape` field is unexported
and accessed across packages via `internal/config.InputShapeOf` (read)
and `internal/config.SetInputShape` (test setup), so the public
`gosnowflake.Config` API surface is unchanged.

This is the reference implementation; equivalents for
`snowflake-connector-python`, `snowflake-connector-nodejs`, and
`snowflake-jdbc` will land separately. The data collection is
time-boxed — removal is tracked in SNOW-3548350 with a target date of
2026-11-30.

### Checklist
- [x] Added proper logging (if possible)

### DIFF
--- a/connection_identifier_shape_telemetry_test.go
+++ b/connection_identifier_shape_telemetry_test.go
@@ -1,0 +1,114 @@
+package gosnowflake
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	sfconfig "github.com/snowflakedb/gosnowflake/v2/internal/config"
+)
+
+// newTestShapeConn returns a snowflakeConn with an enabled in-memory telemetry
+// client suitable for unit-testing connectionIdentifierShapeTelemetry without
+// requiring a real Snowflake account or HTTP plumbing. The returned conn's
+// telemetry.sr is nil, which is safe because the tests only exercise addLog
+// (which appends to st.logs without touching st.sr).
+func newTestShapeConn() *snowflakeConn {
+	return &snowflakeConn{
+		ctx: context.Background(),
+		telemetry: &snowflakeTelemetry{
+			mutex:     &sync.Mutex{},
+			enabled:   true,
+			flushSize: defaultFlushSize,
+		},
+	}
+}
+
+func TestConnectionIdentifierShapeTelemetryEmitsExpectedPayload(t *testing.T) {
+	t.Setenv(disableConnectionShapeEnv, "")
+	cfg := &Config{}
+	sfconfig.SetInputShape(cfg, &sfconfig.ConnectionIdentifierShape{
+		AccountProvided:    true,
+		AccountWithRegion:  true,
+		AccountOrgProvided: true,
+		RegionProvided:     false,
+		HostProvided:       false,
+	})
+	sc := newTestShapeConn()
+	sc.connectionIdentifierShapeTelemetry(cfg)
+
+	assertEqualF(t, len(sc.telemetry.logs), 1, "expected exactly one telemetry record")
+	msg := sc.telemetry.logs[0].Message
+	assertEqualE(t, msg[typeKey], connectionIdentifierShape)
+	assertEqualE(t, msg[sourceKey], telemetrySource)
+	assertEqualE(t, msg[driverTypeKey], "Go")
+	assertEqualE(t, msg[accountProvidedKey], "true")
+	assertEqualE(t, msg[accountWithRegionKey], "true")
+	assertEqualE(t, msg[accountOrgProvidedKey], "true")
+	assertEqualE(t, msg[regionProvidedKey], "false")
+	assertEqualE(t, msg[hostProvidedKey], "false")
+	assertNotEqualE(t, msg[driverVersionKey], "",
+		"driver version field should carry SnowflakeGoDriverVersion")
+	assertNotEqualE(t, msg[golangVersionKey], "",
+		"golang version field should carry runtime.Version()")
+}
+
+func TestConnectionIdentifierShapeTelemetrySkipsWhenInputShapeNil(t *testing.T) {
+	t.Setenv(disableConnectionShapeEnv, "")
+	sc := newTestShapeConn()
+	// A freshly-constructed Config has no shape set (SetInputShape not called).
+	sc.connectionIdentifierShapeTelemetry(&Config{})
+	assertEmptyE(t, sc.telemetry.logs,
+		"expected no telemetry record when shape is nil")
+}
+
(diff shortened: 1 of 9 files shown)


connection_identifier_shape_telemetry_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if got := len(sc.telemetry.logs); got != 1 {
+		t.Fatalf("expected exactly one telemetry record, got %d", got)
+	}
+	msg := sc.telemetry.logs[0].Message
+
+	wantField := func(key, want string) {
+		t.Helper()
+		if got := msg[key]; got != want {
+			t.Errorf("telemetry field %q = %q, want %q", key, got, want)
```
sfc-gh-dheyman: instead of `t.Errorf `or `t.Fatalf` we should use project's `assertNilF`, `assertEqualF`, `assertTrueF` etc.

internal/config/dsn.go
```diff
@@ -304,12 +304,62 @@ func DSN(cfg *Config) (dsn string, err error) {
...
+		shape.AccountOrgProvided = true
+	}
+}
+
+// InferInputShapeIfMissing populates cfg.inputShape from the current Config
+// fields when shape capture has not already run (i.e. cfg was constructed
+// programmatically rather than via ParseDSN or LoadConnectionConfig). It is a
+// no-op when the shape is already non-nil.
+//
+// MUST be called before FillMissingConfigParameters mutates cfg, otherwise
```
sfc-gh-dheyman: nit: can this be enforced?

connection_util.go
```diff
@@ -95,6 +98,54 @@ func (sc *snowflakeConn) connectionTelemetry(cfg *Config) {
...
+	}
+	if err := sc.telemetry.addLog(data); err != nil {
+		logger.WithContext(sc.ctx).Warnf("cannot add connection-identifier-shape telemetry log: %v", err)
+	}
+}
+
+func connectionShapeTelemetryDisabledByEnv() bool {
+	v := strings.ToLower(strings.TrimSpace(os.Getenv(disableConnectionShapeEnv)))
+	switch v {
+	case "1", "true", "yes":
```
sfc-gh-dheyman: if we documented only `true` in changelog, I guess we can stick to that :)

internal/config/dsn.go
```diff
@@ -694,6 +748,17 @@ func parseAccountHostPort(cfg *Config, posAt, posSlash int, dsn string) (err err
...
 	cfg.Host = dsn[posAt+1 : k]
+	// The DSN carries a real host (not an account-form authority that
+	// transformAccountToHost will rewrite) in either of two cases:
+	//   1) the authority contains the Snowflake TLD, e.g. "myacct.snowflakecomputing.com", or
+	//   2) the authority carries an explicit port, e.g. "localhost:443" — the
+	//      transformAccountToHost guard only fires when Port == 0, so any
+	//      explicit port means we never rewrite Host out from under the user.
+	// Capture HostProvided for both cases so non-Snowflake-TLD targets
+	// (localhost, IP, private DNS) aren't silently misclassified as host-not-provided.
+	if cfg.inputShape != nil && (hostIncludesTopLevelDomain(cfg.Host) || cfg.Port != 0) {
```
sfc-gh-dheyman: shouldn't we add `cfg.Host != ""` to the condition?

sfc-gh-dheyman (changes requested):

connection_identifier_shape_telemetry_test.go
sfc-gh-boler: Switched to project helpers in `connection_identifier_shape_telemetry_test.go` and `internal/config/connection_identifier_shape_test.go` — now using `assertNilF` / `assertNotNilF` / `assertEqualF` / `assertEqualE` / `assertEmptyE` / `assertFalseE` etc. Fatal helpers (`*F`) on setup preconditions, error helpers (`*E`) on field comparisons.

internal/config/dsn.go
sfc-gh-boler: Yes — enforced. Shape capture for programmatic Configs is now done as the first step inside `FillMissingConfigParameters`, before any field-normalization runs. Callers (`Connector.Connect`, direct `ParseDSN` / `LoadConnectionConfig` callers) no longer need to remember the ordering invariant; the explicit `InferInputShapeIfMissing` calls in `connector.go` and `driver.go` are gone, and the function itself is now unexported (`inferInputShapeIfMissing`).

connection_util.go
sfc-gh-boler: Good catch — narrowed to `strings.EqualFold(..., "true")`, matching `SF_DISABLE_MINICORE`. The test now exercises `true` / `True` / `TRUE` on the kill-switch side and `1` / `yes` / `Yes` / `false` / `no` / `anything-else` / `""` on the off-by-default side (the former truthy aliases stop disabling, which is the intended behavior change).

internal/config/dsn.go
sfc-gh-boler: Done — added `cfg.Host != ""` to the guard. The degenerate `u:p@:8080/db` form would have wrongly set `HostProvided = true` despite `cfg.Host` being empty after the slice read. Comment above the condition spells out the guard now.

sfc-gh-dheyman (approved):

PR #1798, opened 2026-05-20
SNOW-3438406: adding new envvar to disable OCSP checks (SF_DISABLE_OCSP_CHECKS)

sfc-gh-dszmolka:
### Description

Introducing a new environment variable `SF_DISABLE_OCSP_CHECKS` to be able to toggle `Config.DisableOCSPChecks` from outside the Config/DSN. Main use-case is where it is hard or impossible to access this setting from the business application (e.g. Power BI) but it would be possible to access it from the environment. 

For troubleshooting, testing, or [following new industry trends](https://www.snowflake.com/en/blog/engineering/rethinking-certificate-revocation-snowflake/)  it's sometimes necessary to turn off OCSP checks which are (at the moment) enabled by default. It was always possible from the driver configuration, but sometimes it's not accessible. 

The new env var aims to address this: when set to `true`, it toggles [DisableOCSPChecks=true](https://github.com/snowflakedb/gosnowflake/blob/v2.0.2/doc.go#L109-L110) and bypasses the second layer of certificate checks with OCSP. The first layer (certificate verification during TLS handshake) still remains intact and cannot be disabled. 

The env var, even when set to `true`, **does not** take effect and an appropriate message is logged, when the driver is manually configured for [OCSP FAIL_CLOSE](https://docs.snowflake.com/en/user-guide/ocsp#fail-close) mode

This local driver configuration takes precedence over a wider env variable.

Added unit tests. 

### Checklist
- [x] Added proper logging (if possible)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 New features:
 
-- Added `SF_OCSP_DISABLE_CHECKS` environment variable to govern OCSP checks, same as `Config.DisableOCSPChecks` - by default, even before this change, OCSP checks are enabled (snowflakedb/gosnowflake#1798).
+- Added `SF_DISABLE_OCSP_CHECKS` environment variable as an additional option to override OCSP checks' default behaviour besides the existing `DisableOCSPChecks` config. (snowflakedb/gosnowflake#1798).
   - Note: the env var is explicitly refused when OCSP fail-closed mode is active.
 
 - Added `ArrowStreamBatch.Reset()` method that closes any existing stream and clears the cached reader, allowing callers to retry `GetStream` after a mid-stream failure (e.g. TCP RST) without re-executing the entire query. Inline (RowSetBase64) batches are restored from cached bytes on reset.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ New features:
 
 - Added `SF_OCSP_DISABLE_CHECKS` environment variable to govern OCSP checks, same as `Config.DisableOCSPChecks` - by default, even before this change, OCSP checks are enabled (snowflakedb/gosnowflake#1798).
   - the env var is ignored when the DSN or connection configuration sets `disableOCSPChecks` to any value
-  - is explicitly refused when OCSP fail-closed mode is active 
+  - is explicitly refused when OCSP fail-closed mode is active
 
 - Added `ArrowStreamBatch.Reset()` method that closes any existing stream and clears the cached reader, allowing callers to retry `GetStream` after a mid-stream failure (e.g. TCP RST) without re-executing the entire query. Inline (RowSetBase64) batches are restored from cached bytes on reset.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 New features:
 
-- Added `SF_OCSP_DISABLE_CHECKS` environment variable to govern OCSP checks, same as `Config.DisableOCSPChecks` - by default, even before this change, OCSP checks are enabled (snowflakedb/gosnowflake#XXXX).
+- Added `SF_OCSP_DISABLE_CHECKS` environment variable to govern OCSP checks, same as `Config.DisableOCSPChecks` - by default, even before this change, OCSP checks are enabled (snowflakedb/gosnowflake#1798).
   - the env var is ignored when the DSN or connection configuration sets `disableOCSPChecks` to any value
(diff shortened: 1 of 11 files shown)


ci/gofix.sh
```diff
@@ -46,6 +46,9 @@ discover_custom_tags() {
   )
 }
 
+# Reset tracked files so we only detect go fix changes.
```
sfc-gh-boler: If someone runs this script on their workstation, it will wipe out any changes which will be... surprising to put it mildly. 

Why is this needed?

internal/config/dsn.go
```diff
@@ -43,6 +43,9 @@ const (
 
 const clientType = "Go"
 
+// envVarDisableOCSPChecks is the environment variable name for disabling OCSP certificate revocation checks.
+const envVarDisableOCSPChecks = "SF_OCSP_DISABLE_CHECKS"
```
sfc-gh-boler: let's keep the name similar to the other vars (i.e. no namespacing), so: SF_DISABLE_OCSP_CHECKS.

internal/config/dsn.go
```diff
@@ -617,6 +620,17 @@ func FillMissingConfigParameters(cfg *Config) error {
 		cfg.OCSPFailOpen = OCSPFailOpenTrue
 	}
 
+	if val, _ := GetFromEnv(envVarDisableOCSPChecks, false); val != "" {
+		if v, err := strconv.ParseBool(val); err == nil && v {
+			if cfg.OCSPFailOpen == OCSPFailOpenFalse {
+				logger.Info("SF_OCSP_DISABLE_CHECKS environment variable is set, but OCSP fail-closed mode is active; the environment variable will be ignored")
```
sfc-gh-boler: Use the const instead of repeating the env var.

internal/config/dsn.go
```diff
@@ -617,6 +620,17 @@ func FillMissingConfigParameters(cfg *Config) error {
...
+			} else {
+				cfg.DisableOCSPChecks = true
+				logger.Info("Disabling OCSP checks from SF_OCSP_DISABLE_CHECKS environment variable")
```
sfc-gh-boler: ditto

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 New features:
 
+- Added `SF_OCSP_DISABLE_CHECKS` environment variable to govern OCSP checks, same as `Config.DisableOCSPChecks` - by default, even before this change, OCSP checks are enabled (snowflakedb/gosnowflake#1798).
```
sfc-gh-boler: Let's reword this a bit to explain that it doesn't fully govern the behaviour, but it's the most specific override.

sfc-gh-boler (changes requested):

ci/gofix.sh
sfc-gh-dszmolka: i was baffled too; that while on local, both commands from 
```
ensure_clean_worktree() {
  if ! git diff --quiet --ignore-submodules -- || \
     ! git diff --cached --quiet --ignore-submodules --; then
```

returned `0` locally but on the CI, it constantly failed with 
```
"ERROR: working tree is dirty before go fix runs."
"Run this check from a clean checkout so failures only reflect go fix changes."
```

meaning, one of the commands exited `1` for which i really found no reason. And unfortunately this step is required. 

Taking it out now, let's see how it works.

internal/config/dsn.go
sfc-gh-dszmolka: renamed to `SF_DISABLE_OCSP_CHECKS`

internal/config/dsn.go
sfc-gh-dszmolka: done

internal/config/dsn.go
sfc-gh-dszmolka: done

CHANGELOG.md
sfc-gh-dszmolka: reworded to something which emphasizes override
> - Added `SF_DISABLE_OCSP_CHECKS` environment variable as an additional option to override OCSP checks' default behaviour besides the existing `DisableOCSPChecks` config. (snowflakedb/gosnowflake#1798).
>  - Note: the env var is explicitly refused when OCSP fail-closed mode is active.

ci/gofix.sh
sfc-gh-dszmolka: now it passed  🤷

sfc-gh-boler: :D

sfc-gh-boler (approved):
lgtm

PR #1805, opened 2026-06-11
SNOW-3637978: Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages

sfc-gh-dszmolka:
### Description

Similar to https://github.com/snowflakedb/snowflake-connector-python/pull/2556. Since `s3:GetAccelerateConfiguration` permission is not granted on purpose, and won't be granted on Snowflake-internal buckets, 100% of the s3 client `GetBucketAccelerateConfiguration` calls are failing. This wastes resources as TLS still needs to be performed for something which we know not gonna work and apparently starts to cause operational issues as well in certain tools.

### Fix
Similarly to the Python driver; do not perform `GetBucketAccelerateConfiguration` call for Snowflake-internal buckets. For anything else, proceed as before.

Added unit test. Also confirmed against a live test account that accelerate probe was not launched when performing a real `PUT`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 New features:
 
 Bug fixes:
-- Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#XXXX).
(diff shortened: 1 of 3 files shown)


sfc-gh-boler (approved):
lgtm

PR #1801, opened 2026-05-21
Remove init() function

atoulme:
Remove init() function that leads to the creation of files, even when gosnowflake is just present as a dependency

### Description

Fix #1800 
SNOW-3561155 remove init function

### Checklist
- [x] Added proper logging (if possible)
- [x] Created tests which fail without the change (if possible)
- [x] Extended the README / documentation, if necessary

PR #1807, opened 2026-06-16
SNOW-3561155: load minicore lazily on driver use instead of package init

### LINKED ISSUE #1800: SNOW-3561155: gosnowflake writes a folder to /tmp even when not used
## What version of GO driver are you using?
v2.0.2
   
## What operating system and processor architecture are you using?
Linux
   
## What version of GO are you using?
1.25.9

## Server version:* E.g. 1.90.1
N/A

## What did you do?
Start the OpenTelemetry Collector Contrib executable. We use a configuration that doesn't use any components that are tied to snowflake.

## What did you expect to see?
No files written to /tmp 

## What did you see instead?
The collector starts and runs but writes to /tmp a folder named gosnowflake-cgo.

It looks like this is because the go package will run init() for gosnowflake, and eventually create a "minicore" singleton.

## Can you set logging to DEBUG and collect the logs?
N/A

sfc-gh-boler:
## Overview

`gosnowflake` writes a temp folder (`/tmp/gosnowflake-cgo*`) during package `init()` even when the driver is never used — e.g. when it's only a transitive dependency of a binary that doesn't talk to Snowflake (reported in #1800 by the otel-collector-contrib maintainers).

The `init()` eagerly started minicore loading, which writes the embedded native library to disk and `dlopen`s it (to report the core version in login telemetry). This side effect fired on import alone.

## Change

- Remove the package `init()` in `minicore.go`.
- Kick off the same async load lazily, when the app first references the driver:
  - `NewConnector` — `database/sql.Open()` calls this synchronously via `OpenConnector`, so the common path keeps the same head start.
  - `OpenWithConfig` — catch-all for `SnowflakeDriver.Open(dsn)` and direct callers.
- Both calls are idempotent (`sync.Once`); loading remains fully async, so nothing is added to the connection critical path.

Net effect: no filesystem work when the driver is never used, while real users keep minicore warm by first login.

## Relationship to #1801

#1801 also removes `init()` but does not re-trigger loading, so the first login of every process reports `CORE_VERSION` empty / `"Minicore is still loading"`. This PR re-triggers at driver-reference time, preserving first-login telemetry for real usage. Suggest closing #1801 in favor of this.

## Behavior note (intentional, not a regression)

Minicore loading is async and non-blocking by design (it must never impact customers). A process that references the driver and connects with no gap in between can still report `"still loading"` on its very first login — the same best-effort behavior `init()` had. We deliberately did not add a synchronous/blocking load.

## Testing

- `go build ./...` and `go build -tags minicore_disabled ./...`
- `go vet ./...`
- Minicore loader unit tests (real `dlopen`)
- `TestMiniCoreLoadedE2E` updated: since `init()` no longer pre-warms, it now waits for the load (`awaitMinicoreLoaded`) before connecting, so the login request deterministically carries `CORE_VERSION` with no `CORE_LOAD_ERROR`.
- Manual #1800 repro: an import-only program creates no `gosnowflake-cgo` temp activity; a program that calls `NewConnector` (without connecting) does trigger the load.

Fixes #1800.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ New features:
 
 Bug fixes:
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
+- Fixed gosnowflake writing a `gosnowflake-cgo` directory under the system temp dir at package import time even when the driver was never used (e.g. when imported only as a transitive dependency). Minicore now loads lazily when the driver is first referenced (`NewConnector`/`OpenWithConfig`) instead of in `init()` (snowflakedb/gosnowflake#1807).
 
 Internal changes:
 
--- a/connector.go
+++ b/connector.go
@@ -21,6 +21,12 @@ type Connector struct {
 
 // NewConnector creates a new connector with the given SnowflakeDriver and Config.
 func NewConnector(driver InternalSnowflakeDriver, config Config) driver.Connector {
+	// Begin loading minicore asynchronously as soon as the application
+	// references the driver. database/sql.Open() invokes OpenConnector ->
+	// NewConnector synchronously, so this gives the version a head start to be
+	// warm by first login, without loading at package-import time when
+	// gosnowflake is merely a transitive dependency (SNOW-3561155 / #1800).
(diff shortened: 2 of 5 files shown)


sfc-gh-dheyman (approved):
LGTM!

PR #1808, opened 2026-06-22
Add WithFileGetStreamForExactFile to fix get-stream prefix-sibling corruption

sfc-gh-jchernandez:
## Problem

A `GET` resolves its stage path by **server-side prefix matching**, so a request for a single
file (e.g. `foo`) can come back as several files (`foo` **and** `foobar`, a nested `foo/foo`, …).

The get-stream primitive (`WithFileGetStream`) has exactly one `io.Writer` and one shared
`streamBuffer`, but `download()` still fans out **one goroutine per matched file**, all writing
that *same* buffer (`storage_client.go`):

- **encrypted:** each goroutine `Truncate`s the shared buffer to *its own* file size → a
  nondeterministic concat/truncation of the siblings;
- **unencrypted:** each goroutine reassigns the shared buffer pointer (last-writer-wins),
  silently dropping the others.

Either way the caller can receive corrupt bytes, or an arbitrary sibling's content, **with no
error** — including when the requested file does not exist but a prefix-sibling does.

This continues the file-transfer hardening from #1687 (SNOW-3049473): that made a panicking
download goroutine non-fatal and added a single-file `totalFileSize` bound check, but neither
addresses the *cross-file* shared-buffer contention when a get-stream resolves to N > 1 files.

## Fix

Add an opt-in API:

```go
func WithFileGetStreamForExactFile(ctx context.Context, writer io.Writer, fileName string) context.Context
```

`parseCommand` selects the requested file and narrows `srcFiles` to it **before any download
goroutine is spawned**, structurally eliminating the shared-buffer hazard for that path (exactly
one goroutine ever touches `streamBuffer`). Selection runs *after* the source-file →
encryption-material map is built (the map is keyed by source file name), so a selected file at
index > 0 keeps its own encryption material.

Matching is **case-sensitive, by leaf** — the same identity the driver uses when downloading to a
directory (`baseName(dstFileName)`):

- a file matches when its path equals `fileName` or ends with `"/" + fileName` (so a leaf `foo`
  or a trailing path `dir/foo`);
- when several matches share the leaf, the **shallowest** path wins (the exact `foo`, not a
  deeper `foo/foo`) — compared *relative* to the other candidates, so no physical/logical
  boundary guessing is needed;
- equally-shallow matches (`a/foo`, `b/foo`) → `ErrGetStreamMultipleFiles` (new, 264013);
- no match → `ErrFileNotExists`; an explicit empty name → error.

**Known limitation (documented in godoc + a marked test):** a *lone* deeper namesake can't be
detected — requesting `foo` when the result is only `foo/foo` returns `foo/foo`, since one
physical path gives no boundary and the depth tiebreak needs ≥2 candidates. Full correctness
requires the GET to resolve to the exact object server-side; the driver selection is the backstop
for the multi-file race, not a logical-identity oracle.

## Consumer migration

The GET SQL, `ExecContext`, and buffer read are identical — only the context constructor changes.

**Before** (silent corruption when the path matched >1 file):
```go
var buf bytes.Buffer
ctx := gosnowflake.WithFileGetStream(ctx, &buf)
_, err := db.ExecContext(ctx, "GET 'snow://…/minimap_status' 'file:///tmp/x'")
// buf may hold corrupt/wrong/nondeterministic bytes; err is nil.
```

**After** (opt-in; one line changes):
```go
var buf bytes.Buffer
ctx := gosnowflake.WithFileGetStreamForExactFile(ctx, &buf, "minimap_status") // ← only difference
_, err := db.ExecContext(ctx, "GET 'snow://…/minimap_status' 'file:///tmp/x'")
if err != nil {
    var se *gosnowflake.SnowflakeError
    if errors.As(err, &se) && se.Number == gosnowflake.ErrFileNotExists { /* not there */ }
    if errors.As(err, &se) && se.Number == gosnowflake.ErrGetStreamMultipleFiles { /* ambiguous */ }
    return err
}
data := buf.Bytes() // exactly minimap_status's bytes
```

**Plain `WithFileGetStream` is byte-for-byte unchanged** — selection is a no-op without an
explicit name, so nothing is forced on existing callers. This is a safe opt-in that consumers
adopt explicitly; a follow-up can make single-file selection the default once it is validated in
the field. No public API is removed or changed; this adds one helper and one error constant.

## Tests

Offline unit tests (no live connection — `exceptionTelemetry` tolerates a nil conn), using
realistic `versions/<entity>/<versionId>/<logical>` paths with **differing version ids**:

- leaf match (`foo` not `foobar`), dotted suffixes (`profiles.yml` vs `.template`);
- shallowest-wins across 2 and 3 depths; unique-shallowest when none is top-level;
- equally-shallow tie (and tie-with-a-deeper-also-present) → `ErrGetStreamMultipleFiles`;
- fuller-path disambiguation; case-sensitivity; no-match → `ErrFileNotExists`;
- empty-name guard; **encryption-material alignment** regression;
- a clearly-marked characterization test for the lone-deeper-namesake limitation.

Validated end-to-end against a real internal workspace stage where a GET prefix-matches sibling
files:
- **before:** requesting `foo` returned correct content **0 / 5** runs (nondeterministic
  corruption);
- **after:** `WithFileGetStreamForExactFile("foo")` returned the byte-for-byte correct file every
  run (incl. `foo` alongside a nested `foo/foo`); a nonexistent name → `ErrFileNotExists`; and
  plain `WithFileGetStream` still reproduced the old behavior (confirming it is unchanged).

## Notes for reviewers

- cc the file-transfer / #1687 area owner.
- Complementary server-side narrowing (anchored/exact GET) would close the lone-namesake
  limitation at the source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Upcoming release
 
 New features:
+- Added `WithFileGetStreamForExactFile(ctx, writer, fileName)`, an opt-in variant of `WithFileGetStream` for streaming a single named file when a `GET` prefix-matches more than one (e.g. `foo` alongside `foobar` or a nested `foo/foo`). In that case plain `WithFileGetStream` streams every matched file into one shared buffer and can return corrupt or wrong-file bytes; the new variant selects the requested file before any download goroutine is spawned - matching by leaf and preferring the shallowest path when a name also appears deeper - and returns `ErrFileNotExists` (no match) or the new `ErrGetStreamMultipleFiles` (multiple equally-specific matches) instead of streaming the wrong file. Plain `WithFileGetStream` behavior is unchanged (snowflakedb/gosnowflake#1808).
 
 Bug fixes:
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
--- a/connection_util.go
+++ b/connection_util.go
@@ -212,6 +212,15 @@ func isFileGetStream(ctx context.Context) bool {
 	return v != nil
 }
 
+// getFileGetStreamExactFile returns the file name set by WithFileGetStreamForExactFile and
+// whether it was set. The boolean distinguishes "no target" (plain WithFileGetStream, where
+// ok is false) from "an explicit empty target" (a caller misuse, where ok is true and the
+// value is "").
+func getFileGetStreamExactFile(ctx context.Context) (string, bool) {
+	v, ok := ctx.Value(fileGetStreamExactFile).(string)
+	return v, ok
+}
+
 func getFileTransferOptions(ctx context.Context) *SnowflakeFileTransferOptions {
 	v := ctx.Value(fileTransferOptions)
 	if v == nil {
--- a/doc.go
+++ b/doc.go
@@ -1465,6 +1465,23 @@ To download a file into an in-memory stream (rather than a file) use code simila
 	// streamBuf is now filled with the stream. Use bytes.NewReader(streamBuf.Bytes()) to read uncompressed stream or
 	// use gzip.NewReader(&streamBuf) for to read compressed stream.
 
+A GET resolves its stage path by prefix matching, so it can match more than one file (for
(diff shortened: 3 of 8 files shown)


sfc-gh-jchernandez:
Superseded by #1809, which is a same-repo PR (no fork) now that I have write access. Closing this fork-based one.

PR #1809, opened 2026-06-24
SNOW-3637888: Error on multi-file streaming GET instead of returning corrupt bytes

sfc-gh-jchernandez:
## Problem

[SNOW-3637888] A streaming `GET` (`WithFileGetStream`) corrupts its output when the GET path
prefix-matches more than one file. A GET resolves its stage path by prefix, so `GET @stage/foo`
also matches `foobar`, a nested `foo/foo`, etc. Because a get-stream has a single `io.Writer`,
every matched file's download goroutine writes that one shared buffer, so the caller silently gets
a nondeterministic concat/truncation (encrypted stages) or one arbitrary file (unencrypted) — with
no error.

Continues the file-transfer hardening from #1687, which made a panicking download goroutine
non-fatal but didn't address the cross-file shared-buffer contention.

## Fix

A get-stream can return only one file. When a `GET` matches more than one, the driver now returns
the new `ErrGetStreamMultipleFiles` (error code `264013`) **before any download goroutine is
spawned**, instead of streaming a corrupt mix.

Per review, the driver does **not** parse the SQL or guess which file the caller meant — that
selection logic would be specific to this one driver and can't be preserved under the upcoming
unified-driver work. Instead the error is meaningful and points the caller at the supported,
server-side way to narrow a GET to a single file: the **`PATTERN`** argument, e.g.

```sql
GET @~ file:///tmp/testData PATTERN='.*data1.txt.gz'
```

This adds one exported error constant (`ErrGetStreamMultipleFiles`) and changes no existing
signature. The check runs only for get-stream and after the encryption-material map is built, so a
single streamed file keeps its material.

## Behavior (before → after)

| # | Case | `GET` → server prefix-match | Before | After |
|---|------|------------------------------|--------|-------|
| 1 | Single result | `foo` → `[foo]` | streams `foo` | streams `foo` — unchanged |
| 2 | Whole-stage / folder GET, one file | `@%table` / `@stage` / `@stage/dir` → `[onefile]` | streams `onefile` | streams `onefile` — unchanged |
| 3 | Single result, not the exact name | `foo` → `[foobar]` | streams `foobar` | streams `foobar` — unchanged¹ |
| 4 | Exact file + prefix-sibling | `foo` → `[foo, foobar]` | **corrupt**, no error | `ErrGetStreamMultipleFiles` |
| 5 | Same leaf, different dirs | `foo` → `[a/foo, b/foo]` | **corrupt**, no error | `ErrGetStreamMultipleFiles` |
| 6 | File + same-named subdir | `foo` → `[foo, foo/foo]` | **corrupt**, no error | `ErrGetStreamMultipleFiles` |
| 7 | Whole-stage / folder GET, many files | `@%table` → `[a, b]` | **corrupt**, no error | `ErrGetStreamMultipleFiles` |
| 8 | Server returns nothing | `nope` → `[]` | `ErrInvalidStageLocation` | same — unchanged |
| 9 | Non-stream dir GET (not get-stream) | `@stage/dir` → `[a, b]` → local dir | downloads both | unchanged |
| 10 | PUT / streaming PUT | — | unchanged | unchanged |

"Corrupt" = all matched files share one buffer (encrypted: per-file `Truncate` ⇒ concat/truncation;
unencrypted: last-writer-wins). The multi-file check runs only for get-stream (rows 9–10 and
non-stream GETs are untouched).

**¹ Single-result note (row 3):** a lone result is streamed as-is. The N>1 corruption — the actual
bug — is fully fixed; the caller uses `PATTERN` when they need to guarantee the GET resolves to a
specific single object.

## Tests

Offline unit tests (`getstream_select_test.go`, run with `SKIP_SETUP=1`) cover the single-file
pass-through and every multi-file case erroring with `ErrGetStreamMultipleFiles`, plus a guard that
the single streamed file keeps its own encryption material (the check runs after the
source-file → material map is built). The full `put_get` integration suite (covering
whole-stage/folder get-stream) passes across the matrix.

cc the file-transfer / #1687 area owner.

[SNOW-3637888]: https://snowflakecomputing.atlassian.net/browse/SNOW-3637888

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ New features:
 
 Bug fixes:
 - Fixed nil pointer dereference panic when a corrupt or malformed OCSP cache key is encountered, either from the remote OCSP cache server or the local cache file (snowflakedb/gosnowflake#1819).
+- Fixed `WithFileGetStream` returning corrupt or wrong-file bytes when a `GET` prefix-matched more than one file; a get-stream can return only one file, so a multi-file match now returns the new `ErrGetStreamMultipleFiles` (pointing to the GET `PATTERN` argument) instead of a nondeterministic mix (snowflakedb/gosnowflake#1809).
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
 - Fixed gosnowflake writing a `gosnowflake-cgo` directory under the system temp dir at package import time even when the driver was never used (e.g. when imported only as a transitive dependency). Minicore now loads lazily when the driver is first referenced (`NewConnector`/`OpenWithConfig`) instead of in `init()` (snowflakedb/gosnowflake#1807).
 - Fixed `GET` from a LOCAL_FS stage downloading 0 files and returning `264011: not implemented`. Cloud downloads were unaffected (snowflakedb/gosnowflake#1810).
--- a/file_transfer_agent.go
+++ b/file_transfer_agent.go
@@ -261,6 +261,15 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
 			}, sfa.sc)
 		}
 
+		// A get-stream has a single io.Writer, so a GET matching more than one file would
+		// stream a corrupt mix. Fail deterministically instead of guessing which file the
+		// caller meant, pointing them at the GET PATTERN argument to narrow the match.
+		if isFileGetStream(sfa.ctx) {
+			if err = sfa.ensureSingleGetStreamFile(); err != nil {
+				return err
+			}
+		}
+
 		sfa.localLocation, err = expandUser(sfa.data.LocalLocation)
 		if err != nil {
 			return err
@@ -295,6 +304,22 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
 			QueryID:     sfa.data.QueryID,
 			Message:     errors2.ErrMsgInvalidStageFs,
 			MessageArgs: []any{sfa.stageLocationType},
+		}, sfa.sc)
+	}
+	return nil
+}
+
+// ensureSingleGetStreamFile fails a get-stream that matched more than one file. A GET resolves its
+// stage path by prefix, so it can match several files, but a get-stream has a single io.Writer and
+// streaming them all into it yields a corrupt mix. Rather than guess which file the caller meant,
+// return ErrGetStreamMultipleFiles so the caller narrows the GET with its PATTERN argument.
+func (sfa *snowflakeFileTransferAgent) ensureSingleGetStreamFile() error {
+	if len(sfa.srcFiles) > 1 {
+		return exceptionTelemetry(&SnowflakeError{
+			Number:   ErrGetStreamMultipleFiles,
+			SQLState: sfa.data.SQLState,
+			QueryID:  sfa.data.QueryID,
+			Message:  errors2.ErrMsgGetStreamMultipleFiles,
 		}, sfa.sc)
 	}
 	return nil
--- a/getstream_select_test.go
+++ b/getstream_select_test.go
@@ -1,0 +1,105 @@
+package gosnowflake
+
+import "testing"
+
+// TestEnsureSingleGetStreamFile checks that a get-stream returns its sole matched file unchanged
+// and errors deterministically when the GET matched more than one. No live connection needed: it
+// only reads srcFiles/data. Physical paths follow versions/<entity>/<versionId>/<logical>, with a
+// per-file <versionId>.
+func TestEnsureSingleGetStreamFile(t *testing.T) {
+	ws := "versions/29_559629944250378"
+	v1 := ws + "/1782146627271"
+	v2 := ws + "/1782164111378"
+	v3 := ws + "/1782146628397"
+
+	tests := []struct {
+		name       string
+		srcFiles   []string
+		wantFiles  []string // expected srcFiles on success (nil when an error is expected)
+		wantErrNum int      // 0 when no error expected
+	}{
+		{
+			name:      "single result is streamed",
+			srcFiles:  []string{v1 + "/report.csv"},
+			wantFiles: []string{v1 + "/report.csv"},
+		},
+		{
+			name:      "single prefix-sibling is streamed",
+			srcFiles:  []string{v1 + "/foobar"},
+			wantFiles: []string{v1 + "/foobar"},
+		},
+		{
+			name:       "prefix-sibling match is ambiguous (foo alongside foobar)",
+			srcFiles:   []string{v1 + "/foo", v2 + "/foobar"},
+			wantErrNum: ErrGetStreamMultipleFiles,
+		},
+		{
+			name:       "file and same-named subdir are ambiguous",
+			srcFiles:   []string{v1 + "/foo", v2 + "/foo/foo"},
+			wantErrNum: ErrGetStreamMultipleFiles,
+		},
+		{
+			name:       "same leaf in different directories is ambiguous",
+			srcFiles:   []string{v1 + "/a/foo", v2 + "/b/foo"},
+			wantErrNum: ErrGetStreamMultipleFiles,
+		},
+		{
+			name:       "whole-stage GET with multiple files is ambiguous",
+			srcFiles:   []string{v1 + "/a.csv.gz", v2 + "/b.csv.gz"},
+			wantErrNum: ErrGetStreamMultipleFiles,
+		},
+		{
+			name:       "three prefix-matched files are ambiguous",
+			srcFiles:   []string{v1 + "/foo", v2 + "/foobar", v3 + "/foobaz"},
+			wantErrNum: ErrGetStreamMultipleFiles,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sfa := &snowflakeFileTransferAgent{
+				commandType: downloadCommand,
+				srcFiles:    tt.srcFiles,
+				data:        &execResponseData{},
+			}
+
+			err := sfa.ensureSingleGetStreamFile()
+
+			if tt.wantErrNum != 0 {
+				var sfErr *SnowflakeError
+				assertErrorsAsF(t, err, &sfErr)
+				assertEqualE(t, sfErr.Number, tt.wantErrNum)
+				return
+			}
+			assertNilF(t, err)
+			assertDeepEqualE(t, sfa.srcFiles, tt.wantFiles)
+		})
+	}
+}
+
+// TestEnsureSingleGetStreamFileKeepsEncryptionMaterial guards that the single retained file keeps
+// its own encryption material: the check runs after the source-file -> encryption-material map is
+// built in parseCommand and, on the single-file path, does not disturb srcFiles ordering.
+func TestEnsureSingleGetStreamFileKeepsEncryptionMaterial(t *testing.T) {
+	m0 := &snowflakeFileEncryption{SMKID: 100, QueryID: "q0"}
+	ws := "versions/29_559629944250378"
+	srcFiles := []string{ws + "/1782146627271/alpha"}
+
+	sfa := &snowflakeFileTransferAgent{
+		commandType:        downloadCommand,
+		srcFiles:           srcFiles,
+		encryptionMaterial: []*snowflakeFileEncryption{m0},
+		data:               &execResponseData{SrcLocations: srcFiles},
+	}
+
+	// Mirror parseCommand: build the material map, then run the single-file guard.
+	sfa.srcFileToEncryptionMaterial = make(map[string]*snowflakeFileEncryption)
+	for i, f := range sfa.srcFiles {
+		sfa.srcFileToEncryptionMaterial[f] = sfa.encryptionMaterial[i]
+	}
+	assertNilF(t, sfa.ensureSingleGetStreamFile())
+
+	assertNilF(t, sfa.initFileMetadata())
+	assertEqualE(t, len(sfa.fileMetadata), 1)
+	assertEqualE(t, sfa.fileMetadata[0].encryptionMaterial, m0)
+}
--- a/doc.go
+++ b/doc.go
@@ -1480,6 +1480,14 @@ To download a file into an in-memory stream (rather than a file) use code simila
 	// streamBuf is now filled with the stream. Use bytes.NewReader(streamBuf.Bytes()) to read uncompressed stream or
 	// use gzip.NewReader(&streamBuf) for to read compressed stream.
 
+A GET resolves its stage path by prefix matching, so it can match more than one file (for
+example "foo" alongside "foobar"). Because a get-stream has a single io.Writer, it can return
+only one file: if the GET matches more than one, it returns ErrGetStreamMultipleFiles instead of
+streaming a corrupt mix of them. Use the GET command's PATTERN argument to narrow the match to a
+single file, for example:
+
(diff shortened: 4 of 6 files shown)


util.go
```diff
@@ -82,6 +83,27 @@ func WithFileGetStream(ctx context.Context, writer io.Writer) context.Context {
...
+// shallowest path wins; it returns ErrFileNotExists if nothing matches and
+// ErrGetStreamMultipleFiles if several equally-specific files match.
+//
+// Known limitation: matching is by leaf, so if the GET resolves to ONLY a deeper namesake -
+// you request "foo" but the result contains just "foo/foo" and no top-level "foo" - that file
+// is returned. From a single physical path the driver cannot tell its logical name is "foo/foo"
+// rather than "foo". To be fully exact, ensure the GET resolves to the precise object on the
+// server side; the driver selection is a backstop for the multi-file race, not a substitute for
+// an exact GET.
+func WithFileGetStreamForExactFile(ctx context.Context, writer io.Writer, fileName string) context.Context {
```
sfc-gh-boler: Why aren't we fixing the corruption for all `WithFileGetStream` callers by default — what are the drawbacks? A get-stream that resolves to N>1 files has no correct single-buffer result, so the default should error (e.g. `ErrGetStreamMultipleFiles`) instead of silently corrupting.

As written, `WithFileGetStreamForExactFile` is doing too much: it fixes the multi-file corruption *and* adds exact-file selection in one swapped implementation. Those are separate concerns and bundling them is the wrong approach — it leaves the unsafe default in place. Make the default safe first; add exact-file selection as a distinct feature on top.

file_transfer_agent.go
```diff
@@ -296,6 +305,89 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
...
+			QueryID:     sfa.data.QueryID,
+			Message:     errors2.ErrMsgFileNotExists,
+			MessageArgs: []any{requestedExactFile},
+		}, sfa.sc)
+	}
+
+	// Prefer the shallowest match (fewest path segments): the file named exactly the
+	// request rather than a deeper namesake. Take it when unique; only equally-shallow
+	// namesakes remain ambiguous.
+	minDepth := strings.Count(matched[0], "/")
```
sfc-gh-boler: The "shallowest wins" tiebreak assumes uniform prefix depth. It breaks for general stages: `foo` resolving to `a/foo` and `b/c/foo` are distinct files, but selection silently picks `a/foo`. Any multi-directory leaf collision should return `ErrGetStreamMultipleFiles`. Erroring here isn't a meaningful breaking change — these callers already get corrupt bytes today.

file_transfer_agent.go
```diff
@@ -296,6 +305,89 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
...
+	}
+	if requestedExactFile == "" {
+		// Opted into exact-file streaming but named no file: fail loud rather than
+		// silently falling back to the racy multi-file path (an empty name can never
+		// denote a streamable file).
+		return exceptionTelemetry(&SnowflakeError{
+			Number:   ErrFileNotExists,
+			SQLState: sfa.data.SQLState,
+			QueryID:  sfa.data.QueryID,
+			Message:  errors2.ErrMsgGetStreamExactFileEmpty,
```
sfc-gh-boler: Empty name returns `ErrFileNotExists` with message `ErrMsgGetStreamExactFileEmpty` ("requires a non-empty file name") — the number and message describe different conditions. A dedicated/invalid-argument code would be cleaner for callers matching on `Number`.

sfc-gh-boler (commented):
A few points from review — two main (one design, one correctness) and three minor. Left inline.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Upcoming release
 
 New features:
+- Added `WithFileGetStreamForExactFile(ctx, writer, fileName)`, an opt-in variant of `WithFileGetStream` for streaming a single named file when a `GET` prefix-matches more than one (e.g. `foo` alongside `foobar` or a nested `foo/foo`). In that case plain `WithFileGetStream` streams every matched file into one shared buffer and can return corrupt or wrong-file bytes; the new variant selects the requested file before any download goroutine is spawned - matching by leaf and preferring the shallowest path when a name also appears deeper - and returns `ErrFileNotExists` (no match) or the new `ErrGetStreamMultipleFiles` (multiple equally-specific matches) instead of streaming the wrong file. Plain `WithFileGetStream` behavior is unchanged (snowflakedb/gosnowflake#1808).
```
sfc-gh-boler: References `snowflakedb/gosnowflake#1808`, but this PR is `#1809`. Is `1808` the intended issue link, or a typo?

file_transfer_agent.go
```diff
@@ -257,6 +257,15 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
 			}, sfa.sc)
 		}
 
+		// Narrow a get-stream GET to the single exact file requested. Must run after the
+		// encryption-material map above is built over the full result set (it is keyed
+		// by source file name), so the selected file keeps its correct material.
+		if isFileGetStream(sfa.ctx) {
```
sfc-gh-boler, line 267: `selectGetStreamExactFile()` is gated by `isFileGetStream(sfa.ctx)` here but also no-ops internally when no name is set. The outer guard is redundant (harmless).

util.go
sfc-gh-jchernandez: Agreed — redesigned in 0b305c2. The default `WithFileGetStream` is now safe for everyone: a get-stream resolves the GET to the single file it requested (the leaf of the GET source) and streams that one even when the path prefix-matched several; if it resolves to no file → `ErrFileNotExists`, if ambiguous (>1 match) → `ErrGetStreamMultipleFiles`. No more silent corruption, no opt-in. The separate "exact-file selection" API (`WithFileGetStreamForExactFile`) is removed entirely — the behavior is the default and needs zero caller change.

file_transfer_agent.go
sfc-gh-jchernandez: Agreed — dropped the shallowest-wins tiebreak in 0b305c2. Any multi-directory leaf collision (`a/foo`+`b/c/foo`, or `foo`+`foo/foo`) now returns `ErrGetStreamMultipleFiles`; we never guess. Validated against a real workspace: `foo`+`foo/foo` → 264013.

file_transfer_agent.go
sfc-gh-jchernandez: Resolved by removing the opt-in API in 0b305c2 — there is no longer an explicit name argument, so the empty-name case (and `ErrMsgGetStreamExactFileEmpty`) is gone. The requested file is derived from the GET command instead.

CHANGELOG.md
sfc-gh-jchernandez: Fixed — was a leftover from the closed fork PR; the CHANGELOG now references #1809 (and is reworded as a bug fix, since the default behavior is fixed rather than a new opt-in feature).

file_transfer_agent.go
sfc-gh-jchernandez: Good catch — and with the redesign the guard is now load-bearing rather than redundant: `selectGetStreamFile` errors on N>1, so it must only run for get-stream GETs (a regular dir-mode GET legitimately resolves to many files and must not error). Kept intentionally.

sfc-gh-jchernandez:
Thanks for the review — addressed all five points in `0b305c2` (net −76 lines; the design got simpler). Summary:

**Main redesign (your design + correctness points):** the multi-file corruption is now fixed for **all** `WithFileGetStream` callers **by default**, not behind an opt-in. A get-stream resolves the GET to the single file it requested — the leaf of the GET command's stage source — and streams that one even when the path prefix-matched several. Otherwise it errors instead of corrupting:
- no match → `ErrFileNotExists`
- ambiguous (>1 match, incl. `foo`+`foo/foo` and `a/foo`+`b/c/foo`) → `ErrGetStreamMultipleFiles`

The opt-in `WithFileGetStreamForExactFile` (+ context key, getter, empty-name error) and the shallowest-wins tiebreak are **removed**. This mirrors nuweb's `findFileSourceLocation` and needs **zero consumer change** (PEP's existing call is fixed as-is).

**Minor:** CHANGELOG → #1809 (and recategorized as a bug fix); the `isFileGetStream` guard is now load-bearing (keeps dir-mode GETs out of the new error path).

Empirically re-validated against a real workspace: `GET foo` (alongside `foobar`) now resolves `foo` deterministically under `-race` (was corrupt 0/5); `foo`+`foo/foo` → `ErrGetStreamMultipleFiles`; prefix-without-exact-leaf → `ErrFileNotExists`. The one residual limitation (a lone deeper namesake `foo/foo` for request `foo`) is documented in the godoc + a marked test, with server-side exact-GET as the full fix.

file_transfer_agent.go
```diff
@@ -296,6 +307,84 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
 	return nil
 }
 
+// getStageSourceFromCommand returns the stage-source argument of a GET command (the first
+// argument after GET), with surrounding single quotes and a leading '@' removed. It returns
+// "" if the source cannot be determined. The caller typically takes baseName of the result to
+// learn which file the GET requested.
+func (sfa *snowflakeFileTransferAgent) getStageSourceFromCommand(command string) string {
+	loc := regexp.MustCompile(getRegexp).FindStringIndex(command)
```
sfc-gh-pfus: We shouldn't parse query. To the contrary, I was considering removing those regexps at all. Backend should return the parsed location with the query response.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 New features:
 
 Bug fixes:
+- Fixed `WithFileGetStream` returning corrupt or wrong-file bytes when a streaming `GET` prefix-matched more than one file (e.g. `foo` alongside `foobar`, or a nested `foo/foo`). Every matched file's download goroutine wrote one shared buffer, producing a nondeterministic mix. A get-stream now streams a single file: when the GET names a specific file it returns that one even if the path prefix-matched others, and a multi-file get-stream that can't be resolved to one file returns the new `ErrGetStreamMultipleFiles` instead of corrupt bytes. The single-file case (including whole-stage and folder GETs that resolve to one file) is unchanged (snowflakedb/gosnowflake#1809).
```
sfc-gh-pfus: We usually make one-sentencers here, more details are in PR description. Let's keep it concise.

getstream_select_test.go
```diff
@@ -0,0 +1,271 @@
...
+		{"get @~/data1.txt.gz file:///tmp/testData", "~/data1.txt.gz"},
+		{"/* c */ GET 'snow://w/foo' 'file:///x'", "snow://w/foo"},
+		{"GET 'snow://w/live/' 'file:///tmp'", "snow://w/live/"},
+		{"SELECT 1", ""},
+		{"GET ", ""},
+		{"GET '@stage/unclosed file:///tmp", ""},
+	}
+	for _, tt := range tests {
+		if got := sfa.getStageSourceFromCommand(tt.command); got != tt.want {
+			t.Errorf("getStageSourceFromCommand(%q) = %q, want %q", tt.command, got, tt.want)
```
sfc-gh-pfus: We have a bunch of our assertions like `assertEqualE` that we use in the whole codebase, can you use them in your PR?

file_transfer_agent.go
```diff
@@ -296,6 +307,84 @@ func (sfa *snowflakeFileTransferAgent) parseCommand() error {
...
+// to one file and a file GET whose exact name is absent (only a prefix-sibling remains) are
+// indistinguishable from the command and physical paths alone. Likewise a lone deeper namesake -
+// the only result for "foo" being "foo/foo" - is streamed (its leaf is "foo"). Full exactness
+// requires the GET to resolve to the precise object server-side.
+func (sfa *snowflakeFileTransferAgent) selectGetStreamFile() error {
+	// If the GET names a specific file and exactly one result is that file, stream just it.
+	if requested := baseName(sfa.getStageSourceFromCommand(sfa.command)); requested != "" {
+		matched := make([]string, 0, 1)
+		for _, srcFile := range sfa.srcFiles {
+			if srcFile == requested || strings.HasSuffix(srcFile, "/"+requested) {
```
sfc-gh-pfus: Why can't we use an exact path matching only? With this implementation, if you have files `foo` and `foo2`, `foo` can _never_ be downloaded via streams. I understand that usual GET works on prefixes and downloads all matched files, but in case of streams we always talk about single file (after your fix, which is generally great!).

sfc-gh-pfus (changes requested):
Please adjust PR title, as this is going to be a commit msg eventually.

file_transfer_agent.go
sfc-gh-jchernandez: It already works for both — matching is exact-leaf, not prefix: `baseName(srcFile) == requested`.

- `GET @stage/foo` → server prefix-matches `[foo, foo2]`; `baseName(".../foo2") = "foo2" != "foo"`, so only `foo` matches → `foo` streams.
- `GET @stage/foo2` → server matches `[foo2]` only → `foo2` streams.

So `foo` is downloadable even alongside `foo2` (covered by the `foo`/`foobar` and 3-file test cases, and validated end-to-end). The only thing exact-leaf cannot do is disambiguate two files sharing the *same* leaf in different directories (`a/foo` vs `b/foo`) — those correctly error.

CHANGELOG.md
sfc-gh-jchernandez: Done — collapsed to a single sentence.

getstream_select_test.go
sfc-gh-jchernandez: Done — the tests now use `assertEqualE` / `assertDeepEqualE` / `assertErrorsAsF` / `assertNilF`.

file_transfer_agent.go
sfc-gh-jchernandez: I agree parsing the sql is not ideal, and that having the GET response include the parsed remote location (just like the parsed local location) is cleaner.

However I would like to decouple changing the GET contract, it is a bigger change, used widely, will need to support it forever, confusing you get a parsed source and src_locations list, etc

Parsing the sql is not clean but it is something the driver already needs to do for GET vs PUT, `file://` for put source.

The parsing here is light, it rides on the battle tested getRegexp and if it did not work for whatever reason it falls back to the stricter count must be 1 or throw behavior.

IMO the right thing to do is to go out with this change as is.

sfc-gh-pfus: >Parsing the sql is not clean but it is something the driver already needs to do for GET vs PUT, file:// for put source.

I think it's not correct, I think the backend returns put source as well. It's `src_locations` field IIRC which you mentioned.

And as I said - everything is now controlled by the backend response, these regexps are not needed and I was planning to remove them at all.

file_transfer_agent.go
sfc-gh-pfus: Ah, sorry, I misinterpreted your code 🤦

sfc-gh-boler:
@sfc-gh-pfus what's next for this PR?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ New features:
 
 Bug fixes:
 - Fixed nil pointer dereference panic when a corrupt or malformed OCSP cache key is encountered, either from the remote OCSP cache server or the local cache file (snowflakedb/gosnowflake#1819).
+- Fixed `WithFileGetStream` returning corrupt or wrong-file bytes when a `GET` prefix-matched more than one file. A get-stream can return only one file, so a multi-file match now returns the new `ErrGetStreamMultipleFiles` (pointing to the GET `PATTERN` argument) instead of streaming a nondeterministic mix (snowflakedb/gosnowflake#1809).
```
sfc-gh-pfus: Try concising it to one sentence, will you?

sfc-gh-pfus (approved):

CHANGELOG.md
sfc-gh-jchernandez: Done — collapsed to a single sentence.

PR #1816, opened 2026-07-15
SNOW-3651566 Add cleanupTimeout config option

sfc-gh-pfus:
### Description

Adds a `CleanupTimeout` config option (DSN field `cleanupTimeout`, in seconds) that bounds post-cancellation cleanup operations — specifically the query abort sent after a `QueryContext`/`PingContext`/`ExecContext` cancellation or timeout, and the session logout sent by `Close`.

Previously, cleanup ran synchronously on the caller's goroutine using an unbounded background context, meaning a canceled query could block the caller for minutes or hours past its own deadline. With `CleanupTimeout` set to a positive value, the query abort is detached onto its own goroutine bounded by the specified timeout, allowing the caller to regain control at its own deadline. The session logout in `Close` is also bounded by this timeout while remaining synchronous to honor the `io.Closer` contract.

The default value of `0` preserves the existing unbounded behavior to avoid a breaking change. A value of 5 seconds is recommended as a starting point. Because the abort runs concurrently with potential connection reuse when a positive value is set, keeping this value small limits the overlap window.

Note: this field was added to bypass a Behavior Change Release (BCR) and may be subject to change in the next breaking change release.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1,6 +1,7 @@
 package gosnowflake
 
 import (
+	"cmp"
 	"context"
 	"database/sql"
 	"database/sql/driver"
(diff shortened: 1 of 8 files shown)


sfc-gh-pfus:
* **#1816** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1816?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1816?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

connection.go
```diff
@@ -302,8 +303,16 @@ func (sc *snowflakeConn) Close() (err error) {
 
 	if sc.cfg != nil && !sc.cfg.ServerSessionKeepAlive {
 		logger.WithContext(sc.ctx).Debug("Closing session since ServerSessionKeepAlive is false")
-		// we have to replace context with background, otherwise we can use a one that is cancelled or timed out
-		if err = sc.rest.FuncCloseSession(context.Background(), sc.rest, sc.rest.RequestTimeout); err != nil {
+		// we have to strip cancellation, otherwise we can use a ctx that is cancelled or timed out
```
sfc-gh-boler, line 306: The cancel path explains use of nilUUID and it would be good to add a similar explanation here.

sfc-gh-boler (approved):
lgtm

connection.go
sfc-gh-pfus: Done.

PR #1817, opened 2026-07-15
SNOW-3784429: fix token cache key collisions — versioned SHA256 canonical-JSON key (v2)

sfc-gh-mhofman:
## Summary
- Replace the legacy `{host}:{user}:{type}` cache key with a versioned, uniformly SHA-256-hashed key (`SnowflakeTokenCache.v2.<sha256hex>`) derived from a canonical JSON document containing five sorted fields: `idp`, `role`, `snowflake`, `token_type`, `username`.
- Fix multi-account collisions (shared IdP) and multi-role collisions by threading `idp`, `snowflake`, and `role` through all OAuth, ID-token, MFA call sites, eviction paths, and authentication lock keys.
- Both OS keystore (macOS/Windows) and file (Linux) backends now use the final key verbatim — hashing occurs exactly once at key construction time.

## Test plan
- [x] Golden hash test (`TestBuildCacheKeyGoldenHash`) reproduces `SnowflakeTokenCache.v2.75ff2ad6…` exactly
- [x] `TestNormalizeURL` — scheme stripping, port/path uppercasing, no trailing slash
- [x] `TestNormalizeIdentifier` — unquoted uppercased; quoted verbatim; mixed
- [x] Dimension isolation: different hosts, different roles, different token types all produce different keys; MFA with empty role is stable
- [x] `TestFileBackendStoresKeyVerbatim` — stored key equals `SnowflakeTokenCache.v2.<hash>`, no double-hash
- [x] `TestBuildCacheKeyValidation` — empty `snowflake` / empty `username` rejected
- [x] Existing file-based storage manager tests updated and passing
- [x] `go build`, `go vet`, `go fmt` all clean

Ticket: [SNOW-3784429](https://snowflakecomputing.atlassian.net/browse/SNOW-3784429)

Made with [Cursor](https://cursor.com)

[SNOW-3784429]: https://snowflakecomputing.atlassian.net/browse/SNOW-3784429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


auth.go
```diff
@@ -340,10 +340,10 @@ func authenticate(
 			credentialsStorage.deleteCredential(newMfaTokenSpec(sc.cfg.Host, sc.cfg.User))
 		}
 		if sessionParameters[clientStoreTemporaryCredential] == true && sc.cfg.Authenticator == AuthTypeExternalBrowser {
-			credentialsStorage.deleteCredential(newIDTokenSpec(sc.cfg.Host, sc.cfg.User))
+			credentialsStorage.deleteCredential(newIDTokenSpec(sc.cfg.Host, sc.cfg.User, sc.cfg.Role))
```
sfc-gh-pfus, line 343: Why do we need role for IDToken? I thought this is for oauth only?

auth.go
```diff
@@ -643,29 +649,35 @@ func prepareJWTToken(config *Config) (string, error) {
 }
 
 type tokenLockKey struct {
-	snowflakeHost string
-	user          string
-	tokenType     string
+	idp       string
+	snowflake string
+	user      string
+	role      string
```
sfc-gh-pfus, line 655: Why do we need role here and idp here?

auth.go
```diff
@@ -552,29 +552,35 @@ func createRequestBody(sc *snowflakeConn, sessionParameters map[string]any,
 }
 
 type oauthLockKey struct {
-	tokenRequestURL string
-	user            string
-	flowType        string
+	idp       string
```
sfc-gh-pfus, line 555: I'm wondering whether we should have so long list of params, maybe we should make this global (if someone has 5 unique connection configurations, still will get 5 MFAs or browser tabs at once) 🤔 But probably it doesn't matter now and can wait for UD.

auth_oauth_test.go
```diff
@@ -182,6 +183,7 @@ func TestUnitOAuthClientCredentials(t *testing.T) {
 			ClientStoreTemporaryCredential: ConfigBoolTrue,
 		}
 	}
+	cacheTokenSpec := newOAuthAccessTokenSpec(cfgFactory().OauthTokenRequestURL, cfgFactory().Host, cfgFactory().User, cfgFactory().Role)
```
sfc-gh-pfus: For every parameter to this function you create another `Config` instance, is it correct?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 New features:
 
 Bug fixes:
+- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#N).
```
sfc-gh-pfus: Fill PR number here.

secure_storage_manager.go
```diff
@@ -43,45 +44,62 @@ var defaultLinuxCacheDirConf = []cacheDirConf{
 	{envVar: "HOME", pathSegments: []string{".cache", "snowflake"}},
 }
 
+// cacheKeyInput contains the five fields that define a unique token cache entry.
+// All fields are raw (un-normalized) values; normalization happens inside buildCacheKey.
+type cacheKeyInput struct {
```
sfc-gh-pfus, line 49: I don't think this is a good abstraction. This is a bag of all fields required in all cache keys. I think it may be better to let's say:
1. `secureTokenSpec` - interface with `buildKey` function
2. structs that contains required fields only (`simple`/`snowflake` which has snowflakeHost + username, `mfa`/`id` which inherit it, similar for `oauth`)

We could have a dedicated validations for each type then (IIUC, all fields should be filled for specific token spec). WDYT?

secure_storage_manager.go
```diff
@@ -464,17 +482,70 @@ func (ssm *fileBasedSecureStorageManager) writeTemporaryCacheFile(cache map[stri
...
+
+	keyData := map[string]string{
+		"idp":        normalizeURL(input.idp),
+		"role":       normalizeIdentifier(input.role),
+		"snowflake":  normalizeURL(input.snowflake),
+		"token_type": string(input.tokenType),
+		"username":   normalizeIdentifier(input.username),
+	}
+
+	// json.Marshal on a map[string]string emits compact JSON with keys in
```
sfc-gh-pfus, line 503: Can we guarantee the same for all drivers? Ordering in JSON, URL handling normalisation, indents, new lines and so on. Maybe we should have some tests matching with a reference? Edit: I can see tests, but my concerns are still valid.

secure_storage_manager.go
```diff
@@ -464,17 +482,70 @@ func (ssm *fileBasedSecureStorageManager) writeTemporaryCacheFile(cache map[stri
...
+	}
+
+	sum := sha256.Sum256(jsonBytes)
+	hexHash := hex.EncodeToString(sum[:])
+	return "SnowflakeTokenCache.v2." + hexHash, nil
+}
+
+// normalizeURL strips the scheme and userinfo, drops query/fragment,
+// trims a root-only trailing slash, and uppercases the remainder.
+func normalizeURL(rawURL string) string {
```
sfc-gh-pfus, line 517: This logic is hard to keep the same in all drivers, maybe just hash of the URL is enough?

auth.go
sfc-gh-mhofman: Correct but then the interface is not unique across the board. We have same approach and in the end role is also derived or from the default.

secure_storage_manager.go
sfc-gh-mhofman: i can reply same way here as for the top most comment
https://github.com/snowflakedb/gosnowflake/pull/1817/changes#r3602477866

secure_storage_manager.go
sfc-gh-mhofman: that's why there's this golden key test across each driver. But what do you then suggest instead of this? the key got complicated

auth_oauth_test.go
sfc-gh-mhofman: Good catch — fixed. The four `cfgFactory()` calls on that line now collapse into a single `cfg := cfgFactory()` instance whose fields are used to build the spec and the client. The separate per-subtest `cfgFactory()` calls (which intentionally produce fresh configs) are left untouched.

CHANGELOG.md
sfc-gh-mhofman: Fixed — replaced `#N` with `#1817` in the CHANGELOG entry.

auth.go
sfc-gh-mhofman: To follow up on my earlier reply: the role inclusion for IDToken is being revisited in the fixup PR snowflakedb/gosnowflake#1821, which reworks the per-auth-type key structure so each token type only carries the fields it actually needs.

auth.go
sfc-gh-mhofman: The role and IdP fields in `tokenLockKey` are being trimmed in the fixup PR snowflakedb/gosnowflake#1821, which refines the lock-key construction so each auth type carries only the fields relevant to its cache key.

secure_storage_manager.go
sfc-gh-mhofman: To follow up: the per-type `secureTokenSpec` interface approach you describe is exactly the direction we are heading in the fixup PR snowflakedb/gosnowflake#1821. The current flat `cacheKeyInput` struct is a stepping stone; #1821 introduces type-specific specs that carry only the fields each token type requires, enabling the per-type validation you mention.

auth.go
sfc-gh-mhofman: Agreed — the long parameter list is awkward, and making the lock scope global (or at least broader) is worth exploring. Deferring that to the UD work as you suggest. As a small improvement already landed in the fixup PR snowflakedb/gosnowflake#1821: the MFA and IDToken lock keys are trimmed so they no longer carry the full set of fields, which reduces the surface area slightly until the bigger rearchitecture happens.

secure_storage_manager.go
sfc-gh-mhofman: To follow up on my earlier reply: commit `39ea531c` already softened the "identical across drivers" claim in the comments — we no longer assert bit-for-bit compatibility, only that this driver produces a stable, well-defined key. The implementation uses canonical JSON serialization (sorted keys, no extra whitespace) via `encoding/json` with a deterministic struct layout, and the golden-hash tests pin the exact output so any accidental change is caught. The broader cross-driver normalization guarantee (agreeing on URL canonicalization rules, case folding, etc. across Python/Java/Go) is a separate effort tracked as a follow-up.

secure_storage_manager.go
sfc-gh-mhofman: Hashing the raw URL string would bake in any accidental variations (trailing slash, mixed case, default port present/absent) and make keys differ across callers even when they refer to the same endpoint. `normalizeURL` strips the scheme, lowercases the host, removes the default HTTPS port, and drops trailing slashes so that logically equivalent URLs produce the same key. That said, the exact normalization rules (and whether they should match other drivers) are part of the broader key-spec discussion that is being tracked as a follow-up; this PR establishes the versioned-hash infrastructure, and the normalization details can be refined there.

sfc-gh-pfus (approved):

PR #1819, opened 2026-07-17
SNOW-3801358 fix nil pointer dereference panic upon encountering corrupt OCSP cache entry

sfc-gh-dszmolka:
### Problem

When the OCSP cache server (`ocsp.snowflakecomputing.com`) or the local on-disk cache file contained an entry with a corrupt or malformed cache key (invalid base64, or valid base64 but not a valid ASN.1 `certID` structure), the driver panicked with a nil pointer dereference. 

`decodeCertIDKey()` correctly returns `nil` for such keys, but neither `downloadOCSPCacheServer` nor `initOCSPCache` checked for nil before dereferencing the result when writing to `ocspResponseCache`. 

Because `extractOCSPCacheResponseValueWithoutSubject` handles a nil key gracefully and can return a valid OCSP status (when the response value itself is well-formed), the `isValidOCSPStatus` guard did not protect against the subsequent `ocspResponseCache[*cacheKey]` dereference.

### Fix

Added a nil check on the result of `decodeCertIDKey()` in both `downloadOCSPCacheServer` and `initOCSPCache`, immediately after the call and before any further use. Corrupt entries are skipped with a debug log message. Because the entry is never loaded into the in-memory cache, the driver naturally falls back to a live OCSP Responder request for the affected certificate, and `writeOCSPCacheFile` will subsequently persist only the fresh, valid response — effectively overwriting the corrupt entry.

### Tests

* Added `TestUnitDownloadOCSPCacheServerCorruptKey`, which spins up a local HTTP test server serving a JSON payload with an invalid base64 cache key and verifies that `downloadOCSPCacheServer` does not panic and does not store the entry. 
* Added `TestUnitInitOCSPCacheCorruptKey`, which writes a temp cache file containing a valid-base64-but-not-ASN.1 key (simulating a file mangled by an external tool between driver reads) and verifies that `initOCSPCache` does not panic and does not load the entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ New features:
 - Added `CleanupTimeout` config option (DSN field `cleanupTimeout`, in seconds) that bounds post-cancellation cleanup (snowflakedb/gosnowflake#1816).
 
 Bug fixes:
-- Fixed nil pointer dereference panic in OCSP cache loading when a corrupt or malformed cache key is encountered, either from the remote OCSP cache server or the on-disk cache file. Corrupt entries are now skipped and the driver falls back to the live OCSP Responder for the affected certificate (snowflakedb/gosnowflake#XXXX).
+- Fixed nil pointer dereference panic in OCSP cache loading when a corrupt or malformed cache key is encountered, either from the remote OCSP cache server or the on-disk cache file. Corrupt entries are now skipped and the driver falls back to the live OCSP Responder for the affected certificate (snowflakedb/gosnowflake#1819).
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
 - Fixed gosnowflake writing a `gosnowflake-cgo` directory under the system temp dir at package import time even when the driver was never used (e.g. when imported only as a transitive dependency). Minicore now loads lazily when the driver is first referenced (`NewConnector`/`OpenWithConfig`) instead of in `init()` (snowflakedb/gosnowflake#1807).
 - Fixed `GET` from a LOCAL_FS stage downloading 0 files and returning `264011: not implemented`. Cloud downloads were unaffected (snowflakedb/gosnowflake#1810).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ New features:
 - Added `CleanupTimeout` config option (DSN field `cleanupTimeout`, in seconds) that bounds post-cancellation cleanup (snowflakedb/gosnowflake#1816).
 
 Bug fixes:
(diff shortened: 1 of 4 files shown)


ocsp.go
```diff
@@ -812,6 +812,10 @@ func (ov *ocspValidator) downloadOCSPCacheServer() {
 	ocspResponseCacheLock.Lock()
 	for k, cacheValue := range *ret {
 		cacheKey := decodeCertIDKey(k)
```
sfc-gh-jkasten: optional: It looks like the underlying problem is that this function returns in-band errors. I think this goes against conventions which probably raises the likelihood of bugs.
https://go.dev/wiki/CodeReviewComments#in-band-errors

sfc-gh-jkasten (approved):
Thanks, David! Looks like a strict improvement. (Didn't look for other issues)

ocsp.go
sfc-gh-dszmolka: yup - this would also address the problem and make the new nil check not needed. thank you James!

i'm waiting for the Driver owners to look at this and see if they're good with the current change as-is and maybe do another one as a follow-up

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ New features:
 - Added `CleanupTimeout` config option (DSN field `cleanupTimeout`, in seconds) that bounds post-cancellation cleanup (snowflakedb/gosnowflake#1816).
 
 Bug fixes:
+- Fixed nil pointer dereference panic in OCSP cache loading when a corrupt or malformed cache key is encountered, either from the remote OCSP cache server or the on-disk cache file. Corrupt entries are now skipped and the driver falls back to the live OCSP Responder for the affected certificate (snowflakedb/gosnowflake#1819).
```
sfc-gh-pfus: Keep it concise - try making it one sentence only :)

ocsp.go
sfc-gh-pfus: +1, let's return error here.

ocsp_test.go
```diff
@@ -628,3 +630,80 @@ func syncUpdateOcspResponseCache(f func()) {
...
+	// returns nil) but whose value is a well-formed [ts, ocspRespBase64] pair.
+	payload, err := json.Marshal(map[string]any{
+		"not!valid!base64==": []any{
+			float64(time.Now().UTC().Unix()),
+			base64.StdEncoding.EncodeToString([]byte("junk-not-ocsp")),
+		},
+	})
+	assertNilF(t, err, "marshalling test payload")
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
sfc-gh-pfus: Let's use wiremock for such purposes.

ocsp.go
sfc-gh-dszmolka: okie, changed the signature which now should return error per the best practice James pointed out

CHANGELOG.md
sfc-gh-dszmolka: shortened :)

ocsp_test.go
sfc-gh-dszmolka: test now should use wiremock

sfc-gh-pfus (approved):

PR #1821, opened 2026-07-22
SNOW-3784429 token cache key v2 fixup

sfc-gh-mhofman:
### Description

SNOW-3784429  This is changing previous https://github.com/snowflakedb/gosnowflake/pull/1817 to support token caching in two distinct cases differently: 
* OAuth token keys with added Role and IdP host to avoid token collisions in the cache
* SSO/MFA tokens without Role and IdP as such tokens are irrelevant of those values
* Separate golden hash tests provide interoperability of the tokens in the cache across different drivers

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ New features:
 - Added option to load private key from `connections.toml` file (snowflakedb/gosnowflake#1822).
 
 Bug fixes:
-- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#1817).
+- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key with the token type in the key prefix, applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#1821).
 - Fixed nil pointer dereference panic when a corrupt or malformed OCSP cache key is encountered, either from the remote OCSP cache server or the local cache file (snowflakedb/gosnowflake#1819).
 - Fixed `WithFileGetStream` returning corrupt or wrong-file bytes when a `GET` prefix-matched more than one file; a get-stream can return only one file, so a multi-file match now returns the new `ErrGetStreamMultipleFiles` (pointing to the GET `PATTERN` argument) instead of a nondeterministic mix (snowflakedb/gosnowflake#1809).
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ New features:
 - Added option to load private key from `connections.toml` file (snowflakedb/gosnowflake#1822).
 
 Bug fixes:
-- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key with the token type in the key prefix, applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#1821).
+- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key with the token type in the key prefix, applied uniformly across keyring (macOS/Windows) and file (Linux) backends; also replaced the bag-of-fields spec struct with typed token-spec types (`hostUserTokenSpec` for MFA/ID flows, `oauthTokenSpec` for OAuth flows) so each spec validates and serializes only its own fields (snowflakedb/gosnowflake#1817, snowflakedb/gosnowflake#1821).
 - Fixed nil pointer dereference panic when a corrupt or malformed OCSP cache key is encountered, either from the remote OCSP cache server or the local cache file (snowflakedb/gosnowflake#1819).
 - Fixed `WithFileGetStream` returning corrupt or wrong-file bytes when a `GET` prefix-matched more than one file; a get-stream can return only one file, so a multi-file match now returns the new `ErrGetStreamMultipleFiles` (pointing to the GET `PATTERN` argument) instead of a nondeterministic mix (snowflakedb/gosnowflake#1809).
 - Do not attempt to get S3 bucket accelerate config for Snowflake-internal stages (matched by bucket name `sfc-*`) since s3:GetAccelerateConfiguration not granted anyways (snowflakedb/gosnowflake#1805).
--- a/auth.go
+++ b/auth.go
@@ -338,13 +338,13 @@ func authenticate(
 		logger.WithContext(ctx).Error("Authentication FAILED")
 		sc.rest.TokenAccessor.SetTokens("", "", -1)
 		if sessionParameters[clientRequestMfaToken] == true {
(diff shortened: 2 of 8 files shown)


auth.go
```diff
@@ -340,10 +341,10 @@ func authenticate(
 			credentialsStorage.deleteCredential(newMfaTokenSpec(sc.cfg.Host, sc.cfg.User))
 		}
 		if sessionParameters[clientStoreTemporaryCredential] == true && sc.cfg.Authenticator == AuthTypeExternalBrowser {
-			credentialsStorage.deleteCredential(newIDTokenSpec(sc.cfg.Host, sc.cfg.User, sc.cfg.Role))
+			credentialsStorage.deleteCredential(newIDTokenSpec(sc.cfg.Host, sc.cfg.User))
 		}
 		if sessionParameters[clientStoreTemporaryCredential] == true && isOauthNativeFlow(sc.cfg.Authenticator) {
-			credentialsStorage.deleteCredential(newOAuthAccessTokenSpec(sc.cfg.OauthTokenRequestURL, sc.cfg.Host, sc.cfg.User, sc.cfg.Role))
+			credentialsStorage.deleteCredential(newOAuthAccessTokenSpec(oauthTokenRequestURL(sc.cfg), sc.cfg.Host, sc.cfg.User, sc.cfg.Role))
```
sfc-gh-pfus: I think we should change the contract to pass only `cfg` to `oauthTokenRequestURL`. Currently a logic of constructing it (`oauthTokenRequestURL`) slipped out - for auth.go this should be transparent and it is easy to miss in other places. I think it could apply for all constructors.

auth_oauth.go
```diff
@@ -389,7 +389,7 @@ func (oauthClient *oauthClient) refreshToken() error {
 		logger.Debug("credentials storage is disabled, cannot use refresh tokens")
 		return nil
 	}
-	refreshTokenSpec := newOAuthRefreshTokenSpec(oauthClient.cfg.OauthTokenRequestURL, oauthClient.cfg.Host, oauthClient.cfg.User, oauthClient.cfg.Role)
+	refreshTokenSpec := newOAuthRefreshTokenSpec(oauthClient.tokenURL(), oauthClient.cfg.Host, oauthClient.cfg.User, oauthClient.cfg.Role)
```
sfc-gh-pfus: Exactly, here we have a sibling logic. Let's unify it.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ New features:
 - Added option to load private key from `connections.toml` file (snowflakedb/gosnowflake#1822).
 
 Bug fixes:
-- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#1817).
+- Fixed token cache key collisions for multi-account (shared IdP) and multi-role scenarios by switching to a versioned, SHA256-hashed canonical-JSON key with the token type in the key prefix, applied uniformly across keyring (macOS/Windows) and file (Linux) backends (snowflakedb/gosnowflake#1821).
```
sfc-gh-pfus: Both PRs are important here, right? Let's mention both here.

secure_storage_manager.go
```diff
@@ -482,8 +480,23 @@ func (ssm *fileBasedSecureStorageManager) writeTemporaryCacheFile(cache map[stri
 	return nil
 }
 
-// buildCacheKey produces a versioned, SHA-256-hashed cache key from the
-// normalized values of the five key fields.
+// mfaOrIDTokenTypes are the flows whose keyData contains only snowflake +
+// username. All other (OAuth) flows additionally include idp + role.
+var mfaOrIDTokenTypes = map[tokenType]bool{
```
sfc-gh-pfus: Terrible name :) I still stand on my comment on previous PR - it may be better to have different structures. Currently, there is no validation that all inputs are provided and we end up with such horrific names and strange ifs. If you don't want to have multiple structures, at least add validation in constructors, add `json:` tags with `omitempty` and serialise the struct itself.

auth.go
sfc-gh-mhofman: Totally! fixed

auth_oauth.go
sfc-gh-mhofman: Done

CHANGELOG.md
sfc-gh-mhofman: Makes sense, added

secure_storage_manager.go
sfc-gh-mhofman: True, removed and replaced by dedicated structures per both different auth-specific subsets of values

auth.go
```diff
@@ -583,6 +584,13 @@ func (o *oauthLockKey) lockID() string {
 	return o.idp + "|" + o.snowflake + "|" + o.user + "|" + o.role + "|" + o.flowType
 }
 
+// oauthTokenRequestURL returns the full token endpoint URL for OAuth.
+// oauthClient.tokenURL() delegates to this function so that all cache key
+// construction and eviction paths use one canonical URL formula.
+func oauthTokenRequestURL(cfg *Config) string {
```
sfc-gh-pfus: I still think this function is not needed in auth.go. It should belong to either oauth.go or secure_storage_manager.go - I believe oauth.go is better, we can replace `oauthClient.tokenURL()` with this?

auth.go
```diff
@@ -649,35 +657,29 @@ func prepareJWTToken(config *Config) (string, error) {
...
 		snowflake: snowflakeHost,
 		user:      user,
-		role:      role,
 		tokenType: "ID",
 	}
 }
 
 func (m *tokenLockKey) lockID() string {
-	return m.idp + "|" + m.snowflake + "|" + m.user + "|" + m.role + "|" + m.tokenType
+	return m.snowflake + "|" + m.user + "|" + m.tokenType
```
sfc-gh-pfus: I wonder if we could reuse structs from secure_storage_manager.go - after all, these locks are expected to lock on single access to secure storage, so all the parameters should be the same. In Go it is trivial - you can just define a function (`lockID`) on a struct from a different file and you would satisfy `lockKeyType` interface - this is perfectly normal in Go.

secure_storage_manager.go
```diff
@@ -44,68 +44,132 @@ var defaultLinuxCacheDirConf = []cacheDirConf{
...
+	Snowflake string `json:"snowflake"`
+	Username  string `json:"username"`
 }
 
-func (t *secureTokenSpec) buildKey() (string, error) {
-	return buildCacheKey(t.input)
+// Fields must remain in lexicographic order to match the cross-driver hash contract.
+type oauthKeyData struct {
+	Idp       string `json:"idp"`
+	Role      string `json:"role"`
```
sfc-gh-pfus, line 60: Should we add empty role to JSON if not provided? If not, you can add `omitempty`. If yes, this is fine.

secure_storage_manager.go
```diff
@@ -44,68 +44,132 @@ var defaultLinuxCacheDirConf = []cacheDirConf{
...
+		return "", errors.New("snowflake URL is required for token cache key")
+	}
+	if s.username == "" {
+		return "", errors.New("username is required for token cache key")
+	}
+	keyData := mfaIDKeyData{
+		Snowflake: normalizeURL(s.snowflake),
+		Username:  normalizeIdentifier(s.username),
+	}
+	jsonBytes, err := json.Marshal(keyData)
```
sfc-gh-pfus: This code seems to be identical for all keys? Maybe we can move it to common function?

sfc-gh-pfus (approved):

secure_storage_manager.go
sfc-gh-mhofman: omitempty would break cross-driver contract — role is a must even when empty

auth.go
sfc-gh-mhofman: makes sense

auth.go
sfc-gh-mhofman: moved out

PR #1822, opened 2026-07-23
Parse toml pk path

sfc-gh-dgillis:
### Description

Add a privatekeypath / privatekeyfile case to the TOML parser so keypair auth works when the key is given as a file path (as the Snowflake CLI writes it), not just as inline base64 PKCS8. Previously these keys were silently dropped, leaving PrivateKey nil and failing with "trying to use keypair authentication, but PrivateKey was not provided in the driver config". Reuses the existing parsePrivateKeyFromFile helper; adds positive and negative tests.

Ref: https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections#use-a-private-key-file-for-authentication

### Checklist
- [x] Created tests which fail without the change
- [x] Extended the example documentation


sfc-gh-dgillis:

internal/config/connection_configuration.go
```diff
@@ -183,6 +183,13 @@ func HandleSingleParam(cfg *Config, key string, value any) error {
 			}
 		}
 		cfg.PrivateKey, err = ParsePKCS8PrivateKey(block)
+	case "privatekeypath", "privatekeyfile":
```
sfc-gh-pfus, line 186: In JDBC we use `privateKeyFile`, not path. Let's keep it aligned, unless there is a good reason not to do it?

sfc-gh-dgillis: I assume `private_key_path` is in place for backwards compatibility. The snow CLI supports compatibility for both:

```text
Snowflake CLI looks for the private key in the connection parameters in the following order:

If private_key_file is specified, Snowflake CLI reads the key from the specified file path.
If private_key_path is specified, Snowflake CLI reads the key from the specified file path.
If private_key_file or private_key_path are not specified, Snowflake CLI reads the key directly from the private_key_raw parameter.
```

Also, the environment variable the CLI looks for is called `SNOWFLAKE_PRIVATE_KEY_PATH`, see [https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections#use-environment-variables-for-snowflake-credentials](https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections#use-environment-variables-for-snowflake-credentials)

and in the connection flags when invoke the cli you can pass either:

```shell
╭─ Connection configuration ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --connection,--environment             -c      TEXT     Name of the connection, as defined in your config.toml. Default: default.
│ --host                                         TEXT     Host address for the connection. Overrides the value specified for the connection.
│ --port                                         INTEGER  Port for the connection. Overrides the value specified for the connection.
│ --account,--accountname                        TEXT     Name assigned to your Snowflake account. Overrides the value specified for the connection.
│ --user,--username                              TEXT     Username to connect to Snowflake. Overrides the value specified for the connection.
│ --password                                     TEXT     Snowflake password. Overrides the value specified for the connection.
│ --authenticator                                TEXT     Snowflake authenticator. Overrides the value specified for the connection.
│ --private-key-file,--private-key-path          TEXT     Snowflake private key file path. Overrides the value specified for the connection.
│ --token                                        TEXT     OAuth token to use when connecting to Snowflake.
│ --token-file-path                              TEXT     Path to file with an OAuth token that should be used when connecting to Snowflake.
│ --database,--dbname                            TEXT     Database to use. Overrides the value specified for the connection.
│ --schema,--schemaname                          TEXT     Database schema to use. Overrides the value specified for the connection.
│ --role,--rolename                              TEXT     Role to use. Overrides the value specified for the connection.
│ --warehouse                                    TEXT     Warehouse to use. Overrides the value specified for the connection.
│ --temporary-connection                 -x               Uses connection defined with command-line parameters, instead of one defined in config.
│ --mfa-passcode                                 TEXT     Token to use for multi-factor authentication (MFA).
│ --oauth-client-id                              TEXT     Value of the client ID provided by the identity provider for Snowflake integration.
│ --oauth-client-secret                          TEXT     Value of the client secret provided by the identity provider for Snowflake integration.
│ --oauth-authorization-url                      TEXT     Identity provider endpoint supplying the authorization code to the driver.
│ --oauth-token-request-url                      TEXT     Identity provider endpoint supplying the access tokens to the driver.
│ --oauth-redirect-uri                           TEXT     URI to use for the authorization code.
│ --oauth-scope                                  TEXT     Scope requested in the identity provider authorization request.
│ --oauth-disable-pkce                                    Disables Proof Key for Code Exchange (PKCE). Default: False.
│ --oauth-enable-refresh-tokens                           Enables a silent re-authentication when the actual access token becomes outdated. Default: False.
│ --oauth-enable-single-use-refresh-tokens                Whether to opt in to single-use refresh token semantics. Default: False.
│ --client-store-temporary-credential                     Store the temporary credential.
│ --enable-diag                                           Run the python connector diagnostic test.
│ --diag-log-path                                TEXT     Diagnostic report path.
│ --diag-allowlist-path                          TEXT     Diagnostic report path to optional allowlist.
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Personally, I would actually update the JDBC driver to be compatible with both also for consistency, but your call - we can just look for the one!

sfc-gh-pfus: Ok, this makes sense to have it too, thanks!

sfc-gh-pfus (approved):

PR #1828, opened 2026-07-29
SNOW-3831259 Download OCSP cache only if any cert has OCSP responder

sfc-gh-pfus:
### Description

This PR fixes three related OCSP revocation correctness issues:

Multi-chain acceptance policy: verifyPeerCertificate now accepts a connection as soon as any one verified certificate chain is fully unrevoked, and rejects only when every chain fails. Previously the first chain with an error would abort the connection even if a later chain was clean. This aligns OCSP behaviour with the CRL validator.

Unnecessary OCSP cache server downloads: The cache server is no longer contacted when no certificate in any verified chain has an OCSP responder URL. Certificates without a responder URL are never present in the cache server, so downloading it in that case was wasteful and incorrect.

Fail-closed fast-fail on missing responder URL: In fail-closed mode, if any certificate in a chain lacks an OCSP responder URL the chain is immediately failed without querying responders for the remaining certificates. A chain that cannot be fully verified must be rejected in fail-closed mode, and there is no point contacting other responders once the outcome is determined.

Internally, canEarlyExitForOCSP has been replaced by evaluateChain, which returns a typed ocspChainResult (ocspChainUnrevoked, ocspChainRevoked, or ocspChainError) rather than a raw *ocspStatus. The cache-flush logic is now handled by a dedicated flushOCSPCache method called via defer. The certificate factory helpers shared between the CRL and OCSP test suites have been consolidated into a new revocation_test.go file, and createLeafCert now accepts an ocspServerType parameter to embed OCSP responder URLs directly into test certificates. A new TestVerifyPeerCertificateMultiChain test exercises all multi-chain scenarios entirely from the in-memory cache with no network calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ New features:
 - Added support for Go 1.27, dropped support for Go 1.24 (snowflakedb/gosnowflake#TBD).
 
 Bug fixes:
+- OCSP revocation now accepts a connection when any one verified certificate chain is fully unrevoked (rejecting only when every chain fails), skips the OCSP cache server download when no certificate has a responder URL to check, and in fail-closed mode fails a chain fast when a certificate has no OCSP responder URL instead of querying responders for the others (snowflakedb/gosnowflake#1828).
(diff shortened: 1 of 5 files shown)


sfc-gh-pfus:
* **#1828** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1828?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1828?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #1833, opened 2026-08-06
SNOW-2499292 Add serializable Arrow batches for distributed fetch

sfc-gh-pfus:
### TL;DR

Adds `SerializableArrowBatch` with `ArrowBatch.ToSerializable` and `SerializableArrowBatch.ToArrowBatch` to enable distributed fetch of Arrow result batches across machines with no Snowflake connection.

### What changed?

- Introduced `arrowbatches.SerializableArrowBatch`, a portable JSON-serializable descriptor that captures everything needed to fetch a result chunk: the presigned URL, SSE-C headers, inline IPC bytes (for the first batch), row count, column metadata, timezone, and conversion options.
- Added `ArrowBatch.ToSerializable()` to convert a live batch handle into a `SerializableArrowBatch` before fetching.
- Added `SerializableArrowBatch.ToArrowBatch(opts ...Option)` to reconstruct a fully functional `ArrowBatch` on a worker node using only an injected `*http.Client` — no Snowflake session required.
- Extracted `chunkDownloadHeaders` in `chunk_downloader.go` as a shared helper so the same SSE-C/GCS/Azure header logic is used by both the live download path and the serializable descriptor construction.
- Populated `BatchRaw.Descriptor` (a new `ChunkDescriptor` field on `ia.BatchRaw`) in `rows.go` for both remote chunks (URL + headers) and the inline first batch (base64-decoded `RowSetBase64`).
- Added `internal/arrow/download.go` with `OpenChunkStream`, `DecodeIPCRecords`, and `DownloadBatchRecords` — low-level primitives that perform HTTP download and IPC decoding without any driver state, used by the reconstructed batch's `Download` closure.
- Added JSON struct tags to `ConversionOptions` fields so they survive serialization as part of `SerializableArrowBatch`.
- Added a `cmd/arrow/serializablebatches` example demonstrating the full producer → serialize → worker → fetch flow.
- Updated `doc.go` with documentation covering the distributed fetch pattern, security considerations (presigned URLs and SSE-C keys are sensitive), and expiry caveats.

### How to test?

Unit tests in `arrowbatches/serializable_test.go` cover:
- Inline batch round-trip (no HTTP client needed).
- Remote batch download with plain and gzip-compressed payloads, including SSE-C header forwarding.
- Error cases: missing HTTP client for remote batches, expired/403 URLs, unsupported formats.
- Full JSON marshal/unmarshal round-trip including nested `RowTypes` and all `ConversionOptions` flags.
- `loadLocation` behavior for fixed-offset, `Local`, named IANA, and empty timezone names.
- Verification that `ConversionOptions` (e.g. `HigherPrecision`) carried in the serialized form are actually applied during fetch, not just round-tripped as data.

Integration tests in `arrowbatches/serialization_integration_test.go` (require a live account) cover:
- A 100,000-row result serialized, JSON round-tripped, and re-fetched with a connection-less HTTP client, with value and row-count integrity checks.
- Structured/nested column types surviving the serialize → reconstruct → fetch path.

### Why make this change?

Arrow batches returned by `GetArrowBatches` are live handles tied to the Snowflake connection that produced them. Distributed compute frameworks (Spark, Hadoop, etc.) need to partition a large result on a coordinator node and download chunks in parallel on worker nodes that have no Snowflake session. `SerializableArrowBatch` provides the seam for this: the coordinator serializes batch descriptors (which embed presigned URLs and decryption keys), ships them to workers, and each worker fetches its chunk independently using only a standard HTTP client.

### DIFF
--- a/arrowbatches/serializable.go
+++ b/arrowbatches/serializable.go
@@ -35,9 +35,10 @@ const arrowFormatName = "arrow"
 type SerializableArrowBatch struct {
 	// Format is the chunk payload format; currently always "arrow".
 	Format string `json:"format"`
-	// InlineData holds the Arrow IPC bytes for the first batch (embedded in the query
-	// response). When set, the batch is local and no download is performed.
-	InlineData []byte `json:"inlineData,omitempty"`
+	// InlineData holds the base64-encoded Arrow IPC bytes for the first batch (carried
+	// verbatim from the query response). When set, the batch is local and no download is
+	// performed; the bytes are base64-decoded only when the batch is materialized.
+	InlineData string `json:"inlineData,omitempty"`
 	// URL is the presigned cloud-storage URL for a remote chunk (empty for inline batches).
 	URL string `json:"url,omitempty"`
 	// Headers are the HTTP headers required to fetch URL, including the SSE-C key.
@@ -68,7 +69,7 @@ type SerializableArrowBatch struct {
 // released, though the descriptor itself remains valid for re-serialization.
 func (rb *ArrowBatch) ToSerializable() (SerializableArrowBatch, error) {
 	desc := rb.raw.Descriptor
-	if len(desc.InlineData) == 0 && desc.URL == "" {
+	if desc.InlineDataBase64 == "" && desc.URL == "" {
 		return SerializableArrowBatch{}, fmt.Errorf(
 			"arrowbatches: batch has neither inline data nor a chunk URL and cannot be serialized")
 	}
@@ -77,7 +78,7 @@ func (rb *ArrowBatch) ToSerializable() (SerializableArrowBatch, error) {
 
 	return SerializableArrowBatch{
 		Format:                arrowFormatName,
-		InlineData:            desc.InlineData,
+		InlineData:            desc.InlineDataBase64,
 		URL:                   desc.URL,
 		Headers:               desc.Headers,
 		RowCount:              rb.raw.RowCount,
@@ -98,9 +99,11 @@ type arrowBatchOpts struct {
 // Option configures ToArrowBatch.
 type Option func(*arrowBatchOpts)
 
-// WithHTTPClient injects the HTTP client used to download remote chunks. It is required
-// for batches that reference a URL; the worker has no Snowflake connection to build one
-// from, and this is also the seam for configuring proxy/TLS/timeouts.
+// WithHTTPClient injects the HTTP client used to download remote chunks. It is optional:
+// when not set, http.DefaultClient is used. Since the worker has no Snowflake connection to
+// inherit transport settings from, this is the seam for configuring proxy/TLS/timeouts;
+// note that http.DefaultClient has no timeout, so pass a client with one (or a Fetch context
+// with a deadline) for remote batches.
 func WithHTTPClient(client *http.Client) Option {
 	return func(o *arrowBatchOpts) { o.client = client }
 }
@@ -115,9 +118,10 @@ func WithAllocator(pool memory.Allocator) Option {
 // The returned batch's Fetch downloads (if remote) and transforms the records exactly
 // as a batch obtained directly from GetArrowBatches would.
 //
-// A remote batch (URL set) requires an *http.Client via WithHTTPClient. Inline batches
-// need no client. Call WithContext on the returned batch to bind a cancellation context
-// to Fetch.
+// The HTTP client used for a remote batch defaults to http.DefaultClient; supply your own
+// with WithHTTPClient to configure proxy/TLS/timeouts (recommended, since http.DefaultClient
+// has no timeout). Inline batches perform no download. Call WithContext on the returned batch
+// to bind a cancellation context to Fetch.
 func (s SerializableArrowBatch) ToArrowBatch(opts ...Option) (*ArrowBatch, error) {
 	if s.Format != "" && s.Format != arrowFormatName {
 		return nil, fmt.Errorf("arrowbatches: unsupported serialized batch format %q", s.Format)
@@ -130,17 +134,15 @@ func (s SerializableArrowBatch) ToArrowBatch(opts ...Option) (*ArrowBatch, error
 	if o.allocator == nil {
 		o.allocator = memory.DefaultAllocator
 	}
-
-	remote := len(s.InlineData) == 0 && s.URL != ""
-	if remote && o.client == nil {
-		return nil, fmt.Errorf("arrowbatches: a remote arrow batch requires an *http.Client (use WithHTTPClient)")
+	if o.client == nil {
+		o.client = http.DefaultClient
 	}
 
 	client, alloc := o.client, o.allocator
 	desc := ia.ChunkDescriptor{
 		URL:              s.URL,
 		Headers:          s.Headers,
-		InlineData:       s.InlineData,
+		InlineDataBase64: s.InlineData,
 		UncompressedSize: s.UncompressedSize,
 	}
 
--- a/arrowbatches/serializable.go
+++ b/arrowbatches/serializable.go
@@ -1,0 +1,185 @@
+package arrowbatches
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/ipc"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+
+	ia "github.com/snowflakedb/gosnowflake/v2/internal/arrow"
+	"github.com/snowflakedb/gosnowflake/v2/internal/query"
+)
+
+// arrowFormatName is the only result format arrow batches carry today. The Format
+// field is kept for forward compatibility with the streaming/JSON result path.
+const arrowFormatName = "arrow"
+
+// SerializableArrowBatch is a portable, self-contained description of an ArrowBatch.
+// It can be marshaled (e.g. with encoding/json), shipped to another machine that has
+// no Snowflake connection, and turned back into an ArrowBatch with ToArrowBatch. This
+// enables distributed fetch: partition a result set on one node and download the
+// chunks in parallel across many.
+//
+// Security: for remote batches, Headers carries the SSE-C decryption key and URL is a
+// presigned bearer capability granting read access to the chunk. Treat a serialized
+// batch as sensitive and transmit it only over secure channels.
+//
+// Expiry: presigned URLs are valid only for a server-controlled window (typically a few
+// hours). A remote SerializableArrowBatch must be shipped and fetched within it. Inline
+// batches (InlineData set) embed their data and never expire.
+type SerializableArrowBatch struct {
+	// Format is the chunk payload format; currently always "arrow".
+	Format string `json:"format"`
+	// InlineData holds the Arrow IPC bytes for the first batch (embedded in the query
+	// response). When set, the batch is local and no download is performed.
+	InlineData []byte `json:"inlineData,omitempty"`
+	// URL is the presigned cloud-storage URL for a remote chunk (empty for inline batches).
+	URL string `json:"url,omitempty"`
+	// Headers are the HTTP headers required to fetch URL, including the SSE-C key.
+	Headers map[string]string `json:"headers,omitempty"`
+	// RowCount is the number of rows in the batch.
+	RowCount int `json:"rowCount"`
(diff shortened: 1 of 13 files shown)


arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+	ConversionOptions
+}
+
+// ToSerializable converts a batch into a portable descriptor for distributed fetch.
+// It captures the chunk address (inline bytes or URL + headers), the column metadata,
+// timezone, and the conversion options that were captured when the batch was created.
+//
+// Call ToSerializable before Fetch: after a successful Fetch the in-memory records are
+// released, though the descriptor itself remains valid for re-serialization.
+func (rb *ArrowBatch) ToSerializable() (SerializableArrowBatch, error) {
```
sfc-gh-boler, line 70: Just "Serialize" or "Marshal" to be consistent with the usual naming scheme.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+// records. Defaults to memory.DefaultAllocator.
+func WithAllocator(pool memory.Allocator) Option {
+	return func(o *arrowBatchOpts) { o.allocator = pool }
+}
+
+// ToArrowBatch reconstructs an ArrowBatch from its serialized form on a worker node.
+// The returned batch's Fetch downloads (if remote) and transforms the records exactly
+// as a batch obtained directly from GetArrowBatches would.
+//
+// A remote batch (URL set) requires an *http.Client via WithHTTPClient. Inline batches
```
sfc-gh-boler: It it requires a param it should be passed directly, not through options.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+// chunks in parallel across many.
+//
+// Security: for remote batches, Headers carries the SSE-C decryption key and URL is a
+// presigned bearer capability granting read access to the chunk. Treat a serialized
+// batch as sensitive and transmit it only over secure channels.
+//
+// Expiry: presigned URLs are valid only for a server-controlled window (typically a few
+// hours). A remote SerializableArrowBatch must be shipped and fetched within it. Inline
+// batches (InlineData set) embed their data and never expire.
+type SerializableArrowBatch struct {
```
sfc-gh-boler, line 35: It'd just wall it "Serialized". ArrowBatch suffix is pointless, because it is present in the package name already.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+	Headers map[string]string `json:"headers,omitempty"`
+	// RowCount is the number of rows in the batch.
+	RowCount int `json:"rowCount"`
+	// UncompressedSize is the reported uncompressed size of the chunk in bytes.
+	UncompressedSize int64 `json:"uncompressedSize,omitempty"`
+	// RowTypes is the column metadata needed to transform the raw records.
+	RowTypes []query.ExecResponseRowType `json:"rowTypes"`
+	// TimezoneName is the batch's timezone name (loc.String()).
+	TimezoneName string `json:"timezoneName,omitempty"`
+	// TimezoneOffsetSeconds is the batch's timezone offset, used as a fallback when the
```
sfc-gh-boler, line 54: What's the point of TZName if we give the offset already?

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+	// worker cannot resolve TimezoneName (e.g. fixed-offset zones, or missing tzdata).
+	TimezoneOffsetSeconds int `json:"timezoneOffsetSeconds,omitempty"`
```
sfc-gh-boler, line 56: TZs don't have offsets in seconds. That's confusing. Minutes would be fine.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// ToArrowBatch reconstructs an ArrowBatch from its serialized form on a worker node.
+// The returned batch's Fetch downloads (if remote) and transforms the records exactly
+// as a batch obtained directly from GetArrowBatches would.
+//
+// A remote batch (URL set) requires an *http.Client via WithHTTPClient. Inline batches
+// need no client. Call WithContext on the returned batch to bind a cancellation context
+// to Fetch.
+func (s SerializableArrowBatch) ToArrowBatch(opts ...Option) (*ArrowBatch, error) {
```
sfc-gh-boler, line 125: Look at the Go conventions (json/proto/etc). The DTO types don't have marshal/unmarshal methods, instead those are normally free functions.

internal/arrow/download.go
```diff
@@ -0,0 +1,135 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("arrow: creating chunk request: %w", err)
+	}
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("arrow: fetching chunk: %w", err)
```
sfc-gh-boler, line 61: is this going to print the URL containing keys (presigned)?

internal/arrow/download.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, fmt.Errorf("arrow: no http client provided for chunk download")
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("arrow: creating chunk request: %w", err)
+	}
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+	resp, err := client.Do(req)
```
sfc-gh-boler, line 59: what's the retry and cancelation story?

rows.go
```diff
@@ -201,13 +202,28 @@ func (rows *snowflakeRows) GetArrowBatches() (*ia.BatchDataInfo, error) {
...
 				return actualRaw.records, actualRaw.rowCount, nil
 			}
+		} else if scd.RowSet.RowSetBase64 != "" {
+			// The first (inline) batch carries its Arrow IPC bytes in the query
+			// response; embed them verbatim so the batch can be serialized without
+			// re-encoding (there is no IPC writer in the driver).
+			if inlineBytes, err := base64.StdEncoding.DecodeString(scd.RowSet.RowSetBase64); err == nil {
+				batch.Descriptor = ia.ChunkDescriptor{InlineData: inlineBytes}
+			} else {
+				logger.WithContext(scd.ctx).Warnf("failed to decode RowSetBase64 for arrow batch serialization: %v", err)
```
sfc-gh-boler: So we log an error, but there is no signalling to the caller it seems.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+
+	ia "github.com/snowflakedb/gosnowflake/v2/internal/arrow"
+	"github.com/snowflakedb/gosnowflake/v2/internal/query"
+)
+
+// arrowFormatName is the only result format arrow batches carry today. The Format
+// field is kept for forward compatibility with the streaming/JSON result path.
+const arrowFormatName = "arrow"
+
+// SerializableArrowBatch is a portable, self-contained description of an ArrowBatch.
```
sfc-gh-boler, line 22: Do we even want to expose those fields? They seem internal to Marsha/Unmarshal.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,185 @@
...
+//
+// Security: for remote batches, Headers carries the SSE-C decryption key and URL is a
+// presigned bearer capability granting read access to the chunk. Treat a serialized
+// batch as sensitive and transmit it only over secure channels.
+//
+// Expiry: presigned URLs are valid only for a server-controlled window (typically a few
+// hours). A remote SerializableArrowBatch must be shipped and fetched within it. Inline
+// batches (InlineData set) embed their data and never expire.
+type SerializableArrowBatch struct {
+	// Format is the chunk payload format; currently always "arrow".
```
sfc-gh-boler, line 36: What does this mean specifically? In other words, when is this going to change?

If we want to provide a contract between orchestrators and executors (distributed system), we need a version number here. 

What about orchestration between different drivers? Is the plan to support it?

sfc-gh-boler (changes requested):
Just a first pass and some high level commentary. This is a big enough change that it deserves a technical design doc and a PRD.

arrowbatches/serializable.go
sfc-gh-pfus: Yes, this is needed for JSON serialisation.

arrowbatches/serializable.go
sfc-gh-pfus: But it's not serialised, it's only able to be serialised. I can shorten the name, it may be more go-like, but is `arrowbatches.Serializable` fine?

arrowbatches/serializable.go
sfc-gh-pfus: This means, that this should be always `Arrow`, not `JSON`.
This is never an external executors, it should be always handled by a drivers. So all these should be an implementation detail.
About other drivers - only Go driver supports arrow batches in this form. We can think about improvements in UD, if needed.

arrowbatches/serializable.go
sfc-gh-pfus: Because Go is not Snowflake :) Snowflake uses offset only for timestamps, but to create a correct timezone in Go you need both. It is needed for LTZ.

arrowbatches/serializable.go
sfc-gh-pfus: But Go API requires seconds to create a location:

```
// FixedZone returns a [Location] that always uses
// the given zone name and offset (seconds east of UTC).
func FixedZone(name string, offset int) *Location {
```

arrowbatches/serializable.go
sfc-gh-pfus: But this is a bit different - this is not already serialised, this function just takes not serialisable struct and turns it to the serialisable struct.

arrowbatches/serializable.go
sfc-gh-pfus: This is not marshall/unmarshall, this is just conversion from one type to another. Is your point still valid? I'd be fine to make it a function if this was to actually marshall, but this one just converts to another, it belongs strictly to this type.

arrowbatches/serializable.go
sfc-gh-pfus: It's not required, for inline chunk it is not used at all. But apart from global timeout, `http.DefaultClient` is good enough. I'm marking it as optional.

internal/arrow/download.go
sfc-gh-pfus: Re: cancellation - this request takes context, so it should honour cancellations as usual.
Re: retry - we have an ugly retry mechanism, but it is in our main package, not available here. I can move it to internal package (`http`?), but I'd prefer doing this in another PR.

internal/arrow/download.go
sfc-gh-pfus: It shouldn't, it goes through our secret detector.

sfc-gh-pfus:
* **#1833** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1833?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

rows.go
sfc-gh-pfus: Good catch. I also think that there's no point in decoding it here, since this field is used only for serialising it later. So I kept base64 instead of byte array.

arrowbatches/serializable.go
```diff
@@ -0,0 +1,187 @@
...
+	Format string `json:"format"`
+	// InlineData holds the base64-encoded Arrow IPC bytes for the first batch (carried
+	// verbatim from the query response). When set, the batch is local and no download is
+	// performed; the bytes are base64-decoded only when the batch is materialized.
+	InlineData string `json:"inlineData,omitempty"`
+	// URL is the presigned cloud-storage URL for a remote chunk (empty for inline batches).
+	URL string `json:"url,omitempty"`
+	// Headers are the HTTP headers required to fetch URL, including the SSE-C key.
+	Headers map[string]string `json:"headers,omitempty"`
```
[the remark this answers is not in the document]
sfc-gh-pfus: This is needed - we need to distribute headers with presigned URLs across machines.

PR #1834, opened 2026-08-06
SNOW-2499292 Implement serialisable arrow streams

sfc-gh-pfus:
### TL;DR

Adds `SerializableArrowStreamBatch` to enable distributed Arrow result-set fetching without a Snowflake connection on worker nodes.

### What changed?

Introduced `SerializableArrowStreamBatch`, a JSON-serializable descriptor that captures everything needed to download and read a single Arrow result chunk independently of the originating Snowflake connection. It carries the presigned chunk URL, required HTTP headers (including SSE-C decryption keys for S3, or provider-specific headers for GCS/Azure), inline data for the first batch, row count, and format.

- `ArrowStreamBatch.ToSerializable()` converts a batch into this portable descriptor, handling both inline (first-batch) and remote (presigned URL) cases.
- `SerializableArrowStreamBatch.Reader(opts...)` reconstructs an `ArrowStreamPartitionReader` from the descriptor. Remote batches require an `*http.Client` supplied via `WithStreamHTTPClient`.
- `ArrowStreamPartitionReader.GetStream(ctx)` returns an `io.ReadCloser` over the chunk bytes, downloading and gzip-inflating as needed, with no Snowflake session involved.
- `openStandaloneChunkStream` performs the HTTP GET against the presigned URL, mirrors the gzip-detection logic from the existing downloader, and surfaces HTTP errors as `SnowflakeError`.

### How to test?

Unit tests cover:
- `ToSerializable` for remote SSE-C, remote ChunkHeader (GCS/Azure), inline, and missing-downloader error cases.
- `ArrowStreamPartitionReader` for plain and gzip-compressed remote payloads, inline payloads, missing-client errors, and expired (403) URLs.
- JSON round-trip of the descriptor struct.

An integration test (`TestQueryArrowStreamSerializableRoundTrip`) runs a 50,000-row query, serializes each batch descriptor to JSON, deserializes it, fetches the chunk via a bare `*http.Client`, and asserts that the reconstructed bytes are byte-for-byte identical to the original and decode to the correct total row count via `ipc.NewReader`.

### Why make this change?

Distributed query engines (e.g. ADBC `ExecutePartitions` / `ReadPartition`) need to split a Snowflake result set on a coordinator and download chunks in parallel across worker nodes that have no Snowflake connection. Previously, `ArrowStreamBatch` was tightly coupled to the driver's internal downloader and could not be shipped across process or machine boundaries. `SerializableArrowStreamBatch` decouples the chunk descriptor from the connection, enabling fan-out fetch patterns while keeping the security-sensitive presigned URL and SSE-C key contained within the serialized payload.

### DIFF
--- a/arrow_stream_serializable.go
+++ b/arrow_stream_serializable.go
@@ -1,0 +1,184 @@
+package gosnowflake
+
+import (
+	"bufio"
+	"bytes"
+	"compress/gzip"
+	"context"
+	"fmt"
+	"io"
+	"maps"
+	"net/http"
+)
+
+// SerializableArrowStreamBatch is a portable, self-contained description of an
+// ArrowStreamBatch (see QueryArrowStream / ArrowStreamLoader). It can be marshaled
+// (e.g. with encoding/json), shipped to another process or machine that has no Snowflake
+// connection, and turned back into a readable stream with Reader. This enables distributed
+// fetch: partition a result set on one node and download the chunks in parallel across many
+// (e.g. from an ADBC driver's ExecutePartitions / ReadPartition).
+//
+// The descriptor carries only the chunk address and payload format; it deliberately does not
+// carry column metadata or timezone. A consumer that needs to decode the bytes obtains
+// RowTypes() and Location() from the ArrowStreamLoader and carries them alongside.
+//
+// Security: for a remote batch, Headers carries the SSE-C decryption key and URL is a
+// presigned bearer capability granting read access to the chunk. Treat a serialized batch as
+// sensitive and transmit it only over secure channels.
+//
+// Expiry: presigned URLs are valid only for a server-controlled window (typically a few
+// hours). A remote batch must be shipped and fetched within it. Inline batches (InlineData
+// set) embed their bytes and never expire.
+type SerializableArrowStreamBatch struct {
+	// Format is the payload format of the stream, "arrow" or "json".
+	Format string `json:"format"`
+	// InlineData holds the batch bytes embedded in the query response (the first batch).
+	// When set, the batch is local and no download is performed.
+	InlineData []byte `json:"inlineData,omitempty"`
+	// URL is the presigned cloud-storage URL for a remote chunk (empty for inline batches).
+	URL string `json:"url,omitempty"`
+	// Headers are the HTTP headers required to fetch URL, including the SSE-C key.
+	Headers map[string]string `json:"headers,omitempty"`
+	// NumRows is the number of rows the metadata reported for this batch.
+	NumRows int64 `json:"numRows"`
+	// UncompressedSize is the reported uncompressed size of the chunk in bytes.
+	UncompressedSize int64 `json:"uncompressedSize,omitempty"`
+}
+
+// ToSerializable converts a stream batch into a portable descriptor for distributed fetch.
+// Call it before GetStream. The batch must have been produced by ArrowStreamLoader.GetBatches
+// (i.e. be backed by a downloader).
+func (asb *ArrowStreamBatch) ToSerializable() (SerializableArrowStreamBatch, error) {
+	if asb.scd == nil {
+		return SerializableArrowStreamBatch{}, fmt.Errorf(
+			"arrow stream batch is not backed by a downloader and cannot be serialized")
+	}
+	s := SerializableArrowStreamBatch{
+		Format:  asb.scd.queryResultFormat,
+		NumRows: asb.numrows,
+	}
+	if asb.inlineData != nil {
+		s.InlineData = asb.inlineData
+		return s, nil
+	}
+	meta := asb.scd.ChunkMetas[asb.idx]
+	s.URL = meta.URL
+	s.UncompressedSize = meta.UncompressedSize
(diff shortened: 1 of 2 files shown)


sfc-gh-pfus:
* **#1834** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// openStandaloneChunkStream performs an HTTP GET against a presigned result-chunk URL using
+// the injected client (no Snowflake connection) and returns a reader over the (gzip-inflated
+// if needed) payload. It mirrors the transport handling in downloadChunkStreamHelper. The
+// caller must Close the returned stream.
+func openStandaloneChunkStream(ctx context.Context, client *http.Client, url string, headers map[string]string) (io.ReadCloser, error) {
+	if client == nil {
+		return nil, fmt.Errorf("gosnowflake: no http client provided for chunk download")
+	}
```
zeroshade, line 150: is there any reason not to allow using a default http.Client{} if one wasn't set? that would be preferable to erroring here.

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+	for k, v := range headers {
+		req.Header.Set(k, v)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		defer func() { _ = resp.Body.Close() }()
+		b, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
```
zeroshade, line 164: why ignore the error here?

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+	gzipMagic, _ := bufStream.Peek(2)
+	if len(gzipMagic) == 2 && gzipMagic[0] == 0x1f && gzipMagic[1] == 0x8b {
+		gz, err := gzip.NewReader(bufStream)
+		if err != nil {
+			_ = resp.Body.Close()
+			return nil, err
+		}
+		return &streamWrapReader{Reader: gz, wrapped: resp.Body}, nil
+	}
+	return &streamWrapReader{Reader: bufStream, wrapped: resp.Body}, nil
```
zeroshade, line 183: instead of replicating this here, can we make this a function and call it in both places? (here and in the original chunk handler)

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+	bufStream := bufio.NewReader(resp.Body)
+	// Peek is best-effort: a short or empty payload simply skips gzip detection.
+	gzipMagic, _ := bufStream.Peek(2)
+	if len(gzipMagic) == 2 && gzipMagic[0] == 0x1f && gzipMagic[1] == 0x8b {
+		gz, err := gzip.NewReader(bufStream)
+		if err != nil {
+			_ = resp.Body.Close()
+			return nil, err
+		}
+		return &streamWrapReader{Reader: gz, wrapped: resp.Body}, nil
```
zeroshade, line 181: You can simplify this by doing `bufStream = gz` instead so that you only have a single `return` at the bottom

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+// SerializableArrowStreamBatch is a portable, self-contained description of an
+// ArrowStreamBatch (see QueryArrowStream / ArrowStreamLoader). It can be marshaled
+// (e.g. with encoding/json), shipped to another process or machine that has no Snowflake
+// connection, and turned back into a readable stream with Reader. This enables distributed
+// fetch: partition a result set on one node and download the chunks in parallel across many
+// (e.g. from an ADBC driver's ExecutePartitions / ReadPartition).
+//
+// The descriptor carries only the chunk address and payload format; it deliberately does not
+// carry column metadata or timezone. A consumer that needs to decode the bytes obtains
+// RowTypes() and Location() from the ArrowStreamLoader and carries them alongside.
```
zeroshade, line 23: why do this instead of including them in the serialized data already?

In addition, is there a way to know whether the result set is sorted (order matters) versus when order doesn't matter?

arrow_stream_serializable.go
```diff
@@ -0,0 +1,184 @@
...
+	if asb.inlineData != nil {
+		s.InlineData = asb.inlineData
+		return s, nil
+	}
+	meta := asb.scd.ChunkMetas[asb.idx]
+	s.URL = meta.URL
+	s.UncompressedSize = meta.UncompressedSize
+	s.Headers = streamChunkHeaders(asb.scd.ChunkHeader, asb.scd.Qrmk)
+	return s, nil
+}
```
zeroshade, line 69: why a separate `ToSerializable` method instead of just implementing the UnmarshalJSON directly on ArrowStreamBatch?

zeroshade (changes requested):

PR #1835, opened 2026-08-06
Fix precision loss for fixed-point values with non-zero scale

tanner-lai:
### Description

Both non-higher-precision conversion paths for fixed-point values routed the exact unscaled integer through a binary float before formatting it, silently rounding values carrying more significant digits than the float's mantissa could hold. arrowDecimal128ToValue divided using a 64-bit big.Float, so NUMBER(20,10) holding 1234567890.1234567890 returned 1234567890.1234567889; arrowIntToValue divided as float64, so NUMBER(18,4) holding 12345678901234.5678 returned 12345678901234.5684. Both now format from the unscaled integer via decimal128.ToString, as the scale == 0 branch already did. The higherPrecision paths still return *big.Float and are left unchanged, since that return type is approximate by contract.

### Checklist


sfc-gh-pfus:
Hi @tanner-lai ! Thanks for your contribution, it seems correct. Can you add a changelog.md entry?

tanner-lai:

tanner-lai:
> Hi @tanner-lai ! Thanks for your contribution, it seems correct. Can you add a changelog.md entry?

Just added an entry, thanks!

tanner-lai:
@sfc-gh-pfus Is there anything left to address? Do you guys have a practice to older versions/do I need to make those PRs as well?

sfc-gh-pfus:
@tanner-lai sorry, I was on holiday. I'll get to your PR this week!

sfc-gh-pfus (approved):

PR #1840, opened 2026-08-27
SNOW-4017183 Add minimum TLS version enforcing propertly

sfc-gh-pfus:
### TL;DR

Adds support for configuring a minimum TLS version via the `SNOWFLAKE_MIN_TLS_VERSION` environment variable, enforced globally across all HTTP transports.

### What changed?

A new `GetMinTLSVersion` function reads the `SNOWFLAKE_MIN_TLS_VERSION` environment variable and parses it into a TLS version constant. An `ApplyMinTLSVersion` function applies this version as a floor on any `tls.Config`, overriding any previously set `MinVersion` to enforce it as a security policy. Valid values are `1.0`, `1.1`, `1.2`, and `1.3`.

`createBaseTransport` now calls `ApplyMinTLSVersion` before constructing the transport and returns an error if the environment variable contains an invalid value. As a result, `createNoRevocationTransport` and several other transport creation methods now return errors, and their call sites in `ocsp.go` and `transport.go` have been updated to handle those errors appropriately.

### How to test?

- Set `SNOWFLAKE_MIN_TLS_VERSION=1.3` and verify that connections enforce TLS 1.3 as the minimum.
- Set `SNOWFLAKE_MIN_TLS_VERSION=invalid` and verify that transport creation fails with a descriptive error.
- Leave `SNOWFLAKE_MIN_TLS_VERSION` unset and verify existing behavior is unchanged.
- Run the new unit tests in `internal/config/min_tls_version_test.go` covering `GetMinTLSVersion`, `ApplyMinTLSVersion`, and `getTLSVersionName`.

### Why make this change?

Some environments require a minimum TLS version to be enforced as a security policy. This change provides a way to configure that floor at the environment level, ensuring it cannot be bypassed by custom `tls.Config` values set elsewhere in the driver.

### DIFF
--- a/internal/config/min_tls_version.go
+++ b/internal/config/min_tls_version.go
@@ -4,41 +4,42 @@ import (
 	"crypto/tls"
 	"fmt"
 	"os"
+	"strings"
 )
 
 const envVarMinTLSVersion = "SNOWFLAKE_MIN_TLS_VERSION"
 
 // GetMinTLSVersion reads and parses the SNOWFLAKE_MIN_TLS_VERSION environment variable.
-// Returns 0 if not set or invalid. Valid values: "1.0", "1.1", "1.2", "1.3".
+// Returns 0 if not set or invalid. Valid values: "1.2", "1.3", "TLSv1.2", "TLSv1.3" (case-insensitive, trimmed).
+// TLS 1.0 and 1.1 are not supported as Snowflake requires TLS 1.2 or higher.
 func GetMinTLSVersion() (uint16, error) {
 	value := os.Getenv(envVarMinTLSVersion)
 	if value == "" {
 		return 0, nil
 	}
 
+	// Normalize: trim whitespace and convert to lowercase
+	value = strings.ToLower(strings.TrimSpace(value))
+
 	switch value {
-	case "1.0":
-		return tls.VersionTLS10, nil
-	case "1.1":
-		return tls.VersionTLS11, nil
-	case "1.2":
+	case "1.2", "tlsv1.2":
 		return tls.VersionTLS12, nil
-	case "1.3":
+	case "1.3", "tlsv1.3":
 		return tls.VersionTLS13, nil
 	default:
-		return 0, fmt.Errorf("invalid %s value: %s. Valid values are: 1.0, 1.1, 1.2, 1.3", envVarMinTLSVersion, value)
+		return 0, fmt.Errorf("invalid %s value: %s. Valid values are: 1.2, 1.3, TLSv1.2, TLSv1.3", envVarMinTLSVersion, value)
 	}
 }
 
 // ApplyMinTLSVersion applies the minimum TLS version from environment variable to the given tls.Config.
-// This enforces a global security policy that cannot be bypassed by custom tls.Config.
+// This enforces a global security policy by taking the maximum of the existing MinVersion and the env var value.
 func ApplyMinTLSVersion(tlsConfig *tls.Config) (*tls.Config, error) {
-	minVersion, err := GetMinTLSVersion()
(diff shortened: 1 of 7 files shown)


sfc-gh-pfus:
* **#1840** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

internal/config/min_tls_version.go
```diff
@@ -0,0 +1,71 @@
...
+		return tlsConfig, nil
+	}
+
+	if tlsConfig == nil {
+		tlsConfig = &tls.Config{}
+	}
+
+	// Always set MinVersion when the env var is set, even if tlsConfig already has one
+	// This ensures the environment variable acts as a security policy enforcement
+	tlsConfig.MinVersion = minVersion
```
sfc-gh-dheyman: Do we want this? this actually always overwrites `MinVersion` instead of raising a floor. From what I see go 1.22+ already defaults Min TLS version to 1.2. IMHO we should to max(go_default, min_version_set_by_user) ;)

internal/config/min_tls_version.go
```diff
@@ -0,0 +1,71 @@
...
+// GetMinTLSVersion reads and parses the SNOWFLAKE_MIN_TLS_VERSION environment variable.
+// Returns 0 if not set or invalid. Valid values: "1.0", "1.1", "1.2", "1.3".
+func GetMinTLSVersion() (uint16, error) {
+	value := os.Getenv(envVarMinTLSVersion)
+	if value == "" {
+		return 0, nil
+	}
+
+	switch value {
+	case "1.0":
```
sfc-gh-dheyman: JDBC uses `TLSv1.2` / `TLSv1.3`. Also no trim, no case-fold, so `"1.2 "`, `"TLS1.3"`, `"tlsv1.2"` all will fail transport creation. Just flagging it

transport.go
```diff
@@ -100,8 +100,16 @@ func (tf *transportFactory) createProxy(transportConfig *transportConfig) func(*
 }
 
 // createBaseTransport creates a base HTTP transport with the given configuration
-func (tf *transportFactory) createBaseTransport(transportConfig *transportConfig, tlsConfig *tls.Config) *http.Transport {
+func (tf *transportFactory) createBaseTransport(transportConfig *transportConfig, tlsConfig *tls.Config) (*http.Transport, error) {
 	logger.Debugf("Create a new Base Transport with transportConfig %v", transportConfig.String())
+
+	// Apply minimum TLS version from environment variable
+	var err error
+	tlsConfig, err = sfconfig.ApplyMinTLSVersion(tlsConfig)
```
sfc-gh-dheyman, line 108: this will add a global “minimum TLS version” rule, but only for HTTP clients the driver builds so if you plug in your own HTTP client (Config.Transporter), this is never applied

sfc-gh-dheyman (changes requested):

sfc-gh-dszmolka:
why do we want to allow anything lower than TLSv1.2 when Snowflake itself doesn't support it ? Other drivers also floor it to 1.2

internal/config/min_tls_version.go
sfc-gh-pfus: I'm fine to change it to JDBC version.

transport.go
sfc-gh-pfus: Good catch, added.

sfc-gh-dheyman (approved):

PR #1841, opened 2026-08-28
SNOW-4017185 Fix the region parameter rewriting an explicitly provided host

sfc-gh-pfus:
### TL;DR

Fixed a bug where providing an explicit `host` in the DSN or config was being overwritten when a `region` parameter was also present.

### What changed?

When both `host` and `region` are provided, the driver no longer splices the region into the host. A new helper, `hostProvidedByUser`, checks whether the host was explicitly supplied by the user (via `inputShape.HostProvided`). The region-injection logic in `FillMissingConfigParameters` now only runs when the host was synthesized by the driver from the `account` parameter — not when the user provided it directly.

For example, `host=myacct.snowflakecomputing.com` with `region=us-east-1` previously produced `myacct.us-east-1.snowflakecomputing.com`. The explicit host is now preserved as-is. Non-Snowflake hosts (e.g. TS region endpoints like `myaccount.us-isob-east-1.sc2s.sgov.gov`) are also preserved verbatim when supplied alongside an `account`.

### How to test?

- `TestFillMissingConfigParametersKeepsExplicitHostOverRegion` — verifies that an explicit host is not rewritten when a region is also set.
- `TestFillMissingConfigParametersInjectsRegionIntoSynthesizedHost` — verifies that a driver-synthesized host (built from `account`) still gets the region spliced in correctly.
- `TestFillMissingConfigParametersPreservesNonSnowflakeHostWithAccount` — verifies that non-Snowflake hosts are preserved verbatim when `account` is provided alongside them.
- `TestParseDSNPreservesNonSnowflakeHostWithPort` — verifies that a non-Snowflake host with an explicit port in the DSN is not rewritten.
- Two new DSN parse test cases covering the explicit-host-with-region and non-Snowflake-host-with-port scenarios.

### Why make this change?

The `transformAccountToHost` function builds the `Host` field before `parseDSNParams` reads the `region` parameter. The region-injection step was designed to patch up that synthesized host, but it ran unconditionally — including when the user had explicitly provided a `host`. This caused user-supplied hosts to be silently rewritten, breaking connections to custom or non-standard endpoints (snowflakedb/gosnowflake#1841).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Bug fixes:
 - Fixed `GET` from a LOCAL_FS stage downloading 0 files and returning `264011: not implemented`. Cloud downloads were unaffected (snowflakedb/gosnowflake#1810).
 - Fixed `NUMBER` columns with a non-zero scale losing precision: values needing more significant digits than a binary float's mantissa were silently rounded, so `NUMBER(20,10)` holding `1234567890.1234567890` returned `1234567890.1234567889`. The `WithHigherPrecision` path is unchanged and still returns a 64-bit `*big.Float` (snowflakedb/gosnowflake#1835).
 - Fixed proxy configuration for a literal IPv6 `ProxyHost`, which was concatenated into an unbracketed `host:port` and left the proxy URL unparseable (snowflakedb/gosnowflake#TBD).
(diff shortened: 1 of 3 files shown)


sfc-gh-pfus:
* **#1841** <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1841?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/gosnowflake/1841?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sfc-gh-rsavenok (approved):

PR #1846, opened 2026-09-11
Treat RESTARTED query status as still running

### LINKED ISSUE #1842: SNOW-4053036: GetQueryStatus returns a nil error for queries in RESTARTED state

This is about the source rather than a particular build. The same code is on master as I write this.

NA. `isRunning()` and `isError()` are plain Go with no platform-dependent path.

run `go version` in your console

NA, for the same reason.

You may get the server version by running a query:
```
SELECT CURRENT_VERSION();
```

NA to the mapping, which does not vary by server version. It is relevant to when and why the server emits `RESTARTED`, but that is the part I cannot observe from the client.

5. What did you do?

Traced how `GetQueryStatus` reports a query in `RESTARTED` state.

`SFQueryRestarted` is absent from the case lists of both `isRunning()` and `isError()` ([monitoring.go#L59-L77](https://github.com/snowflakedb/gosnowflake/blob/v2.1.0/monitoring.go#L59-L77)). A `RESTARTED` status therefore passes through both branches of `checkQueryStatus` and reaches the `//success` return at [L212-L213](https://github.com/snowflakedb/gosnowflake/blob/v2.1.0/monitoring.go#L212-L213), so `GetQueryStatus` returns `(*SnowflakeQueryStatus, nil)`.

That error is the only signal a caller receives, and the doc comment on `checkQueryStatus` documents nil plus three error types, so a caller can only write:

```go
_, err := sc.GetQueryStatus(ctx, queryID)
switch {
case err == nil:
	// documented as "there is a complete query result to fetch"
case errors.Is(err, gosnowflake.ErrQueryIsRunning):
case errors.Is(err, gosnowflake.ErrQueryReportedError):
default: // ErrQueryStatus
}
```

`RESTARTED` matches none of the three error cases, so it arrives as nil.

I have no recipe to offer, as I could not find a way to put a query into `RESTARTED` on demand. This comes from reading the source rather than from an incident, so I may be misreading it. I searched open and recently closed issues and did not find it reported.

I could not form a firm expectation, because I could not find documentation of what `RESTARTED` means on this path. It looks like a state the endpoint can return: the driver enumerates it in `queryResultStatus`, `String()` and `strQueryStatusMap`, and snowflake-connector-python lists it in its `QueryStatus` enum too. But it is absent from the documented values of both public `QUERY_HISTORY` surfaces, where the ACCOUNT_USAGE view lists `success`, `fail`, `incident` and the INFORMATION_SCHEMA table function lists `resuming_warehouse`, `running`, `queued`, `blocked`, `success`, `failed_with_error`, `failed_with_incident`.

A caller cannot resolve the ambiguity either. The raw status does reach the driver: `retStatus` carries a `Status string` field. But `GetQueryStatus` does not copy it into `SnowflakeQueryStatus` ([connection.go#L516-L533](https://github.com/snowflakedb/gosnowflake/blob/v2.1.0/connection.go#L516-L533)), which exposes only `SQLText`, `StartTime`, `EndTime`, `ErrorCode`, `ErrorMessage` and the two stats. Any query reaching the status check has an empty `ErrorCode`, so nothing in the exposed struct separates `RESTARTED` from `SUCCESS`. The Python connector omits `RESTARTED` from `is_still_running()` and `is_an_error()` as well, but `get_query_status()` returns the `QueryStatus` enum, so its callers can still apply their own policy.

The two readings lead to different outcomes.

If `RESTARTED` means the result is complete, the current behavior is correct and this can be closed. A line in the `checkQueryStatus` doc comment would spare the next reader the same investigation.

If it does not, a caller that follows the documented contract fetches the result of a query that has not finished, with no error and no signal that anything is wrong. That is the case I cannot rule out, and the reason I am asking rather than working around it.

If it is an oversight, two shapes occurred to me, though you will know better which fits the driver's design. Either resolves it; the first alone would suffice.

1. Copy `retStatus.Status` through to `SnowflakeQueryStatus` as a `Status string` field. Additive, and it also covers statuses added server-side before the driver catches up.
2. Classify `SFQueryRestarted` explicitly: add it to `isRunning()`, or return a dedicated sentinel error.

I am glad to send a PR, but the choice rests on semantics I cannot observe, so I would rather hear the answer first.

7. Can you set logging to DEBUG and collect the logs?

I have not captured debug logs, as there is no failing run behind this.

Hashim1999164:
## Summary
Fixes #1842.

`SFQueryRestarted` was missing from both `isRunning` and `isError`, so `GetQueryStatus` returned nil and callers thought the query was finished. It now counts as still running, same as the other in flight statuses.

## Test plan
- [x] Added `TestQueryResultStatusRestartedIsRunning`
- [ ] Maintainer CI against a live Snowflake account

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@
 
 New features:
 - Added `WorkloadIdentityHost` config option (DSN field `workloadIdentityHost`) that overrides the STS host used by the AWS WIF flows, for endpoints the driver cannot derive from the region (such as an interface VPC endpoint). Must be an STS endpoint for the region the workload runs in (snowflakedb/gosnowflake#5).
- 
+
+Bug fixes:
+- Fixed `GetQueryStatus` treating `RESTARTED` as success by counting that status as still running (snowflakedb/gosnowflake#1842).
+
 ## 2.2.0
 
 New features:
--- a/monitoring.go
+++ b/monitoring.go
@@ -59,7 +59,7 @@ func (qs queryResultStatus) String() string {
 func (qs queryResultStatus) isRunning() bool {
 	switch qs {
 	case SFQueryRunning, SFQueryResumingWarehouse, SFQueryQueued,
-		SFQueryQueueRepairingWarehouse, SFQueryNoData:
+		SFQueryQueueRepairingWarehouse, SFQueryRestarted, SFQueryNoData:
 		return true
 	default:
 		return false
--- a/monitoring_test.go
+++ b/monitoring_test.go
@@ -1,0 +1,12 @@
+package gosnowflake
+
+import "testing"
+
+func TestQueryResultStatusRestartedIsRunning(t *testing.T) {
+	if !SFQueryRestarted.isRunning() {
+		t.Fatal("RESTARTED should be treated as still running")
+	}
+	if SFQueryRestarted.isError() {
+		t.Fatal("RESTARTED should not be treated as an error status")
+	}
+}


Hashim1999164:

Hashim1999164:
recheck

sfc-gh-pfus (approved):
Hi @Hashim1999164 ! Thanks for rising this. I'm approving the PR and importing it to our internal repo. It will be synced to public one when it's merged.

sfc-gh-pfus:
Merged and synced.

